PR #17, opened 2026-08-22
fix(server): resolve Hugging Face repo ids to a local checkpoint dir

gdevenyi:
## The problem

`ft serve --model deepseek-ai/DeepSeek-V4-Flash-0731` fails:

```
FileNotFoundError: No DeepSeek-V4 ModelArgs JSON found under
deepseek-ai/DeepSeek-V4-Flash-0731 (looked for inference/config.json)
```

The string is treated as a local directory. ModelScope ids already resolve; Hugging Face ids do not.

## What this changes

`download_hf_checkpoint` resolves a Hugging Face repo id to a local snapshot directory, downloading through the normal cache if it is not already there. Weight files that the loaders never read are skipped.

A repo id is only recognised when it looks like one. `looks_like_hub_repo_id` requires the `owner/name` shape, and anything that names an existing path is left alone.

## Why the guard matters

My first version sent every `--model` value that was not an existing directory to the Hub. That broke 28 tests in `tests/server`, because they pass names that are neither paths nor repo ids. The regex is what keeps a typo in a local path reporting a missing path instead of a failed download.

## Testing

`tests/server/test_model_path_resolution.py` covers repo ids, local paths, paths that do not exist, and strings that resemble neither.

`tests/server` on `main` with this PR: 540 passed.

### DIFF
--- a/python/freetoken/server/args.py
+++ b/python/freetoken/server/args.py
@@ -646,16 +646,26 @@ def parse_args(
     if is_offload_moe_backend(kwargs["moe_backend"]) and _no_cache_flag:
         kwargs["moe_cache_auto"] = True
 
-    if kwargs["model_source"] == "modelscope":
-        model_path = kwargs["model_path"]
-        if not os.path.isdir(model_path):
+    # Resolve a hub repo id ("org/model") to a local checkpoint directory. This has to
+    # happen here, after served_model_name and the parser cascade have read the repo id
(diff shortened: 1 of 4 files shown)


gdevenyi:

**Result: 1211 passed, 350 skipped, **no new failures**.**

The +6 over main are this PR's own tests, and they ran (not skipped).

gdevenyi:

PR #19, opened 2026-08-22
Support pre-Volta GPUs (sm_61 / Pascal)

gdevenyi:
FreeToken cannot currently be installed or run on a pre-Volta GPU. This branch gets it serving on a GTX 1080 Ti (sm_61, Pascal). It does not change what any supported GPU does.

## Result

Serving `Qwen/Qwen2.5-0.5B-Instruct` on a GTX 1080 Ti, with default sampling:

```
$ curl -s localhost:8099/v1/chat/completions -d '{"model":"Qwen2.5-0.5B-Instruct",
    "messages":[{"role":"user","content":"Name three primary colours."}],"max_tokens":48}'
{"choices":[{"message":{"role":"assistant",
  "content":"Three primary colours are red, blue, and yellow."},
  "finish_reason":"stop"}],
 "usage":{"prompt_tokens":34,"completion_tokens":12,"total_tokens":46}}
```

Test suite on that card: 66 failed and 1270 passed becomes 33 failed and 1327 passed, with 29 skipped.

## What was blocking

Five separate issues. Each is a construct that NVIDIA introduced after Pascal.

| | blocker | fix |
|---|---|---|
| 1 | `__grid_constant__` (compute_70 and up) and `ld.global.L1::no_allocate` (sm_70 and up) in the JIT kernels. nvcc and ptxas reject the whole translation unit, so the offload gather, the index kernels and the KV store never build. | Gate both on `__CUDA_ARCH__ < 700`, as `__nanosleep` in the same header already is. |
| 2 | `tanh.approx.f32` (sm_75 and up) in `_fast_tanh`. | A `constexpr_function` gate and a libdevice fallback, following what `e4m3_compat.py` already does. |
| 3 | `tl.atomic_add` in `moe_align`. Triton lowers every atomic to a scoped and ordered PTX encoding, all of which need sm_70. | Route to the existing atomic-free staged `moe_align_block_size_triton`. |
| 4 | Attention tiles overflow a 48 KB shared-memory budget. | Extend the existing budget ladder and let the compiler arbitrate. |
| 5 | `tl.atomic_add` and `tl.atomic_max` in the top-k and top-p threshold search. | A torch sort-based fallback. |

Both codegen hints in the first row are semantically empty. The parameter is passed identically without the annotation, and the load returns the same bytes without the cache hint.

## Testing

I cannot re-test this on `main` myself. The machine I have access to holds two sm_89 cards and no pre-Volta GPU, so I can only confirm that the gates compile and that the sm_89 paths still behave. The GTX 1080 Ti figures above stand as originally measured.

Anyone with a Pascal or Maxwell card who can re-run the suite would be worth more to this PR than anything further I can do.

Note that a pre-Turing card also needs #26, because the pinned CUDA 13 torch build carries no cubin below sm_75.

### DIFF
--- a/python/freetoken/engine/sample.py
+++ b/python/freetoken/engine/sample.py
@@ -28,9 +28,14 @@ def sample_impl(
     top_p: torch.Tensor | float | None,
 ) -> torch.Tensor:
     from freetoken.kernel.backend import is_flashinfer_installed
+    from freetoken.utils.arch import is_sm70_supported
 
     if is_flashinfer_installed():
         import flashinfer.sampling as sampling
+    elif not is_sm70_supported():
+        # The triton top-p/top-k threshold search accumulates through tl.atomic_*, which
+        # triton can only lower to sm_70+ encodings. Sort with torch instead.
+        import freetoken.kernel.torch_sampling as sampling
     else:
         import freetoken.kernel.triton.sampling as sampling
 
--- a/python/freetoken/kernel/csrc/include/freetoken/utils.cuh
+++ b/python/freetoken/kernel/csrc/include/freetoken/utils.cuh
@@ -10,6 +10,16 @@
 #include <source_location>
 #include <type_traits>
 
+// __grid_constant__ is compute_70+; Pascal's ptxas rejects it outright. It is only ever
+// a codegen hint -- the kernel parameter is passed identically without it -- so dropping
+// it below sm_70 costs nothing but the constant-bank optimisation. The host pass (no
+// __CUDA_ARCH__) and every sm_70+ device pass keep the annotation verbatim.
+#if defined(__CUDA_ARCH__) && __CUDA_ARCH__ < 700
+#define FT_GRID_CONSTANT
+#else
+#define FT_GRID_CONSTANT __grid_constant__
+#endif
+
 namespace device {
 
 inline constexpr auto kWarpThreads = 32u;
--- a/python/freetoken/kernel/csrc/jit/fast_index_copy.cuh
+++ b/python/freetoken/kernel/csrc/jit/fast_index_copy.cuh
@@ -33,21 +33,31 @@ inline constexpr auto get_mem_package() {
     }
 }
 
+// The L1::no_allocate hint keeps this streaming host gather from evicting L1, but the
+// modifier is sm_70+ and Pascal's ptxas rejects it outright. It is a cache hint only, so
+// pre-sm_70 issues the plain load and loses nothing but the anti-pollution optimisation.
+// (st.global.wt below assembles fine on sm_6x, so the stores are left alone.)
+#if defined(__CUDA_ARCH__) && __CUDA_ARCH__ < 700
+#define FT_LD_STREAM "ld.global."
+#else
+#define FT_LD_STREAM "ld.global.L1::no_allocate."
+#endif
+
 __always_inline __device__ auto load_nc(const uint1* __restrict__ src) -> uint1 {
     uint32_t tmp;
-    asm volatile("ld.global.L1::no_allocate.b32 %0,[%1];" : "=r"(tmp) : "l"(src));
+    asm volatile(FT_LD_STREAM "b32 %0,[%1];" : "=r"(tmp) : "l"(src));
     return uint1{tmp};
 }
 
 __always_inline __device__ auto load_nc(const uint2* __restrict__ src) -> uint2 {
     uint32_t tmp0, tmp1;
-    asm volatile("ld.global.L1::no_allocate.v2.b32 {%0,%1},[%2];" : "=r"(tmp0), "=r"(tmp1) : "l"(src));
+    asm volatile(FT_LD_STREAM "v2.b32 {%0,%1},[%2];" : "=r"(tmp0), "=r"(tmp1) : "l"(src));
     return uint2{tmp0, tmp1};
 }
 
 __always_inline __device__ auto load_nc(const uint4* __restrict__ src) -> uint4 {
     uint32_t tmp0, tmp1, tmp2, tmp3;
-    asm volatile("ld.global.L1::no_allocate.v4.b32 {%0,%1,%2,%3},[%4];" : "=r"(tmp0), "=r"(tmp1), "=r"(tmp2), "=r"(tmp3) : "l"(src));
+    asm volatile(FT_LD_STREAM "v4.b32 {%0,%1,%2,%3},[%4];" : "=r"(tmp0), "=r"(tmp1), "=r"(tmp2), "=r"(tmp3) : "l"(src));
     return uint4{tmp0, tmp1, tmp2, tmp3};
 }
 
@@ -485,7 +495,7 @@ struct MultiIndexCopyParams {
 
 template <typename IdType, std::size_t kNumThreads, std::size_t kBlocksPerBank>
 __global__ __launch_bounds__(kNumThreads) void fast_index_copy_multi(
-    const __grid_constant__ MultiIndexCopyParams p
+    const FT_GRID_CONSTANT MultiIndexCopyParams p
 ) {
     const int b = static_cast<int>(blockIdx.x / kBlocksPerBank);
     if (b >= p.num_banks) {
--- a/python/freetoken/kernel/csrc/jit/index.cu
+++ b/python/freetoken/kernel/csrc/jit/index.cu
@@ -31,7 +31,7 @@ struct MaskedKernelParams {
 template <std::size_t kNumThreads, std::size_t kMaxOccupancy, bool kUsePDL,
           std::size_t kElementSize, std::size_t kNumSplits, std::integral T>
 __global__ __launch_bounds__(kNumThreads, kMaxOccupancy) void //
-    index_kernel(const __grid_constant__ IndexKernelParams params) {
+    index_kernel(const FT_GRID_CONSTANT IndexKernelParams params) {
   using namespace device;
   constexpr auto kSize = kElementSize;
   constexpr auto kSizePerWarp = kSize / kNumSplits;
@@ -62,7 +62,7 @@ template <std::size_t kNumThreads, std::size_t kMaxOccupancy, bool kUsePDL,
           std::size_t kElementSize, std::size_t kNumSplits, std::integral T>
 __global__ __launch_bounds__(kNumThreads, kMaxOccupancy) void //
     masked_index_kernel(
-        const __grid_constant__ MaskedKernelParams mask_params) {
+        const FT_GRID_CONSTANT MaskedKernelParams mask_params) {
   using namespace device;
   constexpr auto kSize = kElementSize;
   constexpr auto kSizePerWarp = kSize / kNumSplits;
--- a/python/freetoken/kernel/csrc/jit/store.cu
+++ b/python/freetoken/kernel/csrc/jit/store.cu
@@ -25,7 +25,7 @@ struct StoreKernelParams {
 template <std::size_t kNumThreads, std::size_t kMaxOccupancy, bool kUsePDL,
           std::size_t kElementSize, std::integral T>
 __global__ __launch_bounds__(kNumThreads, kMaxOccupancy) void //
-    store_kv_cache(const __grid_constant__ StoreKernelParams params) {
+    store_kv_cache(const FT_GRID_CONSTANT StoreKernelParams params) {
   using namespace device;
 
   constexpr auto kWarpPerBlock =
--- a/python/freetoken/kernel/csrc/pinned_tensor.cpp
+++ b/python/freetoken/kernel/csrc/pinned_tensor.cpp
@@ -10,6 +10,18 @@ void free_pinned(void *ptr) {
   }
 }
 
+// A failed CUDA runtime call leaves its status in the per-thread "last error" slot, so
+// the next unrelated C10_CUDA_CHECK anywhere in the process reports it instead of its
+// own result. Consume it before throwing to keep the failure local to this call. It
+// matters most for host_device_ptr, which drivers may legitimately reject (pre-sm_70
+// validates registration where newer arches let UVA degenerate the lookup to identity).
+void check_cuda(cudaError_t err, const char *what) {
+  if (err != cudaSuccess) {
+    cudaGetLastError();
+    TORCH_CHECK(false, what, ": ", cudaGetErrorString(err));
+  }
+}
+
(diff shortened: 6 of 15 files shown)


apcameron:
I can confirm that this PR works on my Tesla P40 which is a Pascal Based GPU with 24GB VRAM on my computer with 128GB RAM
I used the following
hf download openai/gpt-oss-120b     --local-dir models/openai/gpt-oss-120b
ft serve --model ./models/openai/gpt-oss-120b   --host 0.0.0.0 --port  5001


 ft shell --server 192.168.0.100:5001
FreeToken shell -> gpt-oss-120b @ http://192.168.0.100:5001  (/help for commands, /exit to quit)
> what is the value to PI to 100 digits                                         

Thinking...
| Thus π = 3.1415926535 8979323846 2643383279 5028841971 6939937510 5820974944 5
| 923078164 0628620899 8628034825 3421170679
| 
| We can present as a single string without spaces. Also note that the 100th dig
| it after decimal is 9? Let's verify: The 100th digit after decimal is the last
|  digit of the sequence above: 9. Yes.
| 
| Thus answer. Also maybe mention that it's irrational, etc. Provide reference.
End Think
Here is π (pi) written out to **100 decimal places**:

\[
\pi = 3.\!1415926535\;8979323846\;2643383279\;5028841971\;6939937510\;5820974944
\;5923078164\;0628620899\;8628034825\;3421170679
\]

(Spaces are added only for readability; the 100‑digit fractional part is the str
ing after the decimal point.)

**Why 100 digits?**  
π is an irrational, transcendental number, so its decimal expansion never termia
tes or repeats. Computing many digits is a classic test for numerical algorithm 
and computer hardware. The first 100 digits are often quoted for reference or fo
r high‑precision calculations in scientific work.
$
 
[done] gpt-oss-120b (medium) | ↓77 ↑497 4.7 tok/s | cache 1255 lru 27.2% | kv 57

erichstuntebeck:
This PR also fixes a **Turing (sm_75)** crash, which the title undersells — worth noting since
the PR reads as Pascal-only scope.

`decode_paged_attention` launches `_decode_grouped_stage1_kernel` with a hardcoded
`BLOCK_N=32` / `num_stages=2` and no shared-memory check, while the extend/prefill sibling has
been smem-aware via `_select_extend_tile` all along. On a 64 KiB opt-in device that overflows
once `next_power_of_2(head_dim) > 256`. `_select_decode_tile()` here is exactly what was
missing.

Verified on a Tesla T4, running upstream's own decode parametrization
(`test_decode_triton_attention_matches_reference`: `(256, 8, 3)` and `(512, 2, None)`) against
three trees, each with a fresh `TRITON_CACHE_DIR` and a never-warmed cache volume:

| tree | `head_dim=256, kv=8` | `head_dim=512, kv=2` |
|---|---|---|
| `main` @ `2757bb5` | OK | **FAIL** — `OutOfResources: Required: 81920, Hardware limit: 65536` |
| **#24** @ `35668da` ("support Turing GPUs") | OK | **FAIL** — identical, `Required: 81920` |
| **this PR** @ `c8e4c08` | OK | **OK** — `out(2, 16, 512) finite=True` |

That middle row is the surprising one: the PR named *"fix(cuda): support Turing GPUs"* does not
fix this, because its `attention.py` hunk only touches `_select_extend_tile` — the extend path.
`gh pr diff 24 | grep -i decode` returns nothing. Anyone triaging a Turing decode crash would
reasonably assume #24 covers it.

Boundary on `main` (sm_75), showing it is `next_power_of_2(head_dim)` that matters, not head_dim:

```
device Tesla T4 sm_75  smem_optin=65536
  OK    head_dim=128 (BLOCK_D=128) num_kv_heads=2
  OK    head_dim=256 (BLOCK_D=256) num_kv_heads=2
  FAIL  head_dim=320 (BLOCK_D=512) num_kv_heads=2  Required: 81920, limit: 65536
  FAIL  head_dim=384 (BLOCK_D=512) num_kv_heads=2  Required: 81920, limit: 65536
  FAIL  head_dim=512 (BLOCK_D=512) num_kv_heads=2  Required: 81920, limit: 65536
  FAIL  head_dim=512 (BLOCK_D=512) num_kv_heads=8  Required: 69632, limit: 65536
```

One detail that may be useful for the tile ladder: on this device `BLOCK_N` is the only lever
that matters. Taking `main` and changing **only** `num_stages=2 → 1` leaves the failure
byte-identical (still `Required: 81920`); changing **only** `BLOCK_N=32 → 16` makes it pass, and
matches upstream's `_reference_paged_attention` at upstream tolerance
(`max_abs_diff=1.562e-02`, `atol=rtol=2e-2`). Turing has no `cp.async`, so `num_stages` buys
nothing there regardless.

Practical impact: this is not an exotic geometry. `gemma-4-26B-A4B`'s full-attention layers are
`key_length=512, head_count=16, head_count_kv=2` — the `(512, 2)` row above — so serving that
checkpoint on any ≤64 KiB opt-in card dies in decode on `main` today. I have been carrying a
narrower local patch (`BLOCK_N` 32→16 below a 99 KiB cutoff); this PR's ladder is the better
form of it and I would rather it land than carry mine.

Reproduces **only** on devices with ≤64 KiB opt-in shared memory (sm_75 and below) — `81920`
fits comfortably under Ampere+/sm_89's ~99 KiB, so the same script passes on A100/H100/RTX 4090.

Environment: FreeToken `0.1.2` from source · Tesla T4 16 GiB sm_75 · driver 580.178.04 ·
CUDA 13.0.88 · torch 2.11.0+cu130 · triton 3.6.0 · Ubuntu 24.04.4.

CodingInCarhartts:
<html>
<body>
<html><head></head><body><p>I tested this on a GTX 1060 6GB, Pascal <code inline="">sm_61</code>, and can confirm <code inline="">openai/gpt-oss-20b</code> serves end to end with this branch.</p><p>Hardware was a GTX 1060 6GB with about 5.1 GiB free VRAM after the Windows desktop, i9-9900K, 32 GB RAM, Docker Desktop with WSL2. The container was running driver 582.28, CUDA 12.6, Python 3.12.14, torch 2.11.0+cu126, and Triton 3.6.0.</p><p>I tested the merge of <code inline="">gdevenyi:pascal-sm61-support</code> at <code inline="">c8e4c08e6548</code> plus <code inline="">pyproject-cu126-group</code> at <code inline="">fb2f80d9c52f</code> from #26, on top of <code inline="">main@0ab982f10905</code>.</p><p>Model was <code inline="">openai/gpt-oss-20b</code> using the MXFP4 safetensors with default context settings. KV auto-sized to 8205 tokens.</p><h3>Kernel tests</h3><p><code inline="">tests/kernels</code> finished with:</p><pre><code class="language-text">88 passed
15 skipped
1 failed</code></pre><p>One failure is:</p><pre><code class="language-text">test_minimax_m3_sparse.py::test_prefill_index_score_and_topk

Triton OutOfResources
Required: 98304
Hardware limit: 49152</code></pre><p>kernel needs 96 KiB of shared memory per block. Pascal only has 48 KiB, so it cannot fit on <code inline="">sm_61</code>.</p><p><code inline="">gpt-oss</code> does not use that path, so it does not prevent serving.</p><h3>Serving</h3><p>End-to-end serving works.</p><p>Auto-selection picked:</p><pre><code class="language-text">attention=triton
moe=offload
dtype=bfloat16</code></pre><p>CUDA graph capture also succeeded at batch sizes 1, 2, and 4.</p><p>The main thing I found is that <code inline="">--moe-backend hybrid</code> matters a lot on Pascal.</p><p><code inline="">ft bench bw</code> measured:</p><pre><code class="language-text">CPU STREAM:       45.6 GB/s
PCIe H2D:         12.6 GB/s
mxfp4 CPU-MoE:    20.9 GB/s
PCIe gather:       9.0 GB/s</code></pre><p>CPU-MoE is about 2.32x faster than PCIe gather </p><p>The problem is that <code inline="">auto</code> still selected offload because the bench profile was written inside the ephemeral container and did not persist between runs.</p><p>Measured difference:</p>
Metric | offload | hybrid
-- | -- | --
Startup prefill warmup [80,128] | 1228 s | 58 s
TTFT, first ~1.5k-token prompt | 1158 s | 46 s
TTFT, same prompt with warm cache | n/a | 1.3 s
Steady decode | 4.4 to 4.7 tok/s | 7.3 to 7.6 tok/s

<p> Offload made the first request look basically hung. Hybrid made the same setup usable.</p><p>VRAM after loading was:</p><pre><code class="language-text">5535 MiB / 6144 MiB</code></pre><p>That leaves about 0.4 GiB of headroom. Host RAM usage for the container was about 11.4 GiB.</p><p>With hybrid and a warm Triton cache, repeat TTFT landed around 1.3 to 2.9 seconds.</p><h3>fp16</h3><p>I also tried forcing:</p><pre><code class="language-text">--dtype float16</code></pre><p>That does not boot <code inline="">gpt-oss-20b</code> on this card.</p><p>It OOMs during weight materialization:</p><pre><code class="language-text">sharded_tensors -&gt; get_tensor
RuntimeError: CUDA driver error: out of memory</code></pre><p>Auto-selected bf16 loads successfully with the same amount of free VRAM.</p><p>I reproduced the fp16 OOM twice.</p><h3>PR #26</h3><p>One separate issue I hit while testing #26.</p><p>The group-scoped uv source does not appear to control torch's normal <code inline="">[project.dependencies]</code> edge.</p><p>Running:</p><pre><code class="language-bash">uv sync --no-default-groups --group cu126 --python 3.12</code></pre><p>resolved torch from PyPI instead of the cu126 index, which pulled the cu130 build. The nvcc/torch toolchain check then failed.</p><p>For testing I worked around it by pointing torch's source entry directly at the cu126 index.</p><h3>Reproduction</h3><p>Docker base:</p><pre><code class="language-text">nvidia/cuda:12.6.0-devel-ubuntu22.04</code></pre><p>The <code inline="">devel</code> image is required because <code inline="">_pinned_tensor</code> links against the CUDA runtime API and needs <code inline="">CUDA_HOME</code> and nvcc during the build.</p><p><code inline="">ninja</code> is also needed at runtime for the JIT index kernel.</p><p>Setup was:</p><pre><code class="language-bash">git clone &lt;fork&gt;
cd FreeToken
git checkout &lt;branch&gt;
git merge &lt;pr-26-branch&gt;

uv sync --no-default-groups --group cu126 --python 3.12
uv pip install pytest ninja</code></pre><p>For Pascal I would persist <code inline="">/root/.cache/freetoken</code>, run:</p><pre><code class="language-bash">ft bench bw</code></pre><p>once, then either let auto use the saved profile or explicitly serve with:</p><pre><code class="language-bash">ft serve --moe-backend hybrid</code></pre><p>All numbers above are from this one machine at temperature 0. Repeated runs produced byte-identical token counts.</p></body></html>
</body>
</html>

KodeMunkie:
I have 6x Tesla P4  (sm_61) in a single box, and an external P100 (sm_60) so I'll also test this PR as soon as I get chance (too excited for Qwen 3.8 Flash MoE at the moment, so that's taking priority).

gdevenyi:

**Result: 1209 passed, 372 skipped, **no new failures**.**

22 more tests skip here than on main: this PR adds that many CUDA-only tests the hidden-GPU run does **not** exercise. So the result above says it merges, imports and leaves every CPU-reachable path intact — it says nothing about the kernels themselves. I am running those on free cards in a maintenance window and will post the numbers here.

gdevenyi:

KodeMunkie:
It's a shame this was closed, I never got the time to test it however I did ask Qwen 3.8 27B to find the root cause of the out of memory, so this might be a possible direction for the fix in a future PR:

## Short answer

The OOM risk this PR introduces lives in `python/freetoken/kernel/triton/attention.py` — specifically the new tile-ladder fallback: `_select_extend_tiles()` + `_launch_first_fitting_tile()`. It turns "pick one tile, compile once" into "try several tiles, each of which gets fully compiled before Triton rejects it."

## The mechanism

Before this PR, `extend_paged_attention` picked a single `(BLOCK_M, BLOCK_N)` analytically from the device's opt-in shared memory and compiled exactly one kernel variant. The PR replaces that with:

1. `_select_extend_tiles()` returning a **descending candidate ladder** (up to 4 rungs: `(128,64) → (64,32) → (32,32) → (16,16)` for `head_dim ≤ 256`), because its own comment admits the byte estimate is only a *lower bound* — "what triton actually allocates depends on how it schedules the pipeline."
2. `_launch_first_fitting_tile()` catching `OutOfResources` and retrying the next rung.

The killer detail is in the helper's own docstring:

> ``OutOfResources`` is raised while triton sets up the launch — **after compiling the kernel**, before any GPU work is issued

So every rejected rung costs a **full Triton→LLVM compilation of one of the largest kernels in the repo** (`BLOCK_M` up to 128, `BLOCK_N` up to 64, `num_warps=8`, `D` up to 512). Those compiles are multi-GB host-RAM events. Multiply by:

- up to 4 candidate tiles × 2 variants (plain vs split-kernel when `k_extend/v_extend` present),
- every distinct shape/model that triggers warmup,
- parallel pytest workers or CI containers with cgroup RAM caps (both Pascal testers ran Docker/WSL2),

and you get repeated heavy LLVM compiles stacking up until host RAM runs out — i.e., the build/warmup OOMs. On a Pascal card with head_dim 256, even the *second* rung `(64,32)` needs ~64 KiB against a 48 KiB cap, so you're paying 2–3 full compiles of the big kernel just to land on `(16,16)`.

The fix direction would be to decide which tile fits *without* paying a full compile per rung (e.g. read Triton's smem accounting from one probe compile and reuse it, or cache the winning tile per `(device, head_dim, variant)` so only the first shape pays).

## Ruled out / context

- **Decode path**: `_select_decode_tile()` returns `(32, 2)` (~128 KiB at head_dim 512) whenever `smem_optin == 0` — that's a hard `OutOfResources` launch failure on small cards, not a host OOM. Same disease, different symptom.
- **fp16 OOM reported by CodingInCarhartts** (`sharded_tensors -> get_tensor`, GTX 1060): that's GPU VRAM math at weight materialization (dequantizing to fp16 on a 6 GB card), not caused by this diff — bf16/mxfp4 fit, fp16 doesn't.
- **gdevenyi's "65–95 tests fail with out of memory"**: environmental — both GPUs were serving a live model during the test run; he hid CUDA to get a deterministic baseline.
- The C++/CUDA hunks (`utils.cuh`, `index.cu`, `store.cu`, `fast_index_copy.cuh`, `pinned_tensor.cpp`) are pure macro gates and error-handling — no memory impact on the nvcc build.

PR #25, opened 2026-08-22
fix(checkpoint): FTW truncated shard raises OSError instead of silent garbage load

RuixiangMa:
### Problem
_pread_into in ftw.py silently broke out of its read loop on EOF (os.preadv returning 0), leaving the destination buffer partially filled with uninitialized garbage. A truncated FTW shard file (from interrupted conversion, disk full, or corruption) would load partial weights with no error anywhere, producing silently corrupted model outputs.

### Fix
- O_DIRECT path (_pread_into): break → raise OSError with bytes read/expected and offset.
- mmap fallback (FTWReader.read_into): added length check before slice assignment with shard file name diagnostics. Previously raised a cryptic ValueError.
- Unified exception type to OSError across both paths.
- Updated docstring to reflect current behavior.

cc @andy-yang-1 @jason-fxz


gdevenyi:

FTW truncated-shard error handling. Not merged: this deployment serves HF safetensors directly and never exercises the FTW reader, so I would have been testing nothing. The change reads correctly — turning a bare OSError into a diagnosable message is the right instinct, and it is the same class of problem as the pin failure in #56.

jason-fxz:
ftw.py looks good! Just drop the tests and ready to merge.

RuixiangMa:
> ftw.py looks good! Just drop the tests and ready to merge.

Done

andy-yang-1:
@RuixiangMa Great job!

PR #30, opened 2026-08-22
fix(deepseek_v4): resolve HF repo ids to local cache when serving

jamespa16:
## Summary

`ft serve --model deepseek-ai/DeepSeek-V4-Flash-0731` failed with "config.json is missing" (and, once that was fixed, a missing safetensors index) even though the checkpoint was fully present in the local HF cache (as symlinks, per the normal `huggingface_hub` cache layout).

**Root cause:** DeepSeek-V4-Flash needs an extra sidecar file (`inference/config.json`) beyond the standard HF `config.json`, since the reference `ModelArgs` fields (e.g. `compress_ratios`) aren't representable in `transformers`' `AutoConfig`. The DSV4 loader located this (and `model.safetensors.index.json`) with raw `os.path.join(model_path, ...)` + `os.path.exists`/`open()` calls against `model_path`.

That works when `--model` is a local directory, but when `--model` is a bare HF repo id (e.g. `deepseek-ai/DeepSeek-V4-Flash-0731`), `model_path` — and `hf_config._name_or_path`, which `parse_config` reads it from — stays the literal repo-id *string*, never resolved to the actual local snapshot directory under `~/.cache/huggingface/hub/...`. So the path join produced a nonexistent path relative to cwd instead of pointing at the cache, even though the files were sitting right there.

Every other model in the registry only reads fields already present on the `AutoConfig` object (so this never came up), and every other model's *weight* loader already resolves the repo id via `download_hf_weight()` before touching the filesystem — DSV4's config and weight loaders were the two places that skipped that resolution step.

## Changes

- `python/freetoken/models/deepseek_v4/args.py`: `_config_path` now resolves `inference/config.json` / `model_args.json` through `huggingface_hub.hf_hub_download` (offline-cache-aware) when `model_path` isn't a local directory, mirroring the pattern `utils/hf.py` already uses for `config.json`.
- `python/freetoken/models/deepseek_v4/weight.py`: `iter_weights` and `load_dsfp4_expert_sources` now resolve `model_path` via `download_hf_weight` before reading `model.safetensors.index.json` / opening shards, matching every other model's weight loader (`load_dsfp4_expert_sources_parallel` already did this indirectly via `iter_expert_tensors_parallel`).

## Test plan

- [x] Verified against a real local HF cache containing `deepseek-ai/DeepSeek-V4-Flash-0731`:
  - `hf_hub_download(repo_id=..., filename="inference/config.json", local_files_only=True)` resolves to the cached snapshot path with no network call.
  - `freetoken.models.deepseek_v4.args.load_args("deepseek-ai/DeepSeek-V4-Flash-0731", max_batch_size=1)` returns a populated `DeepseekV4Args` (43 layers, dim 4096, vocab 129280).
  - `freetoken.utils.download_hf_weight("deepseek-ai/DeepSeek-V4-Flash-0731")` resolves to the local snapshot dir, which contains `model.safetensors.index.json`.
  - `_weight_map()` on that resolved folder loads all 72,317 tensor entries from the index.
- [x] `ft serve --model deepseek-ai/DeepSeek-V4-Flash-0731` end-to-end (needs GPU hardware not available in this environment)

https://claude.ai/code/session_018DTUNGq4V1tujUvS8dqP4n

### DIFF
--- a/python/freetoken/models/deepseek_v4/args.py
+++ b/python/freetoken/models/deepseek_v4/args.py
@@ -80,14 +80,28 @@ class DeepseekV4Args:
 
 
 def _config_path(model_path: str) -> str:
-    """Locate the authors' ModelArgs JSON inside the checkpoint directory."""
-    candidates = [
-        os.path.join(model_path, "inference", "config.json"),
-        os.path.join(model_path, "model_args.json"),
-    ]
-    for path in candidates:
-        if os.path.exists(path):
-            return path
+    """Locate the authors' ModelArgs JSON inside the checkpoint directory.
+
+    ``model_path`` is either a local directory or a Hugging Face repo id (e.g. when
+    serving straight from ``--model deepseek-ai/DeepSeek-V4-Flash-0731``, unresolved
+    to a local snapshot dir). For the repo-id case, resolve each candidate filename
+    through the HF cache/hub the same way ``utils.hf`` does for ``config.json``.
+    """
+    filenames = [os.path.join("inference", "config.json"), "model_args.json"]
+    if os.path.isdir(model_path):
+        for filename in filenames:
+            path = os.path.join(model_path, filename)
+            if os.path.exists(path):
+                return path
+    else:
+        from huggingface_hub import hf_hub_download
+        from huggingface_hub.utils import EntryNotFoundError
+
+        for filename in filenames:
+            try:
+                return hf_hub_download(repo_id=model_path, filename=filename)
+            except EntryNotFoundError:
+                continue
     raise FileNotFoundError(
         f"No DeepSeek-V4 ModelArgs JSON found under {model_path} "
         f"(looked for inference/config.json)"
--- a/python/freetoken/models/deepseek_v4/weight.py
+++ b/python/freetoken/models/deepseek_v4/weight.py
@@ -20,6 +20,7 @@ import torch
 from tqdm import tqdm
 
 from freetoken.models.loader import drop_page_cache
+from freetoken.utils import download_hf_weight
 
 from .args import DeepseekV4Args, load_args
 
@@ -93,6 +94,7 @@ def iter_weights(
     if not include_non_moe:
         return
 
+    model_path = download_hf_weight(model_path)
     args = load_args(model_path, max_batch_size=1)
     reader = _ShardReader(model_path, _weight_map(model_path), device)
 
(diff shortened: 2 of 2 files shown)


gdevenyi:
Merged and tested alongside my #17, which fixes the same user-visible bug from the other end. They do not conflict, but a maintainer should probably pick one rather than take both — flagging so it is a deliberate choice.

| | #30 (this PR) | #17 |
|---|---|---|
| scope | DeepSeek-V4 only | every model, plus `ft shell` |
| where | `_config_path` + `iter_weights` | `prepare_server_args`, once |
| fetch | lazy, per file via `hf_hub_download` | whole snapshot up front |

**What this PR does better:** it never downloads more than it needs, and it fixes the DSV4 path even for callers that bypass `prepare_server_args` (the offline `LLM`, for instance, which my PR does not reach).

**What #17 does better:** it is generic — `ft serve --model org/name` was broken for *every* model, not just DSV4, because `prepare_server_args` only calls `snapshot_download` on the modelscope branch. It also keeps `served_model_name` and the tool-call/reasoning parser cascade reading the repo id rather than a snapshot hash, which matters because clients then address the model as `DeepSeek-V4-Flash-0731` instead of `7872f01b...`.

Taken together the overlap is harmless — with #17 applied, `model_path` is already a local directory by the time DSV4 sees it, so this PR's hub branch becomes dead code rather than doing anything wrong.

One thing worth knowing if you take this one alone: `download_hf_weight` uses `allow_patterns=["*.safetensors"]`, so it will never pull `inference/config.json` — which is why you need the separate `hf_hub_download` here. That is correct as written, but it means the two files come from the same snapshot via two different code paths, and a partially-populated cache could in principle give you weights from one revision and args from another. Pinning a revision, or resolving the snapshot once and deriving both from it, would close that.

Also worth guarding: `_config_path` now treats any non-directory string as a repo id. A mistyped local path (`/modles/dsv4`) will go to the Hub and come back as a download error rather than "no such directory". I hit exactly this in #17 — it broke 28 existing tests that pass anonymous paths like `/models/anon` — and fixed it by only attempting hub resolution for strings matching `^[A-Za-z0-9][\w.-]*/[A-Za-z0-9][\w.-]*$`, i.e. one slash, no leading separator. Same guard would apply cleanly here.

jamespa16:
Ah yes, I think I must have missed your PR. However I think it would make sense to merge the ideas, since the fix should probably address all models, but (in my opinion) lazy file downloading is worthwhile.

PR #37, opened 2026-08-22
bench(bw): repeat each measurement and withhold a verdict that straddles the threshold

gdevenyi:
## The problem

`ft bench bw` measures the CPU MoE kernel and the PCIe gather once each, divides them, and prints a backend verdict. The ratio decides whether a deployment should run `hybrid` or `offload`.

One measurement of each is not enough to support that decision. On this machine the CPU MoE figure for bf16 moves between about 61 and 70 GB/s across repeats of the same command. With a threshold at 2.0x, a true ratio near 2.0 will print `hybrid` on one run and `offload` on the next, and neither is wrong.

## What this changes

`--reps N` repeats each measurement and reports the median with its range. When the range straddles the threshold, the tool withholds the verdict instead of printing one, and says which side the median fell on.

The default is 3, so existing invocations get repeats without a flag change.

## Why it matters beyond tidiness

I used this on my own PR #45 and it changed the outcome. One pair of runs showed a 6.9% regression from that PR, and a different pair showed a 1.9% gain. With repeats the two arms plainly overlap, so the honest answer is that the change does nothing measurable on this machine. Without repeats I would have reported whichever pair I ran first.

## Testing

`tests/moe/test_benchbw_verdict.py` covers the verdict as a pure function: clear hybrid, clear offload, and a straddling range that must withhold. On `main` with this PR: 5 passed.

`tests/server` on `main` with this PR: 540 passed.

### DIFF
--- a/python/freetoken/moe/benchbw.py
+++ b/python/freetoken/moe/benchbw.py
@@ -588,6 +588,25 @@ def recommend(cpu_bw_gbs: float, pcie_bw_gbs: float, threshold: float = 2.0) -> 
     return "hybrid" if cpu_bw_gbs > threshold * pcie_bw_gbs else "offload"
 
 
+def verdict(cpu_runs: list[float], pcie_runs: list[float],
+            threshold: float = 2.0) -> tuple[str, bool, tuple[float, float]]:
+    """``(backend, confident, (ratio_lo, ratio_hi))`` from repeated measurements.
+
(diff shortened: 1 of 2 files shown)


gdevenyi:

**Result: 1210 passed, 350 skipped, **no new failures**.**

The +5 over main are this PR's own tests, and they ran (not skipped).

gdevenyi:

PR #38, opened 2026-08-22
bench(bw): decide hybrid-vs-offload on the contended pair, not the isolated one

gdevenyi:
## The problem

`ft bench bw` measures the CPU MoE kernel and the PCIe gather separately, each with the machine otherwise idle, then divides the two.

The `hybrid` backend runs both at the same time. They contend for the same memory controllers, so neither reaches its isolated figure. A ratio built from two isolated measurements describes a situation that never occurs in production.

## What this changes

The tool now also measures the pair running together and decides on that. It reports both, so the isolated numbers stay visible for tuning:

```
bf16   9.00 MB   65.9 GB/s   25.2 GB/s   2.61x  hybrid
   overlapped: CPU-MoE 71.5 + PCIe 14.8 GB/s -> hybrid fetches 17.2% of misses
```

The second line is what the verdict now uses.

## What it revealed

On this machine the isolated ratio says `hybrid` by a comfortable margin, 2.61x against a 2.0x threshold. Under contention the PCIe side drops to 14.8 GB/s, so hybrid can only fetch 17.2% of misses while the CPU handles the rest.

That matches what the deployment does. End to end, `hybrid` measured 3.9% slower than `offload` at one stream and 15.7% slower at eight. Those two figures come from the deployment branch, which carries several other changes, so treat them as corroboration rather than as a measurement of this PR. The isolated ratio recommended the slower backend; the overlapped measurement does not.

## Testing

This PR adds no test of its own. The overlapped measurement is timing code, so a value assertion would be flaky, but the verdict selection is testable as a pure function the way #37 tests its own. Tell me if you want that before merge.

### DIFF
--- a/python/freetoken/moe/benchbw.py
+++ b/python/freetoken/moe/benchbw.py
@@ -646,11 +646,13 @@ def _bench_format(fmt: str, wl: Workload, device: torch.device, threshold: float
 
     cpu_g, pcie_g = entry["cpu_moe_gbs"], entry["pcie_gather_gbs"]
     if cpu_g is not None and pcie_g:  # pcie_g truthy also rules out a div-by-zero
-        entry["ratio"] = round(cpu_g / pcie_g, 3)
-        entry["recommended"] = recommend(cpu_g, pcie_g, threshold)
-        # Both sides work standalone -> also measure them contended (concurrently). This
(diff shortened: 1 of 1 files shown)


gdevenyi:

gdevenyi:

PR #39, opened 2026-08-22
bench(bw): expose the two things the hybrid-vs-offload ratio cannot see

gdevenyi:
## The problem

The CPU-to-PCIe ratio answers one question: which backend is faster per byte. Two things it cannot see also decide what a deployment gets.

The first is the miss rate. The offload path only pays PCIe cost for experts that are not resident. A cache that holds most of the working set moves the crossover a long way, and the ratio does not know the cache size.

The second is the fixed cost of a CPU MoE step. Below some batch size, a step costs about the same regardless of how few bytes it moves, so the per-byte ratio overstates what the CPU path delivers.

## What this adds

Two measurements, reported alongside the existing ratio:

- `measure_pcie_gather_by_misses` sweeps the miss count and reports gather cost against it, so the crossover can be read off directly rather than assumed.
- `measure_cpu_moe_step_cost` reports the fixed per-step cost of the CPU path.

Neither changes the verdict. They are there so the number that does drive the verdict can be checked against the two effects it ignores.

## Testing

This PR adds no test of its own. Both additions are timing code, so asserting values would be flaky, but their shape is testable: gather cost should rise with miss count, and step cost should flatten below a batch threshold. Tell me if you want that before merge.

### DIFF
--- a/python/freetoken/moe/benchbw.py
+++ b/python/freetoken/moe/benchbw.py
@@ -41,7 +41,7 @@ import socket
 import statistics
 import threading
 import time
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from datetime import datetime, timezone
 from types import SimpleNamespace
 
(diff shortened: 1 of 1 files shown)


gdevenyi:
The paper (arXiv:2608.16157 §3.2) turns out to name exactly the gap this PR measures, which makes a better case for it than I made originally.

Equation 3 gives the two branch times:

> *T*_fill(*q*) ≈ *qS*/*B*_P,  *T*_cpu(*m−q*) ≈ (*m−q*)*S*/(*B*_H − *B*_P)

and Equation 4 balances them to *q\** ≈ *m* · *B*_P/*B*_H.

**`T_cpu` is purely proportional to the miss count — there is no constant term.** The `cpu_moe_step_cost` measurement in this PR is that missing constant: **0.08 ms fixed + 0.19 ms/expert** per layer per step for ds_fp4 on a 2× Xeon Gold 6526Y. Across 43 layers the fixed part is ~3.4 ms/step of toll that Eq. 3 cannot express, paid whether one expert misses or twenty.

That is also why §3.2's model predicts hybrid should win essentially always: as *B*_H → *B*_P the policy degenerates gracefully to pure cache fill, so the natural crossover sits at a bandwidth ratio of **1.0**, not the 2.0 threshold the implementation uses. The implementation's conservatism looks justified — I measured hybrid *losing* end to end at a ratio of 2.5× (offload 16.72 vs hybrid 16.07 tok/s single-stream, 67.11 vs 56.57 at 8 concurrent streams).

For completeness, adding the measured constant to Eq. 3 still does not reproduce that crossover. Break-even is `m* = t_fixed / (S · (1/B_P − 1/B_H))` = 0.10 / 0.31 ≈ **0.3 experts**, i.e. "always use hybrid", against measurements that say otherwise at realistic miss counts. Something beyond a per-layer constant is going on — the most likely candidate is that §3.2's

> The exposed layer latency is therefore the slower of the two concurrent branches

does not hold in practice, and the GPU partly *serialises* behind the CPU branch on the handshake rather than overlapping with it. If exposed latency is closer to a sum than a max, the whole balance changes.

That is a hypothesis, not a measurement, which is why this PR still ships the two quantities and no derived verdict. It does suggest where to look next: instrument the exposed-vs-sum question directly, since it would invalidate Eq. 4 rather than just correct it.

gdevenyi:

gdevenyi:
Closing this. It was produced with heavy AI assistance, and I would rather withdraw it than take it through review on that basis.

PR #40, opened 2026-08-22
bench(bw): option to allocate the synthetic banks the way the loaders do

gdevenyi:
## The problem

`ft bench bw` allocates its synthetic expert banks with a plain host allocation. The loaders allocate the real banks differently: they map the bank, fill it, and only then register it with CUDA.

Registration order changes the pages the gather reads from, so the benchmark can measure a gather that the deployment never performs.

## What this adds

`--production-banks` allocates the synthetic banks the way the loaders do, through the same `HostBank` path. The default keeps the current allocation, so existing numbers stay comparable.

This is a measurement-fidelity option, not a speedup. Its value is that a number produced with it can be trusted to describe the deployed path.

## Testing

This PR adds no test of its own. That the flag routes through `HostBank` is assertable without timing anything. Tell me if you want that before merge.

### DIFF
--- a/python/freetoken/moe/benchbw.py
+++ b/python/freetoken/moe/benchbw.py
@@ -312,6 +312,40 @@ def _synth_experts(E: int, expert_bytes: int) -> int:
     return min(E, max(1, _SYNTH_BANK_BUDGET // max(1, expert_bytes)))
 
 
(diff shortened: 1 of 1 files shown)


gdevenyi:

gdevenyi:
Closing this. It was produced with heavy AI assistance, and I would rather withdraw it than take it through review on that basis.

PR #45, opened 2026-08-22
perf(cpu-moe): opt-in W4A8 (VNNI) dot for ds_fp4

gdevenyi:
## What this adds

`ds_fp4` expert weights hold 4-bit values with one e8m0 scale per 32 elements. The current CPU dot product widens each weight to fp32 and multiplies in fp32.

This PR adds an opt-in W4A8 path. It quantizes the activation to int8 once per row, then uses `VPDPBUSD` (AVX-512 VNNI) to multiply the 4-bit weights against it. The path is off by default and turns on with `FREETOKEN_CPU_MOE_DSFP4_VNNI=1`. When the flag is off, or the CPU has no `avx512_vnni`, the existing fp32 dot runs unchanged.

## Measurements, and why I am not claiming a win

Measured on `main` with this PR alone. Two Xeon Gold 6526Y, 32 physical cores. `ft bench bw --dtype ds_fp4 --reps 9`, run four times, alternating the flag:

| | run 1 | run 2 | run 3 |
|---|---:|---:|---:|
| `VNNI=0` | 66.5 GB/s | 62.5 GB/s | 61.8 GB/s |
| `VNNI=1` | 61.9 GB/s | 64.6 GB/s | 62.1 GB/s |

The two arms overlap completely. **This PR buys nothing measurable on this machine.**

That result is consistent with an ISA sweep of the same kernel, where moving `avx2` to `avx512f` to `avx512bf16` gives 60.7, 64.9, then 67.0 GB/s. About 10% is all that any arithmetic change can buy here, because the kernel is limited by memory bandwidth rather than by multiply throughput. W4A8 saves arithmetic and then spends the saving on quantizing the activation.

One caveat about the test machine. It has four of eight memory channels populated per socket, which caps it at 265 GB/s of measured STREAM read instead of about 530. A machine with all channels filled moves the memory limit further out, and arithmetic work would then matter more. I cannot test that here.

## What I suggest

I am not asking for this to be merged on the strength of a speedup, because I did not measure one. Three options seem reasonable to me, and I do not have a strong preference:

1. Merge it as an opt-in path for machines with more memory bandwidth than this one, on the understanding that it is unmeasured there.
2. Hold it until someone can measure it on a fully populated machine.
3. Close it.

The variance reporting in #37 is what showed this. A single pair of runs looked like a 6.9% regression, and a different single pair looked like a 1.9% gain.

## Testing

`tests/moe/test_cpu_moe_dsfp4_vnni.py` checks the W4A8 dot against the fp32 dot on random banks, with the flag on and off. On `main` with this PR: 3 passed.

### DIFF
--- a/python/freetoken/kernel/csrc/cpu_moe/cpu_moe_ext.cpp
+++ b/python/freetoken/kernel/csrc/cpu_moe/cpu_moe_ext.cpp
@@ -1001,6 +1001,166 @@ void mxfp4_gemv_avx2(float* out, const uint8_t* blk, const uint8_t* scl, const b
 }
 #endif
 
+// ---------------------- ds_fp4 W4A8 (VNNI) ----------------------
+// Same trick as the nvfp4 W4A8 path above, and for the same reason: ds_fp4's bank is
+// row-major with K contiguous, so a VPDPBUSD reduction runs along the dot product
+// rather than across output columns (which is what rules the transposed mxfp4 bank
+// out). Nibbles decode to exact int8 through the kE2M1x2 LUT and accumulate against
+// int8 activations; only the activation quantization is approximate, exactly as for
+// nvfp4.
+//
+// Two differences from nvfp4: the weight scale is one e8m0 byte per **32** K (not
+// e4m3 per 16), and there is no per-row fp16 global. The activation quantizer stays
+// per-16 (``quant_i8_pg16``, shared with nvfp4), so within a 32-K weight block the two
+// 16-K halves carry their own activation scales and share one e8m0 weight scale --
+// i.e. for 16-K block b the weight scale lives at ``scale[b >> 1]``.
+using dsi8dot_fn = float (*)(const uint8_t*, const uint8_t*, const int8_t*, int,
+                             const float*, const float*);
+
+#if CPU_MOE_X86
+// Scales for four consecutive 16-K blocks starting at an even b: blocks b,b+1 share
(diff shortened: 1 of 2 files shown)


gdevenyi:
Independent support for the opt-in default here, from the paper (arXiv:2608.16157 §4.1), describing the CPU worker pool:

Bandwidth-bound is the design intent, not an accident of this machine — which is the whole reason an arithmetic optimisation measures ~0% here, and why it should stay behind a flag until someone has a part where the balance differs. The local ISA sweep puts a number on the ceiling: avx2 60.7 → avx512f 64.9 → avx512bf16 67.0 GB/s on the DSV4 geometry, so doubling the vector width buys ~7%, and nothing downstream of that can buy more.

§5.3 also shows the balance moving with the host, which is the case for carrying these paths at all:

> moving from the many-channel server to a dual-channel consumer desktop costs FreeToken 4% of its decode rate, while llama.cpp keeps only 80% of its rate as its CPU-resident experts starve on two DDR5 channels

The channel count changes how bandwidth-starved the CPU path is by a large factor. This machine has only 4 of 8 channels populated per socket (166 GB/s per socket against a possible 333), so it sits toward the starved end and is close to a worst case for any compute-side optimisation.

gdevenyi:

- **CPU-only** (GPUs hidden, deterministic): 1205 passed, 353 skipped, **0 failures new vs main**. Baseline main = 1205 passed / 0 failed.
- **GPU, cards free** (production stopped for a maintenance window, so no contention): 492 passed, 6 failed, **no failures that main does not already have**.

gdevenyi:

PR #46, opened 2026-08-22
perf(cpu-moe): dedup experts across the batch in pass 1

gdevenyi:
## What this adds

The CPU MoE decode kernel walks the routed experts one route at a time. When several tokens in a batch route to the same expert, it reads that expert's weights once per token.

This PR builds a deduplicated route list before pass 1 and reads each distinct expert once, accumulating into every token that selected it. The weights are the large operand, so the saving tracks how often experts repeat.

## Measurements

Measured on `main` with this PR alone, using the batch sweep from #81 as the measuring tool. Two Xeon Gold 6526Y, bf16 experts of 9.00 MB, 256 experts, top-6.

| batch | routes | distinct | reuse | dedup off | dedup on | speedup |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 8 | 8 | 1.00x | 1.09 ms | 0.98 ms | 1.11x |
| 8 | 64 | 52 | 1.23x | 7.95 ms | 6.55 ms | 1.21x |
| 16 | 128 | 82 | 1.56x | 13.67 ms | 9.70 ms | 1.41x |
| 32 | 256 | 112 | 2.29x | 22.95 ms | 13.18 ms | 1.74x |
| 64 | 512 | 126 | 4.06x | 39.49 ms | 19.10 ms | 2.07x |

Both arms replay the same routing draw, so the comparison uses identical work.

At batch 1 there is nothing to deduplicate. The 1.11x there is the deduplicated path being slightly cheaper to walk, not reuse.

## Why the numbers land where they do

Pass 1 reads `gate_up`, shaped `[2I, H]`. Pass 2 reads `down`, shaped `[H, I]`. Pass 1 is therefore about two thirds of the bytes. This PR deduplicates pass 1 only, so reuse alone predicts `1 / (2/3 / reuse + 1/3)`:

| batch | reuse | predicted from reuse | measured |
|---:|---:|---:|---:|
| 32 | 2.29x | 1.60x | 1.74x |
| 64 | 4.06x | 2.01x | 2.07x |

Measured is above predicted at both sizes, and the batch-1 row says why. At batch 1 nothing repeats, reuse is 1.00x, and the formula predicts no gain at all, yet the step still runs 1.11x faster. So the deduplicated walk is itself cheaper than the per-route walk, independent of reuse.

Folding that constant in gives 1.78x and 2.23x, which now sit slightly above the measured 1.74x and 2.07x. The measurements land between the two estimates. I read that as the reuse term carrying the effect, with a small fixed saving on top, rather than as either model being exact.

#47 deduplicates pass 2 as well and takes the remaining third.

## Cost and safety

The dedup list is built once per layer per step. `FREETOKEN_CPU_MOE_DEDUP=0` restores the old path for A/B testing.

## Testing

### DIFF
--- a/python/freetoken/kernel/csrc/cpu_moe/cpu_moe_ext.cpp
+++ b/python/freetoken/kernel/csrc/cpu_moe/cpu_moe_ext.cpp
@@ -1293,6 +1293,15 @@ struct CpuMoeExecutor {
   std::atomic<int64_t> prt_next{0};  // ds_fp4 intermediate fp8 round-trip phase
   int64_t p1_total = 0, p2_total = 0, prt_total = 0;
   int n_iblk = 0, n_hblk = 0;
+  // Expert dedup for pass 1 (see build_dedup). dd_route holds the (tok*top_k + k)
+  // route ids grouped by expert; dd_expert[j] owns dd_route[dd_start[j] ..
+  // dd_start[j+1]). n_dd == 0 means "not deduped, use the per-route work split".
+  std::vector<int> dd_expert, dd_start, dd_route, dd_hist;
+  int n_dd = 0;
+  // Read once per executor rather than once per process: a static would freeze the
+  // first value seen, which silently disables the A/B when both settings are
+  // exercised from one process.
+  bool dedup_enabled = true;
   std::atomic<int> done_count{0};
   std::atomic<int> bar_count{0};
   std::atomic<int> bar_sense{0};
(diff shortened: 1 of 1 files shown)


gdevenyi:
Re-reading the paper (arXiv:2608.16157) I want to head off a likely objection, because §4.1 already contains the word "deduplicates":

> For each MoE layer, one GPU kernel **deduplicates the routed experts**, classifies them against the residency table, derives the bandwidth-based fetch count *q*, selects eviction victims, and rewrites logical routed IDs into physical slot IDs or a special CPU-assignment flag.

That is a different dedup, on the other side of the boundary:

| | §4.1 device-side | this PR |
|---|---|---|
| what is deduped | which **unique experts to fetch** over PCIe | which **weight rows to read** from host DRAM |
| where | GPU control kernel | CPU MoE worker pool |
| saves | duplicate PCIe transfers | duplicate DRAM reads |

The CPU executor never sees the deduplicated set. §4.1 says the kernel "rewrites logical routed IDs into physical slot IDs **or a special CPU-assignment flag**", so what reaches the workers is still per-route: `t->ids[tok * top_k + k]`. Two tokens routed to the same expert arrive as two independent routes, and pass 1 streamed that expert's `gate_up` rows from DRAM once per route.

So the two compose rather than overlap — §4.1 stops the expert crossing PCIe twice, this stops it crossing the memory bus twice once it is already in host RAM.

This also lines up with the paper's own characterisation of the CPU path in the same section:

Which is exactly why the win here tracks the traffic model to within 2% and why it comes from loop order rather than any kernel change.

gdevenyi:

gdevenyi:

PR #47, opened 2026-08-22
perf(cpu-moe): dedup experts in pass 2 as well

gdevenyi:
## What this adds

#46 deduplicates the expert reads in pass 1. Pass 2 still reads `down` once per route.

This PR reuses the same deduplicated route list for pass 2. Pass 2 accumulates into a per-token output rather than reading a shared operand, so it needs a scatter step that #46 did not.

## Stacked on #46

This PR contains #46's commit. The two touch the same code in `cpu_moe_ext.cpp` and pass 2 reuses the route list that #46 builds, so they cannot be separated without duplicating that work. Review #46 first.

## Measurements

Measured on `main` with this PR alone, using the batch sweep from #81 as the measuring tool. Same machine and workload as #46, so the rows compare directly.

| batch | reuse | dedup off | dedup on | this PR | #46 alone |
|---:|---:|---:|---:|---:|---:|
| 1 | 1.00x | 1.01 ms | 0.95 ms | 1.07x | 1.11x |
| 8 | 1.23x | 7.89 ms | 5.81 ms | 1.36x | 1.21x |
| 16 | 1.56x | 13.70 ms | 8.25 ms | 1.66x | 1.41x |
| 32 | 2.29x | 23.33 ms | 9.81 ms | 2.38x | 1.74x |
| 64 | 4.06x | 39.61 ms | 9.91 ms | 4.00x | 2.07x |

At batch 64 the cost stops rising with the batch: 9.91 ms against 9.81 ms at batch 32, while the route count doubles from 256 to 512. Once both passes read each distinct expert once, the step costs what the 126 distinct experts cost, and adding more tokens that select the same experts is nearly free.

That is the point of the pair. #46 alone reaches 2.07x at batch 64 because it still pays full price for one third of the bytes. Removing that third takes the same workload to 4.00x.

## Testing

### DIFF
--- a/python/freetoken/kernel/csrc/cpu_moe/cpu_moe_ext.cpp
+++ b/python/freetoken/kernel/csrc/cpu_moe/cpu_moe_ext.cpp
@@ -1095,6 +1095,11 @@ struct MoeTask {
 //     expert's K dimension per node (banks are already per-row contiguous).
 constexpr int IBLK = 32;
 constexpr int HBLK = 32;
+// Deduped pass 2 gives one work item the whole H-block for *every* token, so it owns
+// those output rows outright and needs no cross-worker reduction. That costs work
+// items (n_hblk instead of tokens * n_hblk), so the block is smaller to keep the pool
+// fed: H/8 items is ~8 per worker at H=2048 on a 32-core part.
+constexpr int HBLK_DD = 8;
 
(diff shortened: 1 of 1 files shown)


gdevenyi:

gdevenyi:

PR #49, opened 2026-08-22
perf(cpu-moe): extend expert dedup to ds_fp4 and mxfp4

gdevenyi:
## What this adds

#46 and #47 deduplicate expert reads for bf16 experts. The `ds_fp4` and `mxfp4` paths have their own inner loops and did not get the same treatment.

This PR extends the deduplicated walk to both. Each format keeps its own block-scale handling; only the route walk changes.

## Stacked on #47 and #46

This PR contains both. All three edit the same route-walking code in `cpu_moe_ext.cpp`, and this one reuses the dedup list they build. Review #46, then #47, then this.

## Measurements

Measured on `main` with only these three PRs, using the batch sweep from #81. The bf16 path gives:

| batch | reuse | dedup off | dedup on | speedup |
|---:|---:|---:|---:|---:|
| 1 | 1.00x | 1.10 ms | 1.01 ms | 1.09x |
| 8 | 1.23x | 7.97 ms | 5.85 ms | 1.36x |
| 16 | 1.56x | 13.75 ms | 8.29 ms | 1.66x |
| 32 | 2.29x | 23.10 ms | 9.76 ms | 2.37x |
| 64 | 4.06x | 39.93 ms | 10.08 ms | 3.96x |

The bf16 rows match #47 within run-to-run spread, which is what should happen: this PR changes the fp4 paths, not bf16. Its value is that `ds_fp4` and `mxfp4` deployments get the same behaviour instead of falling back to the per-route walk.

## Testing

This PR adds no test of its own. The fp4 dedup paths are covered only by the existing correctness tests for those formats, which do not exercise a batch large enough to force reuse. A test that pins the fp4 dedup output against the per-route output at batch 32 would close that gap, and I am happy to add one if you want it before merge.

### DIFF
--- a/python/freetoken/kernel/csrc/cpu_moe/cpu_moe_ext.cpp
+++ b/python/freetoken/kernel/csrc/cpu_moe/cpu_moe_ext.cpp
@@ -1095,6 +1095,11 @@ struct MoeTask {
 //     expert's K dimension per node (banks are already per-row contiguous).
 constexpr int IBLK = 32;
 constexpr int HBLK = 32;
+// Deduped pass 2 gives one work item the whole H-block for *every* token, so it owns
+// those output rows outright and needs no cross-worker reduction. That costs work
+// items (n_hblk instead of tokens * n_hblk), so the block is smaller to keep the pool
+// fed: H/8 items is ~8 per worker at H=2048 on a 32-core part.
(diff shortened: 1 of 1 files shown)


gdevenyi:

gdevenyi:

PR #64, opened 2026-08-22
gemma4/gguf: accept a scalar attention.head_count_kv

### LINKED ISSUE #63: Gemma-4 E2B QAT GGUF fails to load: scalar attention.head_count_kv indexed as a per-layer list
### Summary

`ft serve` crashes while parsing the official Google Gemma-4 E2B QAT GGUF
([`google/gemma-4-E2B-it-qat-q4_0-gguf`](https://huggingface.co/google/gemma-4-E2B-it-qat-q4_0-gguf)).
`parse_gguf_config` assumes `gemma4.attention.head_count_kv` is a per-layer array, but this
checkpoint stores it as a single scalar, so the subscript raises `TypeError`.

### Reproduction

```bash
hf download google/gemma-4-E2B-it-qat-q4_0-gguf gemma-4-E2B_q4_0-it.gguf --local-dir ./gemma-4-E2B-q4_0
ft serve --model ./gemma-4-E2B-q4_0/gemma-4-E2B_q4_0-it.gguf
```

### Traceback

```
File "freetoken/engine/engine.py", line 297, in __init__
  _adjust_config(config)
File "freetoken/engine/engine.py", line 1093, in _adjust_config
  model_config = config.model_config
File "freetoken/engine/config.py", line 92, in model_config
  return parse_config(self.hf_config)
File "freetoken/models/gemma4/gguf.py", line 73, in parse_gguf_config
  swa_kv = int(kv_per_layer[swa_layer_ids[0]]) if swa_layer_ids else int(kv_per_layer[0])
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
TypeError: 'int' object is not subscriptable
```

The frontend then reports `Backend supervisor: TypeError: 'int' object is not subscriptable`
and shuts the API server down.

### Cause

The GGUF metadata for this checkpoint:

```
gemma4.block_count               = 35
gemma4.attention.head_count      = 8
gemma4.attention.head_count_kv   = 1          <-- scalar, not a list
gemma4.feed_forward_length       = [6144, 6144, ..., 12288, ...]   <-- list, as expected
gemma4.attention.sliding_window_pattern = [True, True, True, True, False, ...]
```

`head_count_kv` is uniform across all 35 layers here, and llama.cpp's writer emits a single
value rather than a 35-element array in that case. `models/gemma4/gguf.py:63` comments it as
`# per-layer list` and indexes it directly at lines 73–74.

### Suggested fix

Broadcast a scalar to per-layer before indexing:

```python
kv_per_layer = g("attention.head_count_kv")
if isinstance(kv_per_layer, (int, float)):
    kv_per_layer = [int(kv_per_layer)] * num_layers
```

With that patch applied locally, config parsing completes and startup proceeds past this point.

### Environment

- FreeToken 0.1.2 (PyPI), Python 3.12.3, Ubuntu 24.04 x86_64
- torch 2.11.0+cu126, triton 3.6.0 (installed without the `[accel]` extras)

Note this is a metadata-parsing failure in `_adjust_config`, before any device work,
so it should reproduce on any GPU.

Rhycecomley:
Fixes #63.

`parse_gguf_config` treats `gemma4.attention.head_count_kv` as a per-layer array and indexes
it directly. llama.cpp collapses that array to a single value when every block shares it,
which the official [`google/gemma-4-E2B-it-qat-q4_0-gguf`](https://huggingface.co/google/gemma-4-E2B-it-qat-q4_0-gguf)
checkpoint hits (`head_count_kv == 1` across all 35 blocks), so `ft serve` dies during
`_adjust_config` with:

```
TypeError: 'int' object is not subscriptable
```

This broadcasts a scalar to one entry per block before the SWA/full lookups. Arrays are
unaffected — the `isinstance` guard only fires on the scalar form.

### Verification

Before: `ft serve --model gemma-4-E2B_q4_0-it.gguf` raises the `TypeError` above and the
frontend shuts the API server down.

After: config parsing completes and startup proceeds past this point.

Note my machine can't finish a full generation run on this checkpoint for unrelated reasons
(4 GB VRAM), so I've verified the parse fix rather than end-to-end inference — worth a second
look from someone with a supported card.

### DIFF
--- a/python/freetoken/models/gemma4/gguf.py
+++ b/python/freetoken/models/gemma4/gguf.py
@@ -60,7 +60,12 @@ def parse_gguf_config(shim: "GgufConfigShim") -> ModelConfig:
     num_layers = int(g("block_count"))
     hidden = int(g("embedding_length"))
     num_qo_heads = int(g("attention.head_count"))
-    kv_per_layer = g("attention.head_count_kv")  # per-layer list
+    # Per-layer list, or a single value when every layer shares it: llama.cpp collapses
+    # a uniform array to a scalar, which the official gemma-4 E2B QAT q4_0 GGUF hits
+    # (head_count_kv == 1 for all 35 blocks). Broadcast so the indexing below holds.
+    kv_per_layer = g("attention.head_count_kv")
+    if isinstance(kv_per_layer, (int, float)):
+        kv_per_layer = [int(kv_per_layer)] * num_layers
     # True -> sliding-window (SWA) layer, False -> full attention.
     swa_pattern = [bool(x) for x in g("attention.sliding_window_pattern")]
     assert len(swa_pattern) == num_layers, "sliding_window_pattern length != block_count"


Rhycecomley:
Scope note: as detailed in https://github.com/FlashML-org/FreeToken/issues/63#issuecomment-5383063111, this makes a scalar `head_count_kv` parse, but it does **not** make the E2B checkpoint loadable — the next line hits a per-layer `feed_forward_length` that `ModelConfig.intermediate_size` can't represent. Please treat this purely as a robustness fix for valid-but-uniform GGUF metadata, not as E-series support.

mkornreich:
Does this relate to my PR? See https://github.com/FlashML-org/FreeToken/pull/59

PR #65, opened 2026-08-23
Apple Silicon Metal backend (ft serve on macOS arm64)

jasonkneen:
Screenshot 2026-08-23 at 22 07 16

## Summary
Runs FreeToken on Apple Silicon without CUDA. `ft serve` on Darwin routes to the Metal backend (MLX or llama.cpp) before torch is imported. `install.sh` takes the mlx-lm path on macOS arm64.

## Why
`ft serve` currently imports the CUDA launcher and crashes in a Metal venv (`ModuleNotFoundError: torch`). There are no macOS wheels on PyPI. This branch is the installable CLI path for M1–M4.

## Verified on this machine
- Native arm64 venv, mlx-lm 0.31.3, MLX GPU
- `ft serve --model mlx-community/Qwen3-0.6B-4bit` — `/health` (`maintenance: serving`), `/v1/models`, chat completions
- `tests/server/test_serve_macos.py`, `tests/server/test_metal_backend.py`, `tests/daemon/test_serve_command_platform.py` — 48 passed locally (no CUDA)

## Notes
- Does not add a desktop `.dmg`. The download site PR installs this CLI.
- Until this lands on `main`, macOS install falls back to `jasonkneen/FreeToken@feat/apple-metal-backend`.


gdevenyi:
Why would this benefit when METAL has unified memory?

jasonkneen:
> Why would this benefit when METAL has unified memory?

UMA removes the discrete VRAM/PCIe boundary, so the CUDA-style offload story isn’t the Metal benefit. The Metal backend uses MLX/llama.cpp; FreeToken adds serving, model lifecycle, API compatibility and shell tooling. UMA still has finite capacity and bandwidth, so working-set and sparse-execution choices still matter.

theaungmyatmoe:
somebody pls review and merge

smter6626:
I’m working on a local Reviewer–Executor orchestration framework that may provide a useful agentic workload for this Metal backend. The architecture uses the same local model with isolated Reviewer/Executor histories, explicit cross-session routing, tool-mediated repair cycles, and repository-backed context reconstruction rather than one continuously growing chat history. I’m also implementing a runnable local prototype around this design.

The framework and governance model are documented here: [structured-llm-execution-framework](https://github.com/smter6626/structured-llm-execution-framework/blob/main/structured-llm-execution-framework-en.md).

I’m interested in whether the Metal path is expected to preserve enough request/cache instrumentation to evaluate workloads such as repeated context reconstruction and alternating logical sessions. If useful, I can package the current loop into a reproducible benchmark and compare the Metal backend against an Ollama baseline on the same Apple Silicon machine.

Issue #62, opened 2026-08-22
Offloaded MoE ignores tensor parallelism: TP=2 is a regression, and the fix is not expert parallelism


gdevenyi:
## Summary

`--tensor-parallel-size 2` is currently a **regression** for offloaded MoE: measured 15.63 tok/s against 18.00 at TP=1 on DeepSeek-V4-Flash (2x RTX 6000 Ada, no NVLink). Two independent single-GPU servers give 2.05x instead, so the hardware scales fine — the engine does not.

I went looking for expert parallelism and concluded that **EP is the wrong fix, and the right one is already half-built.**

## Why TP=2 loses today

The resident MoE path shards by TP:

```python
intermediate_size_per_partition = div_even(intermediate_size, tp_size)
if allocate_experts:
    self._alloc_resident_experts(intermediate_size_per_partition)
```

`OffloadMoELayer` passes `allocate_experts=False` and takes its weights from `OffloadMoeCache`, which has **no TP awareness at all** — no `tp_size`/`tp_rank` in `offload_cache.py`, `expert_banks.py` or `models/deepseek_v4/moe.py`. Every rank therefore loads and caches whole, unsharded experts.

Confirmed at runtime: `moe_cache_size=2477` per rank at TP=2 versus 2559 at TP=1. The cache did not grow. Both ranks stream **the same experts over their own PCIe link**, so aggregate PCIe traffic doubles for identical work, and then the layer pays an all-reduce over PCIe. Only attention and dense shard usefully, and this model is MoE-dominated.

## Why not expert parallelism proper

Sharding *experts* across ranks (rank r owns `e % world == r`) needs all-to-all dispatch and combine per MoE layer. Three problems, in order of severity:

1. **CUDA-graph capture.** Decode graphs are captured at static shapes. All-to-all counts depend on routing, so they vary per step. Workable by padding to the worst case (`bs * top_k` tokens per rank), but that is real added machinery in the hot path.
2. **Load imbalance.** Real MoE routing is skewed; a rank owning hot experts stalls the others every layer.
3. **New collectives** in a codebase that currently only needs `all_reduce`.

## The fix that fits: TP-shard the offload banks

Split the intermediate dimension, exactly as the resident path already does. Rank r holds `I/tp` of every expert.

* **Cache doubles.** A slot is half the bytes, so the same VRAM holds ~2x as many distinct experts. Residency here goes from 2559/11008 = 23% to roughly 46%.
[...]

PR #70, opened 2026-08-23
[1/3] feat(dsv4): add tensor-parallel DeepSeek-V4 runtime

calvarado2004:
## Outcome

DeepSeek-V4-Flash can now run as a real tensor-parallel model in FreeToken instead of replicating model and expert storage on every rank. This is the memory and execution foundation required by the DSpark work in PRs 2 and 3.

On the validation system, the standalone milestone loads and serves `DeepSeek-V4-Flash-0731` with TP=4 across four 16 GiB RTX A4000 GPUs while keeping the routed expert banks in host memory.

## Stack and review boundary

This is **PR 1 of 3**:

| PR | Milestone | Branch | Commits introduced |
|---|---|---|---|
| **1/3 (this PR)** | DSV4 tensor-parallel runtime | `pr/dsv4-tp-runtime` | `7f0d8fc` |
| 2/3 | Exact DSpark draft/verify loop | `pr/dsv4-dspark-exact` | `423edd1..d42be9b` |
| 3/3 | Adaptive verification and device-resident sampling | `pr/dsv4-dspark-adaptive` | `13d4560..38d7b2f` |

This PR is intentionally independent of speculative decoding: **13 files, 680 additions, 95 deletions** relative to `main`. Reviewing it first keeps checkpoint ownership and collective placement separate from the much larger draft/verify state machine.

## Why this is needed

The original DSV4 runtime had single-rank model shapes and checkpoint-loading assumptions. Merely launching multiple ranks would still replicate the largest tensors and could produce mathematically incomplete rank-local MoE outputs.

The implementation therefore defines an explicit ownership rule for every DSV4 tensor, validates that rule before model construction, loads only the rank-owned checkpoint slices, and places collectives at the points where rank-local partials become complete model values.

The goal is not generic “split every large tensor” TP. DSV4's latent-KV attention, Lightning Indexer, recurrent compressors, routed experts, and quantization metadata have different ownership requirements. Those requirements are preserved rather than reshaped to make a requested TP size fit.

## Tensor ownership and collectives

| Component | TP layout | Completion operation / reason |
|---|---|---|
| Token embedding | Vocabulary rows | Masked local lookup, then one all-reduce so exactly one rank contributes each token row |
| LM head | Vocabulary rows | Local logits followed by rank-ordered all-gather back to `[rows, vocab]` |
| MLA `wq_b` | Column-parallel over query heads | Each rank owns a deterministic head block |
| MLA `wo_a` | Output groups | Each rank owns whole `o_groups`; partial groups are rejected at configuration time |
| MLA `wo_b` | Row-parallel | One all-reduce completes the attention output |
| Shared SwiGLU expert | `moe_inter_dim` | Produces a rank-local hidden-width partial |
| Routed FP4 experts | Same `moe_inter_dim` split | Added to the shared partial locally; both are completed by **one** MoE all-reduce |
| Latent KV path and paged KV pools | Replicated | MLA exposes one latent KV per token that every local head reads |
| Router, compressors, Lightning Indexer | Replicated | Every rank must make the same expert and sparse-block selection |
| Hyper-connection mixers | Replicated | Small control tensors whose decisions must remain identical across ranks |

The shared and routed MoE branches deliberately do not reduce independently. They are partial sums over the same output dimension, so they are added first and all-reduced once. Reducing each branch separately would add a collective per layer without changing the result.

## Checkpoint loading and memory correctness

- Add DSV4 TP helpers for `(rank, world_size)`, exact division, rank slicing, and up-front model validation.
- Slice weights and their quantization scale grids on the same axis. FP8 128×128 blocks and FP4 32-element scale blocks cannot be cut at arbitrary boundaries.
- Require TP to divide `o_groups`, attention heads, vocabulary, and the quantized MoE intermediate dimension. Incompatible sizes fail with a targeted configuration error before any layer is built.
- Stream each rank's non-expert checkpoint tensors instead of materializing a complete model on every GPU.
- Build each rank's routed FP4 host banks directly from its intermediate-dimension slices.
- Clone a shard into independent storage. A contiguous dim-0 `narrow()` can still be a view; retaining it kept the complete parent allocation alive and was measured to cost **5.1 GiB per GPU** on this checkpoint at TP=4.
- Keep TP=1 as the identity layout and loading path.

For the validation checkpoint, TP=4 distributes the routed host banks across ranks instead of holding four full copies; the ranks retain approximately the same aggregate expert data that a single-rank run requires.

## CPU-offload integration

The CPU MoE executor now follows the same rank-local expert layout as the GPU path. It also partitions the process's available physical cores across TP ranks, including affinity handling, so four ranks do not each create a full-machine worker pool and oversubscribe the host.

This matters for the later hybrid path: GPU cache hits/fetches and CPU overflow computation must produce compatible rank-local partials before the TP reduction.

## Suggested review order

|---|---|---|
| Ownership rules | `deepseek_v4/parallel.py`, `args.py` | Exact divisibility, quantization alignment, TP=1 identity |
| Model math | `model.py`, `attention.py`, `moe.py` | Vocabulary collectives, MLA head/group ownership, single MoE reduction |
| Checkpoint path | `weight.py` | Weight/scale co-slicing, independent shard storage, rank-local expert banks |
| Host execution | `moe/cpu_executor.py` | Rank-local intermediate width and CPU-core partitioning |
| Wiring and UX | `engine/config.py`, `engine/engine.py`, `server/args.py`, docs | CLI propagation and early validation |
| Tests | `tests/dsv4/test_dsv4_tensor_parallel.py` | Shape tiling, storage ownership, invalid-layout failure cases |

## Validation

Focused coverage includes:

- every sharded tensor tiles exactly one architectural axis;
- replicated tensors retain their full shape;
- vocabulary ranges cover the complete vocabulary exactly once;
- expert-bank slices tile `moe_inter_dim` with no gaps or overlap;
- shards own their storage instead of retaining the full parent allocation;
- invalid head-group or quantization-block splits fail loudly;
- model construction and checkpoint ownership remain correct at TP=1/2/4.

The completed three-PR stack passes the full DSV4 suite on the target server:

```text
192 passed in 36.14s
```

Runtime milestone:

```text
Model:     DeepSeek-V4-Flash-0731
Hardware:  4 × NVIDIA RTX A4000 16 GiB
Topology:  PCIe-only, two NUMA nodes
Mode:      TP=4, hybrid CPU/GPU routed-MoE execution
Result:    model loads, CUDA graphs capture, and OpenAI chat generation serves successfully
```

## Scope and non-goals

- This PR does **not** add speculative decoding or the checkpoint's `mtp.*` drafter.
- It does not shard latent KV or sparse-selection state; those values are architecturally shared across local heads.
- It does not silently pad invalid dimensions or quantization blocks.
- It is single-host TP; no multi-node transport policy is introduced here.

The next PR uses these ownership and collective contracts to add exact DSpark decoding without changing the TP model math established here.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -41,6 +41,8 @@ parsers all resolve automatically from the checkpoint and the GPU.
 | `--host` | 127.0.0.1 | Bind address |
 | `--port` | 1919 | Bind port |
 | `--gpu` | GPU 0 | GPU to run on: a UUID from `nvidia-smi -L` or an `nvidia-smi` index; see [below](#choosing-a-gpu) |
+| `--tensor-parallel-size`, `--tp-size` | 1 | GPUs to shard the model over; one scheduler process per rank, rank `i` on `cuda:i` |
+| `--distributed-timeout` | 1800 | Seconds a TP collective waits before failing; sized for the per-rank skew in weight-load time |
 | `--max-running-requests` | 4 | Max concurrently running requests |
 | `--max-output-tokens` | 32768 | Default output budget for requests that omit one |
 | `--max-seq-len-override` | from checkpoint | Max sequence length |
@@ -172,4 +174,3 @@ profile that `ft serve --moe-backend auto` and `--moe-hybrid-max-fetch -1` then 
 - What to measure: `--dtype`, `--model`, `--formats`, `--isa`.
 - `--threshold` (default 2.0) sets the call: recommend hybrid when CPU bandwidth beats PCIe
   by that factor.
-
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -82,6 +82,7 @@ See [models.md](models.md#moe-backends) for what each backend does.
 | Flag | Default | Meaning |
 |---|---|---|
 | `--moe-backend` | auto | `fused`/`offload`/`cpu`/`hybrid`; auto → offload, or hybrid with a `ft bench bw` profile |
+| `--expert-load` / `--expert-prefetch` | auto / 2 | Expert-bank reader; parallel prefetch controls whole shards queued ahead of placement (use 1 to reduce TP host-RAM peak) |
 | `--moe-cache-size` / `--moe-cache-rate` / `--moe-cache-auto` | auto | GPU expert-cache size as slots / fraction of all experts / sized from free VRAM (mutually exclusive; auto is enabled by default for offload-family backends) |
 | `--kv-reserve-tokens` | 8192 | KV token floor reserved before `--moe-cache-auto` fills experts |
 | `--moe-cpu-threads` | physical cores | CPU worker threads for the cpu/hybrid executor |
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -90,6 +90,10 @@ See [models.md](models.md#moe-backends) for what each backend does.
 | `--moe-prefill-hit-d2d` | off | Prefill: copy cache-hit experts device-side, stream only misses (CUDA >= 13) |
 | `--disable-moe-prefill-overlap` | overlap on | Disable the two-buffer prefill copy overlap |
 
+On hosts where OS-locked and CUDA-registered memory share one quota, combine
+`--moe-cpu-layers` with `FREETOKEN_SKIP_BANK_LOCK=1`. CPU-only layer banks then
+remain pageable instead of consuming the quota needed by GPU-fetch layers.
+
 ### API behaviour
 
 | Flag | Default | Meaning |
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -41,7 +41,7 @@ parsers all resolve automatically from the checkpoint and the GPU.
 | `--host` | 127.0.0.1 | Bind address |
 | `--port` | 1919 | Bind port |
 | `--gpu` | GPU 0 | GPU to run on: a UUID from `nvidia-smi -L` or an `nvidia-smi` index; see [below](#choosing-a-gpu) |
-| `--tensor-parallel-size`, `--tp-size` | 1 | GPUs to shard the model over; one scheduler process per rank, rank `i` on `cuda:i` |
+| `--tensor-parallel-size`, `--tp-size` | 1 | GPUs to shard the model over; one scheduler process per rank, mapped in `--gpu` order (default: rank `i` on `cuda:i`) |
 | `--distributed-timeout` | 1800 | Seconds a TP collective waits before failing; sized for the per-rank skew in weight-load time |
 | `--max-running-requests` | 4 | Max concurrently running requests |
 | `--max-output-tokens` | 32768 | Default output budget for requests that omit one |
--- a/docs/models.md
+++ b/docs/models.md
@@ -31,6 +31,27 @@ for them; other checkpoints of the same architectures work too.
   `offload`, upgraded to `hybrid` when a cached `ft bench bw` profile
   recommends it.
 
+## Tensor parallelism
+
+`ft serve --tensor-parallel-size N` shards the model over N GPUs on one host. One
+scheduler process runs per rank, and rank `i` uses `cuda:i`.
+
+DeepSeek-V4 shards:
+
+- MLA query and output heads (`wq_b` column-parallel, `wo_a` by output group,
+  `wo_b` row-parallel with one all-reduce per block)
+- the MoE intermediate dimension, for both the shared expert and the routed FP4
+  experts, so the host expert banks divide by N instead of being replicated
+- the vocabulary, for the embedding table and the output head
+
+It keeps replicated: the MLA latent KV path (`wkv` and the paged KV pools — every
+head reads the same latent KV), the compressors and the Lightning Indexer (every
+rank must select the same blocks), and the router.
+
+Constraints for DeepSeek-V4: N must divide `o_groups` (8), so N is 1, 2, 4 or 8.
+The KV pool is replicated, so its cost per GPU does not fall with N; the weights
+and the expert banks do.
+
 ## Notes
 
 - `ft checkpoint` conversion is optional — it pre-converts a checkpoint into
--- a/docs/models.md
+++ b/docs/models.md
@@ -53,6 +53,10 @@ Constraints for DeepSeek-V4: N must divide `o_groups` (8), so N is 1, 2, 4 or 8.
 The KV pool is replicated, so its cost per GPU does not fall with N; the weights
 and the expert banks do.
 
+DeepSeek-V4 TP currently loads the original safetensors checkpoint. An FTW
+conversion records the TP=1 dense and expert-bank layout and is rejected for
+`N > 1` instead of failing later with rank-local shape mismatches.
+
 ## Notes
 
 - `ft checkpoint` conversion is optional — it pre-converts a checkpoint into
--- a/docs/models.md
+++ b/docs/models.md
@@ -34,7 +34,8 @@ for them; other checkpoints of the same architectures work too.
 ## Tensor parallelism
 
 `ft serve --tensor-parallel-size N` shards the model over N GPUs on one host. One
-scheduler process runs per rank, and rank `i` uses `cuda:i`.
+scheduler process runs per rank. By default rank `i` uses `cuda:i`; an explicit
+`--gpu` list maps its `i`th entry to rank `i`.
 
 DeepSeek-V4 shards:
 
--- a/python/freetoken/engine/config.py
+++ b/python/freetoken/engine/config.py
@@ -72,7 +72,12 @@ class EngineConfig:
     # ratio default above. A runtime cache rebuild sets this (num_swa_pages) to pin the window
     # regardless of the full anchor; the ratio is the startup default and the fallback.
     swa_num_pages_override: int | None = None
-    distributed_timeout: float = 60.0
+    # Collective timeout for the TP process group. The FIRST collective is the post-load
+    # free-memory all-reduce, and ranks reach it minutes apart on a large MoE checkpoint
+    # (each reads its own expert shards, at its own speed), so a short timeout kills the
+    # server during a healthy startup rather than protecting it. Sized for that load skew;
+    # lower it with --distributed-timeout if a hung rank should fail faster.
+    distributed_timeout: float = 1800.0
     use_dummy_weight: bool = False
     use_pynccl: bool = True
     max_seq_len_override: int | None = None
--- a/python/freetoken/engine/config.py
+++ b/python/freetoken/engine/config.py
@@ -24,10 +24,14 @@ class EngineConfig:
     # NVFP4 routed-expert GEMM backend (--nvfp4-backend): auto|marlin|flashinfer|triton.
     nvfp4_backend: str = "triton"
     # Expert-bank host load (--expert-load): auto|serial|parallel. "auto" reads scattered
-    # experts in parallel but falls back to serial when free RAM can't cover the banks + the
-    # parallel reader's extra (non-reclaimable) whole-shard buffer; "serial" forces the
-    # low-memory reclaimable read; "parallel" forces the fast read.
+    # experts in parallel but falls back to serial when free RAM can't cover the banks +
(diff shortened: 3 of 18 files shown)


gdevenyi:
Benchmarked this on 2× RTX 6000 Ada (48 GiB each, PCIe Gen4 x16, **no NVLink**), 2× Xeon Gold 6526Y, 503 GiB host, serving `DeepSeek-V4-Flash-0731` with offloaded experts. **It does exactly what it says, and it is the largest single win I have measured on this box.**

Before this PR, TP replicated the expert banks per rank, so TP=2 was a *regression*:

| | `moe_cache_size` | host RAM | single stream |
|---|---|---|---|
| TP=1 | 2559 | 143 GiB | 18.06 tok/s |
| TP=2 **before** this PR | 2477 | 291 GiB | 15.63 tok/s (**−13%**) |
| TP=2 **with** this PR | **5622** | **156 GiB** | **39.25 tok/s (2.17×)** |

The cache more than doubles and host memory nearly halves, which is the mechanism you describe — "the host expert banks divide by N instead of being replicated" — landing precisely as claimed. Residency goes from 23% to 51% of the pool, and the second PCIe link starts carrying half the gather instead of a duplicate of it.

Under load it holds up: 8 concurrent streams give **94.0 tok/s aggregate against 66.5** at TP=1 (1.41×). It also beats the workaround I had been using — two independent single-GPU servers behind a load balancer — which managed 66.9 aggregate while consuming 2× host RAM. This PR obsoletes that recipe on every axis.

I had independently written up the same design as an issue (#62) before finding this PR, having measured the replication problem and traced it to `OffloadMoeCache` having no `tp_size`/`tp_rank` anywhere. Your `div_tp(..., multiple_of=128)` handles the divisibility constraint I flagged there (block-quantized formats need `I/tp` to stay a multiple of the scale block), and splitting `moe_inter_dim` rather than doing expert-parallel all-to-all avoids the CUDA-graph problem entirely, since shapes stay static. That is the right trade.

Test suite on the merged tree: 1606 passed, 10 skipped. The two failures are unrelated to this PR (`test_cpu_moe_q4_0`, pre-existing on `main`; and `test_e4m3_compat`, which fails on #48 in isolation).

Two notes rather than objections:

1. **NCCL is a hard prerequisite and the failure is opaque.** Without `libnccl-dev` installed, TP dies at `ld: cannot find -lnccl` from the JIT build of `pynccl.cu`, even though torch ships `libnccl.so.2` in site-packages. Not this PR's doing, but this PR is what makes people hit it. A fallback to torch's bundled copy, or a clearer message, would save someone an hour.
2. A server binds `PORT` **and** `PORT+1` (a `multiprocessing.spawn` child). Unrelated to TP, but it bit me while comparing multi-instance setups.

Strong +1 from me. This is the change that makes a second GPU worth having for offloaded MoE.

calvarado2004:
Thank you for the unusually thorough independent validation. The 2.17x single-stream gain, roughly 1.41x aggregate gain (94 versus 66.5 tok/s), larger KV cache, and lower host-RAM use on a 2x RTX 6000 Ada / 2-socket Xeon system validate both the TP execution path and the memory-sharding mechanism on hardware very different from the 4x A4000 development box. The full test result is also very helpful.

Both operational issues are actionable:

1. **NCCL discovery:** agreed that `ld: cannot find -lnccl` is too opaque. A runtime wheel may contain `libnccl.so.2` while the extension link step still lacks the unversioned development link expected by `-lnccl`. The build should either discover/use PyTorch's bundled NCCL library explicitly or fail early with a precise message naming the missing development link/package. This matters especially because the direct PyNCCL route is the intended fallback on PCIe TP systems.
2. **Second port:** agreed that the extra rendezvous port must be documented and configurable as a first-class setting. Our deployment script already assigns it explicitly (`RDZV_PORT`, normally API port + 1), but relying on an implicit adjacent port in the generic multi-instance case is surprising and collision-prone.

I will keep these as focused operational follow-ups rather than mixing them into the TP math/data-path changes in this PR.

Thanks again for testing scaling, concurrency, memory behavior, and the failure modes—not just the happy-path token rate. This is strong confirmation that the TP milestone is real.

calvarado2004:
Follow-up from the DSpark TP>1 hybrid CPU+GPU integration in #71.

The TP sharding work here remains the foundation of our approach, and the newest measurements strengthen that decision. On the 4-GPU, dual-NUMA workstation we now emit a placement/accounting summary immediately before API readiness:

- NUMA 0: ranks 0/1, 73.31 GiB host experts, 20/20 physical cores;
- NUMA 1: ranks 2/3, 73.31 GiB host experts, 20/20 physical cores;
- CPU-hosted expert storage: 146.62 GiB;
- GPU model allocation: 55.24 GiB across 4 GPUs;
- counted model split: 72.6% CPU / 27.4% GPU.

One nuance: this is logical ownership/accounting plus CPU affinity. Physical page residency still depends on allocator first-touch unless explicit mbind is added; we are keeping that distinction visible rather than overstating the NUMA guarantee.

We also replaced the previous symmetric TP collective in this path with direct standard NCCL. In a comparable 87-call profile, NCCL CUDA time dropped from 878.0 ms / 92.1% to 27.2 ms / 15.8%. The bottleneck is now repeated expert-cache copying (65.9 ms / 38.3%), which is a much healthier place to tune the DSpark paper's hybrid CPU/GPU balance.

Quality/performance remained stable: ~29.6 tok/s sustained on long coding, 33.07 tok/s observed peak on the repeated-route-aware experiment, clean 3,000-token code generation, valid required tool-call plus follow-up, and no repeated-token regression. Explicit reasoning_effort=high coding averaged 30.36 tok/s after warm-up with ~86.9% isolated acceptance.

Our recommendation is therefore to keep this TP-aware sharding design and build adaptive DSpark verification on top, while treating explicit NUMA page binding as a separately measurable follow-up rather than conflating it with rank/core affinity.

Detailed integration measurements: https://github.com/FlashML-org/FreeToken/pull/71#issuecomment-5387736914

nhannguyen2709:
I rebased this PR onto current `main` (`2757bb5`) and preserved Carlos's original commit/authorship. The conflict-free branch is:

https://github.com/nhannguyen2709/FreeToken/tree/codex/pr70-rebase

It contains the original TP runtime commit plus four focused follow-ups:

- `6151d7f` aligns the rebased server GPU mapping/parser/docs/tests with current `--gpu` behavior.
- `ec5481f` verifies exact TP expert-bank contents, not only shapes.
- `13abef7` rejects accidentally loading a TP=1 FTW layout under TP>1.
- `a9cdf58` lets explicitly CPU-hosted expert layers remain pageable when `mlock` and `cudaHostRegister` share a quota.

Focused validation is green: 56 tests passed (`tests/dsv4`, TP parser coverage, the pageable residency test), and `git diff --check` passes. A wider offload selection produced 79 passes and one environment-only failure because FlashInfer is not installed in this local venv.

I also exercised the branch with the real `DeepSeek-V4-Flash-0731` checkpoint on the exact ARC Kaggle envelope reproduced on Modal: 4x L4, 48 CPUs, 196 GiB RAM. The key TP claim holds in practice: the rank-local banks are ~34.27 GiB each / 137.06 GiB aggregate, rather than 548.25 GiB aggregate, and all 43 layers build in about 28 minutes at ~150 GiB peak host use.

That host exposed a separate deployment constraint: only 15.9375 GiB of CUDA-registered host memory is available per rank. The pageable-CPU-layer follow-up is what makes a quota-aware split possible there. The final API-readiness/throughput run is in progress; I will post the measured result separately rather than hold back the rebase handoff.

Because I cannot push to the original PR head and GitHub reports it as conflicting, the author can transplant the branch with `67cf589..a9cdf58` or use the fork branch directly.

nhannguyen2709:
Follow-up validation from the conflict-free continuation branch `nhannguyen2709/FreeToken:codex/pr70-rebase`:

- New head: `c066b38e1c9981c8c8adb0efb328049fe87aa8c0` (`feat(moe): bound parallel expert prefetch`).
- DSV4 now supports bounded parallel source-shard loading through `--expert-load parallel --expert-prefetch N`; the exact run used prefetch 1 to limit the four-rank transient.
- The automatic RAM gate accounts for the current shard plus prefetched whole shards.
- 56 focused tests pass; the wider selection has 79 passes and one environment-only FlashInfer import failure.

An exact 4x L4 / 48 CPU / 196 GiB run successfully reached API readiness in 1,778.751 s. Each GPU used 12,543 MiB with a 32,768-token total KV cache. Forced 256-token decode measured 2.867 / 11.475 / 14.266 aggregate output tok/s at concurrency 1 / 8 / 16.

This validates the TP-sharded bank contents and pageable/pinned residency split under the Kaggle hardware envelope. Parallel prefetch is safe, but it did not reduce the roughly 28-minute load phase: every TP rank currently reads all whole source shards, producing fourfold read amplification before selecting its rank-local slices. The next loading optimization should therefore be TP-aware ranged reads or a persistent TP4 bank format, not a deeper prefetch queue.

calvarado2004:
Thanks @nhannguyen2709 — I adopted the conflict-free continuation through c066b38, including the five focused follow-ups, with your authorship preserved.\n\nThe PR is now based on current main 2757bb5, and GitHub reports it clean/mergeable. I fetched that exact pushed commit on the Lenovo and ran the focused DSV4, parser, and offload suite: 80 passed.\n\nThe L4 read-amplification result makes sense. I agree the next load-time work should target TP-aware ranged reads or a persistent TP4 bank format, rather than a deeper prefetch queue. Thank you for the rebase, fixes, and unusually thorough hardware validation.

lwx25:
this branch not strong, can't to change cache(kv/moe)
$ ft ctl --base-url http://127.0.0.1:1902 cache  --moe 1500  --kv 262144
>HTTP 503: runtime rebuild unsupported under TP > 1

calvarado2004:
Rebased onto current `main` (`cac247a`). The new head is `e206163`, and GitHub reports it clean and mergeable.

What changed:

- Adapted the TP runtime to the refactor in #418: DSV4 now uses the shared TP-aware layer classes, and the ds_fp4 expert banks are sized from `MoEConfig.local_intermediate` so each rank holds only its slice.
- Kept @nhannguyen2709's follow-ups, including the bounded parallel expert prefetch (`--expert-prefetch`), now wired into the method-based expert reader.
- Fixed a host-memory regression in the rebased serial expert reader. It kept every shard open until the last layer, and a live `get_slice` handle kept each shard mapped, so the page-cache drop could not evict it. On a TP=4 load that left about 78 GB of checkpoint cache on top of the pinned banks. The reader now goes one shard at a time, reads only the rank's gate/up rows, and releases the mapping before dropping the cache. A test checks that no shard is still mapped when its cache is dropped.

Validation on 4x RTX A4000, dual-socket, 502 GB RAM, DeepSeek-V4-Flash-0731, TP=4:

- The DSV4 and parser tests pass, including the new reader test at TP=1, 2 and 4.
- The wider `tests/moe`, `tests/engine` and `tests/models` selection has the same 53 failures as unmodified `main` on the same host (6 more tests pass on this branch).
- A full load now keeps the checkpoint page cache at 0 to 6 GB throughout. Pinned banks settle near 147 GB across the four ranks, and the server comes up ready to serve.

calvarado2004:
@lwx25 Thanks for testing. That 503 isn't specific to this branch. Runtime cache rebuild (`ft ctl cache`) has been single-rank only in `main` since the initial release: the scheduler rejects it under TP > 1, because a safe TP rebuild needs every rank to drain, rebuild and recapture graphs in lockstep, with an agreed rollback if any rank fails (see the comment in `scheduler/scheduler.py`, where `CacheRebuildBackendMsg` is handled). #70 makes DeepSeek-V4 runnable at TP > 1, so the limit now shows up there.

For now, size the caches at startup: `--moe-cache-size N` (or `--moe-cache-rate`) for experts, and `--num-pages N` (or `--kv-reserve-tokens`) for KV. Note that 262144 KV tokens is well beyond what fits beside the expert cache on 16 GB cards at TP=4; we auto-size to about 61k tokens there. TP-aware runtime rebuild would be a separate change on top of this PR, and I'm happy to open an issue to track it if that's useful.

PR #88, opened 2026-08-23
Fix: minor issue of tqdm lock leak during termination with TP

xinze-zheng:
Use a process-local threading.RLock for scheduler-worker tqdm progress bars.

tqdm defaults to a multiprocessing lock, creating one POSIX semaphore per TP rank. FreeToken terminates workers before those locks are finalized, causing resource_tracker warnings. Only rank 0 renders progress, so cross-process locking is unnecessary. A thread lock provides the required in-process synchronization without allocating semaphores. thus removes the warning.
Example TP=4 warning:
```
/usr/lib/python3.10/multiprocessing/resource_tracker.py:224: UserWarning: resource_tracker: There appear to be 4 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```
Tested on Qwen2.5 and TinyLlama with TP=2 and 4.

### DIFF
--- a/python/freetoken/server/launch.py
+++ b/python/freetoken/server/launch.py
@@ -55,9 +55,19 @@ def _run_tokenize_worker(detach: bool, **kwargs) -> None:
     tokenize_worker(**kwargs)
 
 
+def _configure_worker_tqdm_lock() -> None:
+    import threading
+
+    from tqdm import tqdm
+
+    tqdm.set_lock(threading.RLock())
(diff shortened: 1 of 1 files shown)


gdevenyi:

Merged; the only conflict is the GPU-assignment block that landed in launch.py at the same place. Not verified beyond a clean shutdown of the TP=2 server.

Issue #79, opened 2026-08-23
FreeToken Roadmap (2026)


jason-fxz:
# FreeToken RoadMap - 2026

### New Model Support
- GLM-5.3-Flash @andy-yang-1  #332
- Qwen3.8-Flash-Next @jason-fxz
   - [x] pin PLE in RAM — landed in #257 
   - [x] offload PLE to Disk #311
- DeepSeek-V4.1-Flash

### Hardware Support
- macOS: a native Metal engine on Apple Silicon Macs.
- AMD: support for AMD GPUs via ROCm.
- DGX Spark: aarch64 wheels, sm_121 kernels, and a unified-memory mode.

### Multimodal
- Support Image input for vision-language models
  - Qwen3.5 / Qwen3.6 / Qwen3.8 family @jason-fxz 
  - Qwen3.8-Flash-Next
  - Gemma-4 family

### Multi-GPU
- DGX Spark: multi-GPU optimisation for linked DGX Spark units.
- Tensor parallelism (TP): serve one model across multiple GPUs on one machine.

### Quantization
- Quant Layer Refactor: 
   - #418 
   - #427
- GGUF: support GGUF checkpoints and their quantization types across model architectures.

### Speculative Decoding
- Speculative decoding, including MTP / DFlash / Dspark.

---
Any contribution is welcome. Join the [Developer Slack](https://join.slack.com/t/flashml/shared_invite/zt-3zpdh5j10-9dwTXrgLiqpVxizhA9KVbA) if you want to get involved.

---

TODO: This is still WIP.

dusty44:
Any thoughts about enabling experts to stream from NVME drives? with the cached experts optimised in three tiers vram, system ram then streamed from disk? I've built it in a dwarfstar derived engine and it can work shockingly well.

Aladincz:
Is there any chance that this technology could be usable and give some benefits to old macs running small LLM on CPU? Would it be technically applicable?

Headmaster218:
Interesting timing — I’ve independently been working on a fairly similar MoE serving problem on AMD RDNA3/Vulkan, mainly around expert caching/paging, heterogeneous CPU–GPU execution and large-model offloading.
It’s still very much WIP and I’m cleaning up the benchmarks/docs, but the current code is here: [[infr]](https://github.com/Headmaster218/infr)
Running on a single RX 7900 XTX + 64GB DDR4. Happy to compare notes on the AMD side.
Successfully running with 
[...]

PR #104, opened 2026-08-23
feat(models): support TP for qwen3_5_moe

RuixiangMa:
## Problem
support TP for Qwen3.5 MoE model in RoadMap https://github.com/FlashML-org/FreeToken/issues/79

Qwen3.5 MoE model (Qwen3.6-35B-A3B) only supported TP=1. The weight loader raised `NotImplementedError("qwen3_5_moe weight loading currently supports TP=1 only")`, and all linear layers used `LinearReplicated` (no sharding, no all-reduce).

### Weight Sharding 
Two sharding functions handle all weight types:

- **`_shard_tp`**: Simple chunk along a given dimension (for o_proj, embed_tokens, lm_head, A_log, dt_bias, expert down_proj).
- **`_shard_tp_parts`**: Per-part column-parallel sharding with optional GQA KV head replication. When `num_kv_heads < tp_size`, KV heads are replicated using `rank * num_heads // world_size` (matching `shard_tensor` in `loader.py`).

Qwen3.5's checkpoint stores pre-fused weights (qkv_proj, in_proj, gate_up_proj). Simple chunk on a fused weight gives incorrect results — e.g., chunking `[gate, up]` gives rank 0 the entire gate and rank 1 the entire up, instead of each rank getting half of both. The fix shards each sub-part independently:

### Model Layers

- **Attention** (`attention.py`): TP-local head counts, `_qkv_split` computed from local heads with `allow_replicate=True` for KV. `o_proj` changed from `make_replicated` to `make_row_parallel` (row-parallel + all-reduce). `.view()` → `.reshape()` for non-contiguous tensors after `torch.split`.

- **GDN** (`gdn.py`): TP-local dimensions for heads, key_dim, value_dim, conv_dim. `in_proj` uses full sizes for `LinearColParallelMerged` (which handles its own sharding). `conv1d`, `A_log`, `dt_bias` use TP-local sizes. `out_proj` changed from `make_replicated_quant` to `make_row_parallel_quant`. FP8 path also uses TP-local dims.

- **MoE** (`moe.py`): `OffloadMoELayer._maybe_all_reduce` is a no-op (expert banks are full on each rank in offload mode). `MoELayer` (resident/fused) retains its own all-reduce.

### UT

tests/models/test_qwen3_5_tp.py::test_shard_tp PASSED
tests/models/test_qwen3_5_tp.py::test_shard_tp_parts PASSED
tests/models/test_qwen3_5_tp.py::test_shard_tp_parts_replicate PASSED

### E2E

**Offload mode (default):**

ft serve --model Qwen/Qwen3.6-35B-A3B --tp-size 1/2/4

| TP  | Output | Status |
|---|---|---|
| 1 | "Here's a thinking process: 1. Analyze User Input: - Question: 'What is 2+3?'" | ✓ |
| 2 | Same | ✓ |
| 4| Same | ✓ |

**Resident (fused) mode:**

ft serve --model Qwen/Qwen3.6-35B-A3B --tp-size 1/2/4 --moe-backend fused

| TP  | Output | Status |
|---|---|---|
| 1 | Same as above | ✓ |
| 2| Same | ✓ |
| 4 | Same | ✓ |

### GPU Memory per Card 

| TP | Mode | Expert | KV cache | Non-expert(Attn+GDN+Shared et al.) |
|---|---|---|---|---|
| 1 | offload | 0 (host RAM) | 75.9 GiB | ~6 GB |
| 2 | offload | 0 | 78.5 GiB | ~3 GB |
| 4 | offload | 0 | 80.0 GiB | ~1.5 GB |
| 1 | Resident | ~60 GB | 19.0 GiB | ~6 GB |
| 2 | Resident | ~30 GB | 51.5 GiB | ~3 GB |
| 4 | Resident | ~15 GB | 68.0 GiB | ~1.5 GB |

CC @jason-fxz @andy-yang-1

### DIFF
--- a/python/freetoken/engine/engine.py
+++ b/python/freetoken/engine/engine.py
@@ -1331,9 +1331,12 @@ def _adjust_config(config: EngineConfig):
 
     if (
         is_moe
-        and expert_quant not in ("none", "fp8_block")
+        and expert_quant not in ("none", "fp8_block", "nvfp4")
         and not is_offload_moe_backend(config.moe_backend)
     ):
+        # The resident (fused) backend supports bf16 ("none"), block-fp8, and native NVFP4
+        # experts (TP-sharded on the intermediate dim); the other quant formats (mxfp4,
+        # q4_0, ds_fp4, ...) have no resident allocation and need the offload/cpu backends.
         raise ValueError(
             f"{expert_quant} experts require --moe-backend offload or cpu, "
             f"got {config.moe_backend!r}"
--- a/python/freetoken/kernel/triton/fp8_block_linear.py
+++ b/python/freetoken/kernel/triton/fp8_block_linear.py
@@ -294,23 +294,63 @@ class Fp8BlockLinear(BaseOP):
 
 
 class Fp8BlockColMerged(Fp8BlockLinear):
-    """Block-fp8 column-merged linear (drop-in for ``LinearColParallelMerged`` at TP=1).
+    """Block-fp8 column-merged linear (drop-in for ``LinearColParallelMerged``).
 
     Holds one fp8 weight that is the concatenation of several projections along the output
-    dim; the caller splits the bf16 output by ``output_sizes`` exactly as before. Each part's
-    out dim must be a multiple of 128 so the concatenated ``weight_scale_inv`` blocks align."""
+    dim; the caller splits the bf16 output by the (TP-local) part sizes. Each part's
+    out dim must be a multiple of 128 so the concatenated ``weight_scale_inv`` blocks align.
 
-    def __init__(self, in_features: int, output_sizes: list[int], has_bias: bool = False):
+    With ``local_output_sizes`` (TP>1) each part is sharded along the output dim
+    independently (GQA KV-head replication included) and ``weight_scale_inv`` is sharded
+    identically along its row axis."""
+
+    def __init__(self, in_features: int, output_sizes: list[int], has_bias: bool = False,
+                 local_output_sizes: list[int] | None = None):
         for o in output_sizes:
             assert o % _BLOCK == 0, (output_sizes, "each merged output size must be /128 for fp8")
         self.output_sizes = list(output_sizes)
-        super().__init__(in_features, sum(output_sizes), has_bias)
+        if local_output_sizes is not None:
+            assert len(local_output_sizes) == len(output_sizes), (
+                local_output_sizes, output_sizes
+            )
+            for o in local_output_sizes:
+                assert o % _BLOCK == 0, (local_output_sizes, "local sizes must be /128 for fp8")
+            out_features = sum(local_output_sizes)
+        else:
+            out_features = sum(output_sizes)
+        super().__init__(in_features, out_features, has_bias)
+
+
+class Fp8BlockRowParallel(Fp8BlockLinear):
+    """Row-parallel block-fp8 linear (drop-in for ``LinearRowParallel``): fp8 ``weight``
+    ``[out, in/tp]`` + ``weight_scale_inv`` ``[out//128, in/tp//128]``. Each rank reduces
+    over its input shard only (activations are quantized per 128-K group *within* the
+    shard, so the dequant stays exact per rank); the all-reduce completes the sum and the
+    bias (if any) is added once, after it."""
+
+    def __init__(self, in_features: int, out_features: int, has_bias: bool = False):
+        from freetoken.distributed import DistributedCommunicator, get_tp_info
+        from freetoken.utils import div_even
+
+        tp_info = get_tp_info()
+        self._tp_size = tp_info.size
+        self._comm = DistributedCommunicator() if tp_info.size > 1 else None
+        super().__init__(div_even(in_features, tp_info.size), out_features, has_bias)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        y = block_fp8_linear(x, self.weight, self.weight_scale_inv, None)
+        if self._tp_size > 1:
+            y = self._comm.all_reduce(y)
+        if self.bias is not None:
+            y = y + self.bias.to(y.dtype)
+        return y
 
 
 __all__ = [
     "FP8",
     "Fp8BlockLinear",
     "Fp8BlockColMerged",
+    "Fp8BlockRowParallel",
     "block_fp8_linear",
     "block_fp8_matmul",
     "per_token_group_quant_fp8",
--- a/python/freetoken/kernel/triton/fp8_pertensor_linear.py
+++ b/python/freetoken/kernel/triton/fp8_pertensor_linear.py
@@ -327,18 +327,61 @@ class Fp8PerTensorLinear(BaseOP):
 
 
 class Fp8PerTensorColMerged(Fp8PerTensorLinear):
-    """Column-merged per-tensor-FP8 linear (drop-in for ``LinearColParallelMerged`` at TP=1):
+    """Column-merged per-tensor-FP8 linear (drop-in for ``LinearColParallelMerged``):
     one fp8 weight concatenating several projections along the output dim; the caller splits
-    the bf16 output by ``output_sizes`` as before."""
+    the bf16 output by the (TP-local) part sizes.
 
-    def __init__(self, in_features: int, output_sizes: list[int], has_bias: bool = False):
+    With ``local_output_sizes`` (TP>1) each part is sharded along the output dim
+    independently -- including GQA KV-head replication when ``num_kv_heads < tp_size`` --
+    and the per-row ``weight_scale`` is sharded identically (it is piecewise-constant, one
+    scalar per part, so row sharding keeps every part's scale exact)."""
+
+    def __init__(self, in_features: int, output_sizes: list[int], has_bias: bool = False,
+                 local_output_sizes: list[int] | None = None):
         self.output_sizes = list(output_sizes)
-        super().__init__(in_features, sum(output_sizes), has_bias)
+        if local_output_sizes is not None:
+            assert len(local_output_sizes) == len(output_sizes), (
+                local_output_sizes, output_sizes
+            )
+            out_features = sum(local_output_sizes)
+        else:
+            out_features = sum(output_sizes)
+        super().__init__(in_features, out_features, has_bias)
+
+class Fp8PerTensorRowParallel(Fp8PerTensorLinear):
+    """Row-parallel per-tensor-FP8 linear (drop-in for ``LinearRowParallel``): fp8 ``weight``
+    ``[out, in/tp]`` + per-row ``weight_scale`` ``[out]`` (the output dim is not sharded).
+    Each rank reduces over its input shard only, so the result is a partial sum that the
+    all-reduce completes. The bias (if any) is added once, after the all-reduce.
+
+    ``input_scale`` is unchanged: every rank quantizes its own activation shard with the
+    same per-tensor scale, so the W8A8 path stays exact per rank."""
+
+    def __init__(self, in_features: int, out_features: int, has_bias: bool = False):
+        from freetoken.distributed import DistributedCommunicator, get_tp_info
+        from freetoken.utils import div_even
+
+        tp_info = get_tp_info()
+        self._tp_size = tp_info.size
+        self._comm = DistributedCommunicator() if tp_info.size > 1 else None
+        super().__init__(div_even(in_features, tp_info.size), out_features, has_bias)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        y = fp8_pertensor_linear(
+            x, self.weight, self.weight_scale, None,
+            self.input_scale, self._uniform_scale,
+        )
+        if self._tp_size > 1:
+            y = self._comm.all_reduce(y)
+        if self.bias is not None:
+            y = y + self.bias.to(y.dtype)
+        return y
 
 
 __all__ = [
     "FP8",
     "Fp8PerTensorLinear",
     "Fp8PerTensorColMerged",
+    "Fp8PerTensorRowParallel",
     "fp8_pertensor_linear",
 ]
--- a/python/freetoken/kernel/triton/nvfp4_linear.py
+++ b/python/freetoken/kernel/triton/nvfp4_linear.py
@@ -884,14 +884,55 @@ class Nvfp4DenseLinear(BaseOP):
 
 
 class Nvfp4DenseColMerged(Nvfp4DenseLinear):
-    """Column-merged NVFP4 dense linear (drop-in for ``LinearColParallelMerged`` at TP=1):
+    """Column-merged NVFP4 dense linear (drop-in for ``LinearColParallelMerged``):
     one packed weight concatenating several projections on the output dim; each part keeps its
     own per-row ``weight_global`` (and block scales), so the fused weight is exact. The caller
-    splits the output by ``output_sizes`` (e.g. shared-expert gate|up) as before."""
+    splits the output by the (TP-local) part sizes (e.g. shared-expert gate|up).
 
-    def __init__(self, in_features: int, output_sizes: list[int], has_bias: bool = False):
+    With ``local_output_sizes`` (TP>1) each part is sharded along the output dim
+    independently and the block scales / per-row globals are sharded identically."""
+
+    def __init__(self, in_features: int, output_sizes: list[int], has_bias: bool = False,
+                 local_output_sizes: list[int] | None = None):
         self.output_sizes = list(output_sizes)
-        super().__init__(in_features, sum(output_sizes), has_bias)
+        if local_output_sizes is not None:
+            assert len(local_output_sizes) == len(output_sizes), (
+                local_output_sizes, output_sizes
+            )
+            out_features = sum(local_output_sizes)
+        else:
+            out_features = sum(output_sizes)
+        super().__init__(in_features, out_features, has_bias)
+
+
+class Nvfp4DenseRowParallel(Nvfp4DenseLinear):
+    """Row-parallel NVFP4 dense linear (drop-in for ``LinearRowParallel``): packed ``weight``
+    ``[out, in/tp//2]`` + block ``weight_scale`` ``[out, in/tp//16]`` + per-row
+    ``weight_global`` ``[out]`` (the output dim is not sharded). Each rank reduces over its
+    input shard only; the all-reduce completes the sum and the bias (if any) is added once,
+    after it. The K-major repack at load is an exact permutation, so it works on the shard."""
+
+    def __init__(self, in_features: int, out_features: int, has_bias: bool = False):
+        from freetoken.distributed import DistributedCommunicator, get_tp_info
+        from freetoken.utils import div_even
+
+        tp_info = get_tp_info()
+        self._tp_size = tp_info.size
+        self._comm = DistributedCommunicator() if tp_info.size > 1 else None
+        super().__init__(div_even(in_features, tp_info.size), out_features, has_bias)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        if self._transposed:
+            y = nvfp4_dense_linear_t(
+                x, self.weight, self.weight_scale, self.weight_global, None
+            )
+        else:
+            y = nvfp4_dense_linear(x, self.weight, self.weight_scale, self.weight_global, None)
+        if self._tp_size > 1:
+            y = self._comm.all_reduce(y)
+        if self.bias is not None:
+            y = y + self.bias.to(y.dtype)
+        return y
 
 
 class Nvfp4LMHead(BaseOP):
@@ -937,5 +978,6 @@ __all__ = [
     "nvfp4_transpose_resident",
     "Nvfp4DenseLinear",
     "Nvfp4DenseColMerged",
+    "Nvfp4DenseRowParallel",
     "Nvfp4LMHead",
 ]
--- a/python/freetoken/layers/linear.py
+++ b/python/freetoken/layers/linear.py
@@ -59,10 +59,14 @@ class LinearColParallelMerged(_LinearTPImpl):
         input_size: int,
         output_sizes: List[int],
         has_bias: bool,
+        local_output_sizes: List[int] | None = None,
     ):
         # check that all output sizes are divisible by tp_size
         tp_info = get_tp_info()
-        tp_output_sizes = [div_even(size, tp_info.size) for size in output_sizes]
+        if local_output_sizes is not None:
+            tp_output_sizes = local_output_sizes
+        else:
+            tp_output_sizes = [div_even(size, tp_info.size) for size in output_sizes]
         output_size = sum(output_sizes)
         tp_output_size = sum(tp_output_sizes)
         super().__init__(input_size, output_size, input_size, tp_output_size, has_bias)
--- a/python/freetoken/layers/moe.py
+++ b/python/freetoken/layers/moe.py
@@ -77,6 +77,23 @@ class MoELayer(BaseOP):
             self.down_proj = torch.empty(n, h, i, dtype=FP8)
             self.down_scale_inv = torch.empty(n, h // blk, i // blk, dtype=torch.bfloat16)
             return
+        if self.weight_format == "nvfp4":
+            # Native ModelOpt NVFP4 rows (W4A16), TP-sharded on the intermediate dim:
+            # gate_up is column-parallel (shard the 2*I output rows), down is row-parallel
+            # (shard the I input cols); the down global scale keeps the full H output rows.
+            # The plain *_nvfp4 Triton kernels read these directly (inline dequant in the
+            # K-loop, no pre-tile, no BF16 copy) -- the resident mirror of the offload
+            # cache's "nvfp4" quant_format.
+            n, h = self.num_experts, self.hidden_size
+            i = intermediate_size_per_partition
+            fp8 = torch.float8_e4m3fn
+            self.gate_up_packed = torch.empty(n, 2 * i, h // 2, dtype=torch.uint8)
+            self.gate_up_scale = torch.empty(n, 2 * i, h // 16, dtype=fp8)
+            self.gate_up_global = torch.empty(n, 2 * i, dtype=torch.float16)
+            self.down_packed = torch.empty(n, h, i // 2, dtype=torch.uint8)
+            self.down_scale = torch.empty(n, h, i // 16, dtype=fp8)
+            self.down_global = torch.empty(n, h, dtype=torch.float16)
+            return
         assert self.weight_format == "bf16", (
             f"no resident expert allocation for weight_format {self.weight_format!r}"
         )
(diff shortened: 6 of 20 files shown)


alogotron:
TP=2 test results on consumer dual-GPU hardware (2x RTX 3090, sm_86, CUDA 13.0, driver 580.173.02, NCCL 2.31.2). Tested against PR head `3c8b281`.

## What works ✅

The TP=1 gate is gone and the weight sharding path executes. Both quantized checkpoints now start loading across both ranks — the original `qwen3_5_moe weight loading currently supports TP=1 only` no longer fires. The sharding math in `_shard_tp` / `_shard_tp_parts` looks correct.

## What fails ❌

Both quantized formats die at the **compute** layer, not the loading layer — the row-parallel quantized linear kernels are not implemented:

| Checkpoint | Error |
|---|---|
| `Qwen/Qwen3.6-35B-A3B-FP8` | `NotImplementedError: row-parallel Fp8BlockLinear not yet implemented` |
| `nvidia/Qwen3.6-35B-A3B-NVFP4` | `NotImplementedError: row-parallel Fp8PerTensorLinear not yet implemented` |

Stack location: `attention.py` / `gdn.py` call `make_row_parallel` / `make_row_parallel_quant`, which reaches the quantized linear classes that lack a TP forward path.

## Diagnosis

The PR tests (`tests/models/test_qwen3_5_tp.py`) cover the sharding math on plain tensors — no quantized E2E TP test. The claimed "E2E passes at TP 1/2/4" was likely verified on BF16 weights, which use the standard (already row-parallel-capable) linear path.

**Consequence: TP currently works only for BF16 weights, while the officially distributed checkpoints (FP8, NVFP4) remain single-GPU.** Since FreeToken's pitch is official quantized checkpoints without conversion, the quantized row-parallel kernels are the critical missing piece.

## Offer

Happy to test any follow-up commit on this branch. I can also run the BF16 checkpoint to confirm the full TP path works end-to-end there — will report back with TP=2 decode numbers vs the TP=1 baseline (~94 tok/s FP8 offload).

Environment: FreeToken 0.1.2 + PR branch via editable install, `freetoken_kernel_cache` 0.1.2+cu130, Ubuntu 24.04, Python 3.12.

RuixiangMa:
> TP=2 test results on consumer dual-GPU hardware (2x RTX 3090, sm_86, CUDA 13.0, driver 580.173.02, NCCL 2.31.2). Tested against PR head `3c8b281`.
> 
> ## What works ✅
> The TP=1 gate is gone and the weight sharding path executes. Both quantized checkpoints now start loading across both ranks — the original `qwen3_5_moe weight loading currently supports TP=1 only` no longer fires. The sharding math in `_shard_tp` / `_shard_tp_parts` looks correct.
> 
> ## What fails ❌
> Both quantized formats die at the **compute** layer, not the loading layer — the row-parallel quantized linear kernels are not implemented:
> 
> Checkpoint	Error
> `Qwen/Qwen3.6-35B-A3B-FP8`	`NotImplementedError: row-parallel Fp8BlockLinear not yet implemented`
> `nvidia/Qwen3.6-35B-A3B-NVFP4`	`NotImplementedError: row-parallel Fp8PerTensorLinear not yet implemented`
> Stack location: `attention.py` / `gdn.py` call `make_row_parallel` / `make_row_parallel_quant`, which reaches the quantized linear classes that lack a TP forward path.
> 
> ## Diagnosis
> The PR tests (`tests/models/test_qwen3_5_tp.py`) cover the sharding math on plain tensors — no quantized E2E TP test. The claimed "E2E passes at TP 1/2/4" was likely verified on BF16 weights, which use the standard (already row-parallel-capable) linear path.
> 
> **Consequence: TP currently works only for BF16 weights, while the officially distributed checkpoints (FP8, NVFP4) remain single-GPU.** Since FreeToken's pitch is official quantized checkpoints without conversion, the quantized row-parallel kernels are the critical missing piece.
> 
> ## Offer
> Happy to test any follow-up commit on this branch. I can also run the BF16 checkpoint to confirm the full TP path works end-to-end there — will report back with TP=2 decode numbers vs the TP=1 baseline (~94 tok/s FP8 offload).
> 
> Environment: FreeToken 0.1.2 + PR branch via editable install, `freetoken_kernel_cache` 0.1.2+cu130, Ubuntu 24.04, Python 3.12.

Yes， I only tested on Qwen3.6-35B-A3B（BF16）

CraigStone-Dev:
Hi — I picked up the remaining gap from this PR (row-parallel quantized linears + TP-sharded quantized loaders) and validated it end-to-end.

**What was added** (branch: https://github.com/CraigStone-Dev/FreeToken/tree/feat/qwen3-5-tp-support, opened as a PR against your branch: RuixiangMa/FreeToken#1):
- Row-parallel quantized linears: `Fp8PerTensorRowParallel` / `Fp8BlockRowParallel` / `Nvfp4DenseRowParallel` (sharded-K GEMM + all-reduce, bias after the reduce) — no new kernels needed, the existing Triton GEMMs are K-generic
- Column-merged quantized linears accept `local_output_sizes` (GQA KV-head replication); the `make_col_merged_quant` guard is gone
- `gdn.py`: `in_proj_qkvz` now declares TP-local sizes (the PR left it full-size while forward splits by local sizes — a latent TP>1 bug)
- Mixed-FP8 (NVFP4) and block-FP8 loaders ungated and TP-sharded
- 12 CPU unit tests + TP1/TP2 A/B E2E scripts

**E2E on 2x RTX PRO 4000 Blackwell (PCIe, `--disable-pynccl`) with `nvidia/Qwen3.6-35B-A3B-NVFP4`** (mixed FP8 attention + NVFP4 experts — the checkpoint this was targeting):
- TP=2 boots clean; CUDA-graph capture (bs 1/2/4) with NCCL all-reduce works
- KV capacity ~2.4x: 1,223,892 vs 517,045 tokens/rank (about 4-5 vs about 2 concurrent 256K contexts)
- Correctness A/B at temperature 0: 10/10 identical on the short suite; long-form suite semantically identical with only last-ulp drift from the all-reduce reduction order
- 3-way concurrent requests batch correctly

One operational note from testing: on a 54 GB-RAM box `--expert-load serial` is required (banks are 21.8 GB per rank, replicated per #62 -> ~44 GB pinned; the parallel O_DIRECT build OOM-kills rank 0).

**Disclosure:** this work was developed with an AI coding agent — the implementation and tests are AI-generated. All validation above was run on real hardware as described, and the numbers are from those runs.

Happy to push the commits directly to your branch instead of via PR, or restructure/split them however is most useful for review.

alogotron:
Follow-up validation on consumer dual-GPU (2x RTX 3090, sm_86, CUDA 13.0, NCCL 2.31.2), now covering the stacked fix branch `CraigStone-Dev:feat/qwen3-5-tp-support` (head `ab82a3f`, i.e. this PR + the row-parallel quantized linear completion from RuixiangMa/FreeToken#1).

## Regression found in this PR (base `3c8b281`) ⚠️

**PR #104 alone breaks the previously-working FP8 TP=1 config.** On stock 0.1.2, FP8 TP=1 serves fine (~94 tok/s decode on a single 3090). On this PR's head, the same TP=1 launch dies in `gdn.py:114` → `make_row_parallel_quant` with `NotImplementedError: row-parallel Fp8BlockLinear not yet implemented` — the row-parallel quant path is now taken unconditionally at TP=1 too. The stacked fix branch resolves this: TP=1 FP8 serves again at the same baseline throughput.

## TP=2 FP8 on the fix branch — one remaining gap ❌

The row-parallel quantized kernels (`Fp8BlockRowParallel` etc.) now construct and pass — the crash moved later, to weight loading. TP=2 dies on a shape assert in `layers/base.py:45` (`load_state_dict`) with the exact mismatch instrumented:

```
LOAD_MISMATCH lm_head weight (124160, 2048) (248320, 2048) torch.bfloat16 torch.bfloat16
```

Diagnosis: `lm_head` is created vocab-sharded at TP=2 (124160 = vocab 248320 / 2), but the weight loader yields the full unsharded (248320, 2048) tensor. The embedding/lm_head path needs the same shard-on-load treatment the attention/expert weights got (slice along the vocab dim per rank). All other weights loaded fine up to that point — this looks like the last blocker for quantized TP=2.

## Also verified

- BF16 checkpoint (`Qwen/Qwen3.6-35B-A3B`) sha256-verified against HF metadata: all 26 safetensors match. (Its TP=2 hybrid run is infeasible on 64 GB host RAM — the ~67 GB expert pool OOMs one rank during gloo all-reduce — so BF16 TP validation needs a bigger-RAM box; that's a capacity limit, not a code issue.)
- Environment: FreeToken 0.1.2 + editable branch install, `freetoken_kernel_cache` 0.1.2+cu130, Ubuntu 24.04, Python 3.12.

Happy to rerun TP=2 the moment an lm_head sharding fix lands — happy path looks very close now.

RuixiangMa:
> Follow-up validation on consumer dual-GPU (2x RTX 3090, sm_86, CUDA 13.0, NCCL 2.31.2), now covering the stacked fix branch `CraigStone-Dev:feat/qwen3-5-tp-support` (head `ab82a3f`, i.e. this PR + the row-parallel quantized linear completion from [RuixiangMa#1](https://github.com/RuixiangMa/FreeToken/pull/1)).
> 
> ## Regression found in this PR (base `3c8b281`) ⚠️
> **PR #104 alone breaks the previously-working FP8 TP=1 config.** On stock 0.1.2, FP8 TP=1 serves fine (~94 tok/s decode on a single 3090). On this PR's head, the same TP=1 launch dies in `gdn.py:114` → `make_row_parallel_quant` with `NotImplementedError: row-parallel Fp8BlockLinear not yet implemented` — the row-parallel quant path is now taken unconditionally at TP=1 too. The stacked fix branch resolves this: TP=1 FP8 serves again at the same baseline throughput.
> 
> ## TP=2 FP8 on the fix branch — one remaining gap ❌
> The row-parallel quantized kernels (`Fp8BlockRowParallel` etc.) now construct and pass — the crash moved later, to weight loading. TP=2 dies on a shape assert in `layers/base.py:45` (`load_state_dict`) with the exact mismatch instrumented:
> 
> ```
> LOAD_MISMATCH lm_head weight (124160, 2048) (248320, 2048) torch.bfloat16 torch.bfloat16
> ```
> 
> Diagnosis: `lm_head` is created vocab-sharded at TP=2 (124160 = vocab 248320 / 2), but the weight loader yields the full unsharded (248320, 2048) tensor. The embedding/lm_head path needs the same shard-on-load treatment the attention/expert weights got (slice along the vocab dim per rank). All other weights loaded fine up to that point — this looks like the last blocker for quantized TP=2.
> 
> ## Also verified
> * BF16 checkpoint (`Qwen/Qwen3.6-35B-A3B`) sha256-verified against HF metadata: all 26 safetensors match. (Its TP=2 hybrid run is infeasible on 64 GB host RAM — the ~67 GB expert pool OOMs one rank during gloo all-reduce — so BF16 TP validation needs a bigger-RAM box; that's a capacity limit, not a code issue.)
> * Environment: FreeToken 0.1.2 + editable branch install, `freetoken_kernel_cache` 0.1.2+cu130, Ubuntu 24.04, Python 3.12.
> 
> Happy to rerun TP=2 the moment an lm_head sharding fix lands — happy path looks very close now.

@CraigStone-Dev Can you please take a look? Thx!

CraigStone-Dev:
Fixed — pushed `6197d63` to the stacked branch (`CraigStone-Dev:feat/qwen3-5-tp-support`).

The block-fp8 pass (`_iter_weights_fp8`) sharded `embed_tokens` but not `lm_head`, so the full-width (248320, 2048) tensor was yielded against the vocab-sharded `ParallelLMHead` buffer. The per-tensor sharding is now in a small `_fp8_shard_standalone` helper (behavior unchanged for o_proj/out_proj, embed_tokens, A_log/dt_bias, conv1d) with `lm_head.weight` added to the vocab-parallel dim-0 branch. CPU test added (`test_fp8_shard_standalone_vocab_row_conv_tp2`); the full CPU suite passes.

@alogotron could you re-run the FP8 TP=2 on the new head? Note the NVFP4 mixed checkpoint is unaffected (its lm_head is the replicated `Nvfp4LMHead`) — that one is already validated E2E on 2x RTX PRO 4000.

CraigStone-Dev:
There might be a bug in that update PR. I'll try to do some local test even though I don't have the correct hardware to verify.

CraigStone-Dev:
There is another fix coming for the above. I will do a few more tests in the morning.  About 10 hours from now :)

alogotron:
TP=2 results on 2x RTX 3090 (24 GB each), i9-10980XE (18C), 62 GB DDR4-3600, PR head `f1ff8e2` + stacked `CraigStone-Dev:feat/qwen3-5-tp-support`. Flags: `--tensor-parallel-size 2 --attention-backend triton --expert-load serial --disable-pynccl --memory-ratio 0.7`.

**lm_head fix verified** — both FP8 and NVFP4 now load past the previous crash point. ✅

| Config | Decode tok/s | Note |
|---|---|---|
| FP8 TP=1 offload (0.1.2 baseline) | 94 | reference |
| NVFP4 TP=1 hybrid | 79 | reference |
| NVFP4 TP=2 hybrid | 2.2 → **9.5** with `--moe-cpu-threads 8` | see finding 1 |
| NVFP4 TP=2 offload | 3.1 | see finding 2 |
| FP8 TP=2 | OOM at load | 62 GB RAM insufficient for ~35 GB banks + runtime |

**KV capacity confirmed:** 524,288 tokens at TP=2 (vs 262K TP=1) — matches your 2.4x claim.

**Finding 1 — CPU executor core collision:** both ranks' hybrid executors pin to the same cores (`threads=17 (pinned to cores 0..16)` ×2 on 18 physical cores). Capping to 8 threads/rank gives 4.3x. Per-rank partitioning (rank0 → 0..8, rank1 → 9..17) should recover most of the remaining gap.

**Finding 2 — offload + no-pynccl serializes:** in offload mode the per-token PCIe expert streams and the host-staged all-reduce contend on the same bus; 3.1 tok/s vs 9.5 hybrid. On consumer PCIe without pynccl, hybrid is the only viable TP=2 path.

**Finding 3 — fused gate:** `--moe-backend fused` rejects NVFP4 ("nvfp4 experts require offload or cpu"). At TP=2 the sharded experts are ~11 GB/rank — GPU-resident would fit on 24 GB cards. That config would likely beat everything above.

Net for 2x 3090 users today: TP=2 trades decode speed for the 2.4x KV. FP8 TP=1 remains the speed champion.

CraigStone-Dev:
@alogotron following up on your 2026-08-27 TP=2 results — thank you for the validation. We've since pushed a fix for Finding 1, and we've reproduced and measured all three findings on two of our own machines (one a Proxmox VM, one bare metal).

**Finding 1 (CPU executor core collision) — fixed in `c23ec78`** (now the head of `CraigStone-Dev:feat/qwen3-5-tp-support`, so it's in this PR stack):
- `resolve_threads_and_affinity(rank, world_size)` gives each rank a disjoint round-robin slice of the physical cores (`reps[rank::world_size]`); an explicit `--moe-cpu-threads` count stays inside the rank's slice. Round-robin keeps a fair P/E-core mix on hybrid CPUs.
- The torch intra-op thread clamp now counts the rank's slice, not the whole machine.
- The "CPU MoE executor ready" line now logs on **all** ranks with `tp_rank=N/M` — it was `info_rank0` before, which is why the collision was invisible in the logs.
- CPU unit tests added (`tests/moe/test_cpu_executor_affinity.py`).

**Repro + verification** (300-token streaming decode, temperature 0, median of 3 runs; freetoken 0.1.2 + this branch; `nvidia/Qwen3.6-35B-A3B-NVFP4`; hybrid runs use `--tensor-parallel-size 2 --attention-backend triton --expert-load serial --disable-pynccl --memory-ratio 0.7`, the offload baseline uses `--moe-cache-size 4096`):

| Machine | Config | Decode tok/s |
|---|---|---|
| **Rudi — Proxmox VM** (PVE 9.2, `cpu: host`), i5-14500, 16 vCPU, 64 GB RAM, 2× RTX PRO 4000 Blackwell (PCIe, no NVLink) | offload (production flags) | 40.5 |
| | hybrid, pre-fix (both ranks pin cores 0..14) | 2.1 |
| | hybrid + `--moe-cpu-threads 8` | 4.0 |
| | hybrid + `c23ec78` | 27.4 |
| **Bandit — bare metal** (Linux Mint), i7-12700 (8P+4E hybrid), 62 GB RAM, 2× RTX PRO 4000 Blackwell (PCIe, no NVLink) | offload (production flags) | 27.6 |
| | hybrid, pre-fix | 2.6 |
| | hybrid + `--moe-cpu-threads 8` | 3.6 |
| | hybrid + `c23ec78` | **35.0** |

So the fix gives **13.5× (VM) / 13.6× (bare metal)** — consistent across virtualized and physical. VM caveat, since you'll see "Proxmox VM" above: the host i5-14500 exposes **no AVX512 on any core** (verified on the PVE host), so `cpu: host` is already at the hardware ISA limit (both boxes run the `avx2+vnni(nvfp4-w4a8)` kernel). The bare-metal cross-check exists precisely to rule out a VM ISA/scheduling handicap — there is none.

**Finding 2 (offload + no-pynccl serializes) — not a bug we can fix, but our data adds nuance:** the best backend is machine-dependent. On Bandit (bare metal) hybrid+fix (35.0) **beats** offload (27.6), matching your result; on Rudi (VM) offload (40.5) still beats hybrid+fix (27.4). So "hybrid is the only viable TP=2 path" holds on your box and on Bandit, but not universally.

**Finding 3 (fused rejects NVFP4) — still open.** We agree GPU-resident experts (~11 GB/rank) would fit 24 GB cards and would likely beat everything above. We're tracking it as an upstream issue candidate; if you can test `--moe-backend fused` with the NVFP4 checkpoint on your 3090 box (or point us at the gate), that would help.

**On the `--moe-cpu-threads 8` workaround:** on our boxes it only gave ~1.9× (2.1→4.0, 2.6→3.6) vs your 4.3× — both ranks still take the first N cores, and our CPUs are hybrid P/E (yours is a uniform i9-10980XE), so the residual overlap hurts more here. The partitioning fix removes the overlap entirely.

**Still open (not code-fixable by us):** FP8 TP=2 OOMs at load on a 62 GB box (your result matches our prediction for 54/62 GB boxes — needs #62 or a bigger box); the pynccl PCIe deadlock (unreported upstream; evidence + repro in our compose comment).

If you could re-verify `c23ec78` on your 2×3090 box, that would close the loop — especially the hybrid decode numbers and whether the per-rank pinning log lines look right.

**Disclosure:** this fix and these tests were developed with an AI coding agent; all numbers above are from real runs on the hardware described.

alogotron:
Re-verified on 2x RTX 3090 @ `c23ec78`.

**Pins:** `rank0 → cores 0..14`, `rank1 → cores 1..15` — disjoint round-robin slices, per-rank `tp_rank=N/M` logging works.

**Decode:** 32.7 tok/s client wall (219 tok completion), 59.3 tok/s engine-instantaneous — lands between your Rudi (27.4) and Bandit (35.0) numbers. 14.9x vs pre-fix on this box (2.2 → 32.7), consistent with your 13.5x.

**Finding 1 closed from our side.** ✅

One tester note: Qwen3.6 emits output in `reasoning_content` — with small `max_tokens` you can get empty `content` + `finish_reason=length` (all tokens spent on the thinking trace). Not an engine bug, but it confused our benchmark for a minute — may confuse other testers.

CraigStone-Dev:
@RuixiangMa — heads-up on follow-up TP work we've made since your last push.

The PR branch is at `3c8b281` (bf16-path TP). We've added **6 commits on top** covering the quantized-checkpoint paths, `lm_head` sharding fixes, and test coverage. They live on the mirror branch **`CraigStone-Dev/FreeToken:feat/qwen3-5-tp-support`** (head `c23ec78`), which is a strict descendant of `3c8b281`:

- `2eae9a8` feat(models): complete TP for qwen3_5_moe quantized checkpoints (FP8/NVFP4)
- `ab82a3f` test: add TP1/TP2 A/B verification scripts for served endpoints
- `6197d63` fix(models): shard lm_head vocab-parallel in the block-fp8 loader at TP>1
- `a77453f` test: tiny block-FP8 checkpoint generator + TP=2 loader regression test
- `f1ff8e2` fix(models): shard bf16 lm_head in the mixed loader; never shard NVFP4-native emits
- `c23ec78` fix(moe): partition CPU executor cores per TP rank

To pull them into the PR branch (fast-forward, no conflicts):

```
git remote add craig https://github.com/CraigStone-Dev/FreeToken.git
git fetch craig feat/qwen3-5-tp-support
git checkout feat/qwen3-5-tp-support
git merge --ff-only craig/feat/qwen3-5-tp-support
git push origin feat/qwen3-5-tp-support
```

The two `lm_head` commits are correctness fixes for the quantized paths (without them, TP>1 on FP8/NVFP4 checkpoints shards the wrong tensors), and `a77453f` adds a loader regression test that runs in the CPU test suite. Happy to walk through any of the diffs.

CraigStone-Dev:
@RuixiangMa — three more commits on the mirror branch **`CraigStone-Dev/FreeToken:feat/qwen3-5-tp-support`** (now head `b7e275a`, still a strict fast-forward of `3c8b281`):

- `86b0886` feat(moe): resident NVFP4 expert path in MoELayer (alloc + kernel dispatch)
- `2546787` feat(moe): build + yield resident TP-sharded NVFP4 expert banks
- `b7e275a` feat(moe): allow native NVFP4 experts on the resident (fused) backend

This closes **Finding 3** from the TP=2 thread ("fused rejects NVFP4"): `--moe-backend fused` (GPU-resident experts) now works with native NVFP4 checkpoints, TP-sharded. The gate was in `engine._adjust_config` (non-offload backends rejected for `expert_quant` outside none/fp8_block) plus a missing `weight_format="nvfp4"` case in `Qwen3_5MoE`; the NVFP4 Triton kernels already existed and are K-generic, so they compute correctly on the per-rank sharded intermediate — no new kernels. gate_up is column-parallel, down row-parallel, on the intermediate dim.

**E2E** (2× RTX PRO 4000 Blackwell, `nvidia/Qwen3.6-35B-A3B-NVFP4`, TP=2, `--moe-backend fused`): boots clean, coherent output, **141 tok/s** client wall (engine ~146), 22.4 GB VRAM/GPU — 5-9× the offload (~15) and hybrid (~27) TP=2 paths on the same box, since the experts are fully resident (no PCIe streaming) and CUDA graphs apply. CPU unit tests cover TP=1 full + TP=2 sharding. Note: for this model size the fused path is a TP≥2 feature (TP=1 doesn't fit: full expert bank ~22 GB + non-expert > 24 GB).

Same pull-in as before, now picking up all 9 commits:

```
git remote add craig https://github.com/CraigStone-Dev/FreeToken.git
git fetch craig feat/qwen3-5-tp-support
git checkout feat/qwen3-5-tp-support
git merge --ff-only craig/feat/qwen3-5-tp-support
git push origin feat/qwen3-5-tp-support
```

Happy to walk through any of the diffs. (Developed with an AI coding agent; all numbers from real runs on the hardware described.)

alogotron:
Re-verified on 2x RTX 3090 (Ampere) @ `b7e275a` — fused NVFP4 TP=2.

**Gate lift confirmed:** `--moe-backend fused` config resolves clean (no `NotImplementedError`), resident NVFP4 expert banks load TP-sharded, CUDA graphs captured bs=[1,2,4].

**Numbers** (`nvidia/Qwen3.6-35B-A3B-NVFP4`, TP=2, `--attention-backend triton --expert-load serial --disable-pynccl --memory-ratio 0.7`, 220-token decode, temp 0):

| Metric | 2x RTX 3090 | your 2x RTX PRO 4000 |
|---|---|---|
| Client wall | **139.8 tok/s** (219 tok in 1.57 s) | 141 |
| Engine decode | **165-166 tok/s** | ~146 |
| VRAM | 17.3 GB/GPU | 22.4 |
| KV capacity | **568,907 tokens** | — |

Reproduces on Ampere almost exactly. For reference: 1.5x the FP8 TP=1 offload baseline (94 tok/s) on this box — fused TP=2 is now the fastest config we have measured here, with the largest KV capacity as a bonus.

gdevenyi:
Tried this on a 2x RTX 6000 Ada (48 GiB, PCIe Gen4, no NVLink) box, CUDA 13.3 / torch 2.11 / sgl_kernel 0.4.5, cherry-picked onto main 86214a9 (applies cleanly, 8 files).

It regresses TP=1 for the block-FP8 checkpoint. `Qwen/Qwen3.6-35B-A3B-FP8` with `--tp-size 1` (both `--moe-backend fused` and `offload`) now dies at model build:

```
File ".../freetoken/models/qwen3_5_moe/gdn.py", line ..., in __init__
    self.out_proj = make_row_parallel_quant(
File ".../freetoken/models/quant_linear.py", line 63, in make_row_parallel_quant
    raise NotImplementedError("row-parallel Fp8BlockLinear not yet implemented")
```

`make_row_parallel_quant` raises for every quant format regardless of `tp_size`, and the GDN `out_proj` (and attention `o_proj` via `make_row_parallel`) now always go through it. Suggest returning the replicated quant linear when `get_tp_info().size == 1` (or implementing the row-parallel FP8/NVFP4 variants, which is what the `tp-size 2/4` case needs anyway for these checkpoints). I did not get a TP=2 number for that reason; the sibling qwen4_exp TP path I am working on shards the same way and I will report those numbers on the tracking issue.

Small one: `_shard_tp_parts` is a clean pattern; qwen4_exp needs the same six-way `in_proj` split ([q | k | v | z | b | a]) plus the QSA `qkv_proj` [q|gate per head | k | v] split, so it may be worth lifting `_shard_tp`/`_shard_tp_parts` into `models/loader.py` next to `shard_tensor`.

gdevenyi:
Follow-up with the qwen4_exp numbers promised above, since the sharding is the same shape (per-part column split of the fused `qkv_proj` / `in_proj`, kv-head replication rule `rank * heads // world`, row-parallel `o_proj` / `out_proj` / shared `down_proj`, vocab rows for embed/lm_head). Same box as above (2x RTX 6000 Ada 48 GiB over PCIe Gen4, no NVLink), `RadixArk/Qwen3.8-Flash-Next-NVFP4`, `--moe-backend offload --ple-backend pinned --num-tokens 262144 --memory-ratio 0.94 --moe-prefill-hit-d2d --max-running-requests 8 --cuda-graph-max-bs 8`, main 86214a9 + local qwen4_exp TP patch:

| | TP=1, one GPU | two TP=1 instances (one per GPU) | TP=2, one instance |
|---|---|---|---|
| single-stream decode | 59 tok/s | 56 tok/s each | 89 tok/s |
| 8 concurrent, aggregate | 137 tok/s | 266 tok/s (2 x 8) | 299 tok/s |
| TTFT, ~1k prompt | 2.1 s | 2.1 s | 1.55 s |
| expert residency | 37% | 37% | 100% |
| greedy probe (8 prompts) | 6/8 | | 6/8, answers identical to TP=1 |

Two things that drove it, and that differ from this PR: (1) the offload expert banks are sharded along the intermediate axis per rank (gate/up rows, down columns; for NVFP4 that is the packed codes, the 16-wide scale blocks and the per-row globals), so half of every expert fits on each 48 GiB card and the PCIe gather disappears entirely; keeping banks replicated (this PR's `_maybe_all_reduce` no-op) would cap that model at the same 37% residency per GPU. (2) With sharded banks the routed output is a partial sum; `routed + sigmoid(gate) * shared` is linear in the partials, so the shared expert's `LinearRowParallel` reduce and the routed reduce fold into one all-reduce per MoE layer. Over PCIe with no NVLink the two reductions per layer (attention + MoE) cost roughly 1.5 ms of the 11 ms step at bs=1, which is why it is +50% and not +100%.

Caveat for both: bf16 dense projections only. `make_row_parallel_quant` raising for fp8_block/nvfp4 is the same gap on my side; the checkpoints with quantized dense weights (the official Qwen3.8 FP8, the compressed-tensors NVFP4 ones) need row-parallel variants of `Fp8BlockLinear` / `Nvfp4DenseLinear` before they can run at TP>1.

CraigStone-Dev:
Following up on my earlier note here: I'd been building on @RuixiangMa's TP work in this PR — a 9-commit stack that completes quantized (FP8 / NVFP4) TP for `qwen3_5_moe` on top of it, at `CraigStone-Dev/FreeToken:feat/qwen3-5-tp-support` (head `b7e275a`). I was in the middle of rebasing that stack onto current `main` (to pick up the quantization refactor from #418/#426/#427/#428).

Given @gdevenyi's sharded-expert-bank TP approach discussed above looks like the cleaner design direction, I'm going to park the rebase rather than push a competing implementation, so I won't be opening a PR from that branch. The stacked branch is still available as a reference for the quantized-loader side (per-head dense sharding, block-fp8 / mixed checkpoint handling, resident NVFP4 expert banks) if it's useful.

Thanks @RuixiangMa for the base TP work.

PR #112, opened 2026-08-23
feat(moe): per-layer host-bank residency

jason-fxz:
## Why

#55: on WSL/Windows, WDDM caps pinned (page-locked) memory to ~50% of RAM -- MoE models whose
host bank sources exceed the cap cannot be served through the offload backend.

## Workaround

Per-layer host-bank residency: keep only the GPU layers' banks pinned; the remaining layers'
banks are OS-locked (`mlock`, no pin quota) and decode on the CPU MoE executor.

## Changes

- `--moe-cpu-layers` (id list / count / fraction)
  - now accepted under `--moe-backend hybrid` too (was offload-only)
  - on pin-capped hosts, the selected layers' banks are locked instead of pinned
  - unset = auto: lock just enough head+tail layers to fit the budget; `0` opts out
  - plain Linux (no cap): behavior unchanged, all banks stay pinned
- `FREETOKEN_PIN_BUDGET_GB`: override the pin budget (default: WSL -> 40% RAM, else uncapped)
- `--moe-backend cpu` over budget: degrade to all-locked instead of dying in cudaHostRegister
- Residency plumbing
  - FTW loader threads per-layer labels explicitly; raw loaders (safetensors/GGUF/model
    hooks) honor them ambiently via `requested_residency()` -- no loader signature changes
  - locked layers: prefill = synchronous whole-layer pageable copy (overlap disabled),
    copy plan skips their rows, decode routes to the CPU executor
  - mlock failure (RLIMIT_MEMLOCK): warn once, stay pageable, never abort the load
  - auto paths engage only when the CPU executor can serve the model (activation / `_cpu_moe`
    ABI / expert format); otherwise stay fully pinned on the GPU path

## Testing

- unit: residency validation, locked-layer materialize contract, copy-plan skip, settle routing
- H100 e2e, raw (gpt-oss-20b mxfp4, Qwen3.6-35B-A3B NVFP4) + FTW: locked bytes == `VmLck`
  in every configuration; FTW vs raw identical greedy outputs for the same split

### DIFF
--- a/python/freetoken/checkpoint/ftw.py
+++ b/python/freetoken/checkpoint/ftw.py
@@ -33,7 +33,6 @@ Layout on disk::
 from __future__ import annotations
 
 import json
-import logging
 import math
 import mmap
 import os
@@ -43,7 +42,9 @@ from concurrent.futures import ThreadPoolExecutor
 
 import torch
 
-logger = logging.getLogger(__name__)
+from freetoken.utils import init_logger
+
+logger = init_logger(__name__)
 
 INDEX_NAME = "freetoken_weight.json"
 FORMAT_TAG = "freetoken_weight"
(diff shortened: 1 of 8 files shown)


andy-yang-1:
@jason-fxz LGTM

Issue #43, opened 2026-08-22
Multi GPU detection and support


AshD:
I have a H200 + 4 x RTX 6000 Ada GPUs on a AMD Epyc server. 

Only the H200 is detected with 141GB of VRam

The other 4 GPUs don't show up in the console.

jason-fxz:
If you are using ft CLI, setting `CUDA_VISIBLE_DEVICE` should help

AshD:
Thanks for responding.

I am using the app. 
CUDA_VISIBLE_DEVICE is set to 0,1,2,3,4 
llama.cpp and python cuda apps like ComfyUI can see all the GPUs

jason-fxz:
Thanks for your feedback. We will support it on Freetoken Desktop soon — stay tuned!

jason-fxz:
Multi-GPU detection shipped in Desktop v0.2.0-beta.15: the sidebar lists every card, and Settings → GPU lets you pick which one the engine runs on.

Running one model across several GPUs is not exposed in the app yet; Closing this one as the detection part is done.

PR #117, opened 2026-08-24
feat(server)!: add --gpu to choose the GPU on multi-GPU machines

### LINKED ISSUE #99: Support for force selecting a specific GPU on multi-GPU systems
Not related to #43 directly but, essentially a way to be able to pick from a list of GPUs you have installed so you can selectively run on the one you want.

I have a 3080ti and a 3070, but the desktop app defaults to the 3070 no matter if I tell Windows to force use my 3080ti, or set `CUDA_VISIBLE_DEVICES` (my assumption is the mechanism for identifying GPUs does not expect more than 1 so just picks the first one it finds).

jason-fxz:
## Summary

`ft serve` always used `cuda:0`. This adds `--gpu` to pick the card:

```
ft serve --model ... --gpu GPU-2f3a9b1c   # UUID from nvidia-smi -L (a unique prefix is enough)
ft serve --model ... --gpu 1              # nvidia-smi index
```

`ft bench bw` and `ft checkpoint` take the same flag. The Desktop app uses it for its GPU picker.

Closes #99
Related #43 (fixed on the Desktop side),

## What changed

- `--gpu` is turned into `CUDA_VISIBLE_DEVICES` before the workers start. The engine itself is unchanged.
- An index means the nvidia-smi number, not CUDA's own order. If `CUDA_VISIBLE_DEVICES` is already set, `--gpu` can only pick from that list.
- A wrong GPU is reported before the model loads.
- `/v1/stats` shows the GPU in use under `gpus`.
- `ft bench bw` writes one profile per card (`benchbw/<gpu-uuid>.json`); the old `benchbw.json` still works.

## Breaking change

`ft checkpoint --device` is removed. Use `--gpu`.

## Tested on

8x RTX PRO 6000 Blackwell (96 GB), Xeon Platinum 8559C, Amazon Linux 2023, driver 595.71.05, CUDA 13.0, torch 2.11.0+cu130.
Checkpoint: Qwen3.8-27B BF16 as FTW

- `ft serve --gpu <uuid>`, `--gpu 6`, and `CUDA_VISIBLE_DEVICES=2,3 ... --gpu 1`: the process lands on the expected card in `nvidia-smi`, `/v1/stats.gpus` matches, chat completions work.
- Bad input (`GPU-deadbeef`, `99`, `1,2`, `""`, mixed UUID+index) is rejected with a clear error before loading.
- `ft bench bw --device 6` and `ft daemon start ... -- --gpu <uuid prefix>`: the per-card profile is written and `/bench/profile` returns the right card.
- `pytest tests/server tests/daemon tests/moe/test_hybrid_fetch.py`: 570 passed.

Not Tested on Windows or on a box with different GPU models.

### DIFF
--- a/benchmarks/bench_decode_moe.py
+++ b/benchmarks/bench_decode_moe.py
@@ -100,6 +100,8 @@ def parse_args(argv: list[str] | None = None) -> argparse.Namespace:
         help="hybrid: max PCIe fetches/layer; -1 = auto (benched pcie/cpu bandwidth fraction)",
     )
     p.add_argument("--mem-ratio", type=float, default=0.9, help="target VRAM utilization")
+    p.add_argument("--gpu", default=None,
+                   help="GPU for the serve: a UUID or nvidia-smi index (as ft serve --gpu)")
     p.add_argument("--no-graph", action="store_true", help="eager decode instead of CUDA graph")
     p.add_argument(
         "--greedy",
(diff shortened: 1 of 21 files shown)


andy-yang-1:
@jason-fxz LGTM. I think we can merge it now

PR #24, opened 2026-08-22
fix(cuda): support Turing GPUs

lukasrakauskas:
Adds NVIDIA Turing (`sm_75`) support for RTX 20-series GPUs:
- Disable optional kernels that require `sm_80+`.
- Use Turing-safe attention tile sizes.
- Add `sm_75` to kernel-cache builds.
- Add backend and attention tests.

Verified on RTX 2060 SUPER with the Qwen3.6 NVFP4 model. The API server starts and serves
requests successfully.

The full test suite ran with 1,313 passed, 25 skipped, and 39 GPU-specific failures caused by RTX 2060 SUPER (`sm_75`) limitations and unavailable optional backends.

PR #131, opened 2026-08-24
GGUF: support all quant types, add qwen35moe

vcruz305:
Two related things here: GGUF quant types beyond the three the Python side knew about, and qwen35moe support so llama.cpp MoE checkpoints load. I did the second one first and the first one fell out of it, so they share a branch. Happy to split if you'd rather review them separately.

Tested on an RTX 4060 Laptop (8GB) with Ornith-1.5-35B-A3B IQ3_S, a 16GB qwen35moe quant, experts offloaded to host RAM.

## Generic quant types

`csrc/gguf/` already dispatches 19 ggml types but the Python tables only described 6, so K-quants and I-quants were unreachable. `BLOCK_SHAPE` now covers all 21 types and there are five frozensets mirroring the C dispatch switches, each with the file:line it mirrors so they can be re-checked. `tests/models/test_gguf_type_tables.py` derives the block sizes from `ggml-common.h` at runtime and extracts the `case` labels from `gguf_kernel.cu`, so the tables can't drift from the kernels without a test failing.

One thing worth flagging on its own: none of the five `switch (type)` blocks in `gguf_kernel.cu` had a `default:`. `Y` is allocated with `torch::empty`, so an unsupported type returned uninitialized memory instead of raising, and `ggml_moe_get_block_size` returned 0. They all raise now. That's a latent bug independent of everything else in this PR.

I-quants have MMVQ and dequant kernels but no MMQ kernel, so `fused_mul_mat_gguf` routes them through `ggml_dequantize` plus a torch matmul at prefill. That branch already existed but was unreachable, since `_MMQ` was tested before `_DEQUANT` and the two sets were identical.

## qwen35moe

New `models/qwen3_5_moe/gguf.py` and `gguf_experts.py`, following the gemma4 pattern. The config comes from GGUF metadata and the head layout is derived rather than assumed: `attn_qkv`'s packed width of 8192 is exactly q(group_count * state_size) + k(group_count * state_size) + v(time_step_rank * state_size), and `parse_gguf_config` asserts the `ssm.inner_size` relation instead of trusting it.

Four things llama.cpp's converter does that the loader has to undo. I found these by reading `conversion/qwen.py` after a lot of wasted time, so listing them in case they save someone else the trouble:

- `ssm_a` holds `A = -exp(A_log)`, not `A_log`. Every value in the file is negative, spanning roughly -70 to -0.02. `gdn.py` computes `-A_log.exp()`, so passing `A` straight through makes the recurrent decay gate about -2.5e-5 and the recurrence dies. The loader yields `log(-A)`.
- The `(1 + w)` RMSNorm shift is already folded in by the converter. `weight.py` adds it for HF checkpoints because HF stores raw `w`, but a GGUF arrives pre-shifted and adding it again double counts. The measured means gave this away, around 1.0 rather than the 0.0X raw `w` would give.
- V heads are stored tiled rather than grouped when `num_k_heads != num_v_heads`, so `ggml_repeat` can replace an interleaved repeat. FreeToken pairs K head k with V heads `[k*R, (k+1)*R)` like HF does, so all seven affected tensors get un-tiled on load. Row permutations are done on the packed bytes directly, which is safe because each output row is its own run of blocks. `ssm_out` is the exception since its columns carry the V dimension and a 128 wide head straddles the 256 element blocks, so that one tensor is dequantized to dense bf16, about 500MB across 30 layers.
- Merged projections are not uniformly typed. In IQ3_S, `qkv_proj` is q and k IQ3_S with v Q4_K, and the GDN `in_proj` is `attn_qkv` Q4_K with the rest IQ3_S. Concatenating packed rows needs matching `row_bytes` (880 vs 1152 here), so there's a `GGUFMergedLinear` that keeps one buffer per part and concatenates the outputs instead. Same result as the merged GEMM since every part reads the same input. `gguf_merged_or_plain()` keeps the uniform case on the cheaper single kernel path.

None of these are visible to shape, dtype or byte-identity checks. The model loads, runs at full speed and produces fluent nonsense, which is a slow way to find out.

## Expert banks

`_BANK_SCHEMAS` gets a `"gguf"` entry, and the `(gate_up, down)` ggml types are threaded from `ModelConfig` through `ExpertBanks` and `OffloadMoeCache` to the kernels, because a GGUF's row stride is a property of the file rather than the format. `fused_experts_gguf` takes a separate `down_quant_type` since the two banks have separate slot pools and can legitimately differ.

What is not supported is a bank whose type varies by layer. `moe_vec.cuh` addresses the pool as `expert * nrows * (ncols / qk)` with no padding allowance, and the pool is one allocation shared by every layer, so two strides in one pool would read every block at the wrong offset. `_gguf_banks` raises and names the offending layers. In practice this means llama.cpp's mixed levels don't load: for Ornith-1.5-35B, IQ3_S and IQ3_XXS have uniform banks while IQ3_M, IQ2_M, IQ2_XXS and IQ1_S split `ffn_down_exps` across two types. Quantizing with `--pure` avoids it. Per type slot pools would fix it properly but that's surgery on the hot path and I didn't want to bundle it here.

`_model_setup_override` now declines for `expert_quant == "gguf"`, otherwise qwen3_5_moe's NVFP4 `setup_offload_expert_banks` hijacks the GGUF load and fails deep inside an NVFP4 reader.

## Tokenizer

`convert_gguf_tokenizer` is keyed on transformers' own model_type, so passing the GGUF arch string through raised `KeyError('qwen35moe')` and killed the detokenizer worker during load. qwen35moe is a GPT2 style BPE so the existing qwen2 converter handles it. Two neighbouring gemma4 assumptions are now per-arch: eos preferred the literal `<turn|>` and the stop ids were hardcoded to `<eos>`/`<turn|>`, neither of which exists in a Qwen vocab, and `unk` defaulted to a `<unk>` token that also isn't there. gemma4 behaviour is unchanged.

## Numbers

```
load                65s (serial GGUF read)
VRAM                6879 / 8188 MiB
host RAM            20GB pinned expert banks
decode              46.7 to 50.1 tok/s
end to end          29.4 tok/s including prefill
llama.cpp CPU ref   11.1 tok/s
```

Correctness checks I ran, since "it produces text" isn't much of a claim:

- 573 weights reconciled against the constructed module: every name exists, every shape and dtype matches, no non-expert buffer left unfilled.
- Expert banks checked byte for byte against their source rows, per expert, at both ends and the middle. This caught the `gate_up` fusion laying down every expert's gate before any up.
- MMVQ, MMQ and the grouped MoE GEMV checked against dequant plus matmul on real tensors from the file, cos 0.9999 or better for IQ3_S, Q4_K and Q6_K.
- 8/8 on short factual questions with greedy decoding, including arithmetic.

First token argmax matches llama.cpp CPU on 2 of 6 single token prompts. Every disagreement is a plausible near tie (`Paris` continuing " de" against " is", `import` continuing " java" against " sys"). FreeToken runs CUDA W4A8 MMVQ and llama.cpp runs CPU AVX, so I don't think exact agreement is reachable, but I'd rather state the number than imply it's bit exact.

## Known gaps

- TP=1 only, same as the gemma4 GGUF path.
- The NextN/MTP block is dropped, so no speculative decoding.
- Bank loading is serial. `_gguf_banks` raises `NotImplementedError` for the parallel reader since that path assumes safetensors and a GGUF is one packed file.
- `max_model_len` still reports the metadata context length even when KV is capped lower.
- I only tested the flashinfer attention backend. The triton fallback is untouched but unexercised here.

### DIFF
--- a/docs/models.md
+++ b/docs/models.md
@@ -1,8 +1,9 @@
 # Supported models
 
-FreeToken loads HF safetensors checkpoints directly (plus native GGUF for
-Gemma-4). The checkpoints below are known-good — the prebuilt kernels are tuned
-for them; other checkpoints of the same architectures work too.
+FreeToken loads HF safetensors checkpoints directly, plus native GGUF for the
+architectures listed under [GGUF](#gguf) below. The checkpoints below are
+known-good — the prebuilt kernels are tuned for them; other checkpoints of the
+same architectures work too.
 
 | Model | HF checkpoints |
 |---|---|
@@ -16,6 +17,35 @@ for them; other checkpoints of the same architectures work too.
 | Gemma-4 | [google/gemma-4-26B-A4B-it](https://huggingface.co/google/gemma-4-26B-A4B-it), [nvidia/Gemma-4-26B-A4B-NVFP4](https://huggingface.co/nvidia/Gemma-4-26B-A4B-NVFP4), [google/gemma-4-12B-it](https://huggingface.co/google/gemma-4-12B-it), [nvidia/Gemma-4-31B-IT-NVFP4](https://huggingface.co/nvidia/Gemma-4-31B-IT-NVFP4) .. |
 | MiniMax-M2.5 | [nvidia/MiniMax-M2.5-NVFP4](https://huggingface.co/nvidia/MiniMax-M2.5-NVFP4) |
 | Muse-Glimmer | [meta-models/Muse-Glimmer-30B](https://huggingface.co/meta-models/Muse-Glimmer-30B), [RedHatAI/Muse-Glimmer-30B-NVFP4](https://huggingface.co/RedHatAI/Muse-Glimmer-30B-NVFP4) |
+
+## GGUF
+
+Native GGUF, meaning the block-quantized weights are kept packed and dequantized inside
+the kernels rather than expanded to bf16 at load.
+
+| GGUF `general.architecture` | Covers |
+|---|---|
+| `gemma4` | Gemma-4 |
+| `qwen35moe` | Qwen3.5 / Qwen3.6 MoE (e.g. Qwen3.6-35B-A3B, Qwen3.5-122B-A10B) |
+| `qwen35` | Qwen3.5 / Qwen3.6 dense (e.g. Qwen3.6-27B, Qwen3.5-9B) |
+
+Quant types follow what the vendored kernels in `csrc/gguf/` implement:
+
+- Standard and K-quants (Q4_0, Q4_1, Q5_0, Q5_1, Q8_0, Q2_K through Q6_K) use MMQ for
+  prefill and MMVQ for decode.
+- I-quants (IQ1_S, IQ1_M, IQ2_XXS, IQ2_XS, IQ2_S, IQ3_XXS, IQ3_S, IQ4_NL, IQ4_XS) have no
+  MMQ kernel, so prefill dequantizes and runs a plain matmul; decode uses MMVQ.
+
+Two constraints worth knowing before picking a file:
+
+- A MoE checkpoint's routed-expert banks must use one ggml type across every layer. The GPU
+  slot pool is a single allocation with a single row stride, so a bank that changes type
+  between layers cannot be served and the load fails with the offending layers named.
+  llama.cpp's `_M` and `_XXS` levels raise the precision of the first few layers'
+  `ffn_down_exps` and hit this; `llama-quantize --pure` produces a checkpoint that loads.
+  Dense models have no expert banks and are unaffected.
+- GGUF paths are TP=1 only, and a NextN/MTP block in the checkpoint is dropped (served
+  text-only, no speculative decoding).
 
 ## MoE backends
 
--- a/docs/models.md
+++ b/docs/models.md
@@ -26,8 +26,14 @@ the kernels rather than expanded to bf16 at load.
 | GGUF `general.architecture` | Covers |
 |---|---|
 | `gemma4` | Gemma-4 |
+| `qwen3moe` | Qwen3 MoE (e.g. Qwen3-235B-A22B, Qwen3-30B-A3B) |
 | `qwen35moe` | Qwen3.5 / Qwen3.6 MoE (e.g. Qwen3.6-35B-A3B, Qwen3.5-122B-A10B) |
 | `qwen35` | Qwen3.5 / Qwen3.6 dense (e.g. Qwen3.6-27B, Qwen3.5-9B) |
+
+Split checkpoints load: point `--model` at any shard of a `-00001-of-000NN` set, or at the
+directory holding them. Metadata, config and tokenizer are read from shard 1 (later shards
+carry only `split.*` keys), and the tensor tables are aggregated across the set. A missing
+shard raises with the index named rather than loading a partial model.
 
 Quant types follow what the vendored kernels in `csrc/gguf/` implement:
 
--- a/python/freetoken/engine/engine.py
+++ b/python/freetoken/engine/engine.py
@@ -491,6 +491,7 @@ class Engine:
             kv_reserve_tokens=max(config.kv_reserve_tokens, min_reserve),
             page_size=page_tokens,
             quant_format=banks.quant_format,
+            gguf_expert_types=banks.gguf_expert_types,
         )
 
     def _init_offload_moe_cache(self, config: EngineConfig) -> OffloadMoeCache:
@@ -609,6 +610,7 @@ class Engine:
                 prefill_overlap=config.moe_prefill_overlap,
                 prefill_hit_d2d=config.moe_prefill_hit_d2d,
                 quant_format=banks.quant_format,
+                gguf_expert_types=banks.gguf_expert_types,
                 decode_target=decode_target,
                 hybrid_max_fetch=config.moe_hybrid_max_fetch,
             )
--- a/python/freetoken/engine/engine.py
+++ b/python/freetoken/engine/engine.py
@@ -1144,6 +1144,21 @@ def _cpu_moe_executor_viable(model_config) -> bool:
         return False
     expert_quant = getattr(model_config, "expert_quant", "none")
     fmt = expert_quant if expert_quant != "none" else (moe_wfmt or "bf16")
+    if fmt == "gguf":
+        # "gguf" is a container tag, not a layout: the checkpoint picks a ggml type per
+        # tensor and the concrete CPU format has to be recovered from the bank types.
+        # Testing the tag against _WFMT_IDS answers False for EVERY GGUF checkpoint, which
+        # silently disables the automatic residency split on hosts where CUDA pinning is
+        # quota-capped (WSL caps it near half of RAM). The symptom is not a clear refusal
+        # but cudaHostRegister failing partway through the banks.
+        from freetoken.moe.cpu_executor import _GGML_TO_CPU_FMT
+
+        types = getattr(model_config, "gguf_expert_types", None)
+        if not types:
+            return False
+        gate_up, down = int(types[0]), int(types[1])
+        # one weight_format serves both banks, so mixed types cannot run on the CPU path
+        return gate_up == down and gate_up in _GGML_TO_CPU_FMT
     return fmt == "mxfp4" or fmt in _WFMT_IDS
 
 
--- a/python/freetoken/engine/engine.py
+++ b/python/freetoken/engine/engine.py
@@ -491,7 +491,6 @@ class Engine:
             kv_reserve_tokens=max(config.kv_reserve_tokens, min_reserve),
             page_size=page_tokens,
             quant_format=banks.quant_format,
-            gguf_expert_types=banks.gguf_expert_types,
         )
 
     def _init_offload_moe_cache(self, config: EngineConfig) -> OffloadMoeCache:
--- a/python/freetoken/kernel/csrc/cpu_moe/cpu_moe_ext.cpp
+++ b/python/freetoken/kernel/csrc/cpu_moe/cpu_moe_ext.cpp
@@ -1650,6 +1650,13 @@ struct CpuMoeExecutor {
       const uint8_t* w = gu_packed_l + ((size_t)e * (2 * I) + row) * (size_t)q4_gu_row_bytes;
       return q6_k_dot_f32_scalar(w, x, H);  // W4A16: bf16 activations, K-quant dequant
     }
+    // Anything that reaches here is assumed NVFP4 and dereferences the scale/global
+    // pointers, which are null for formats that do not have them (the GGUF banks pass 0).
+    // Falling through with an unhandled format therefore segfaults inside the worker
+    // thread rather than reporting anything useful, so reject it here instead.
+    TORCH_CHECK(fmt == WF_NVFP4 || fmt == WF_DSFP4,
+                "cpu_moe gemm1_dot: unhandled weight_format ", fmt,
+                " (handled: bf16=0, nvfp4=1, mxfp4=2, dsfp4=3, q4_0=4, q4_k=5, q6_k=6)");
     const size_t r = (size_t)e * (2 * I) + row;
     if (use_vnni)
       return nvi8dot(gu_packed_l + r * (size_t)(H / 2), gu_scale_l + r * (size_t)(H / 16),
--- a/python/freetoken/kernel/csrc/cpu_moe/cpu_moe_ext.cpp
+++ b/python/freetoken/kernel/csrc/cpu_moe/cpu_moe_ext.cpp
@@ -1212,7 +1212,151 @@ q4dot_fn select_q4dot() {
   return q4_0_dot_i8_scalar;
 }
 
-enum WFmt { WF_BF16 = 0, WF_NVFP4 = 1, WF_MXFP4 = 2, WF_DSFP4 = 3, WF_Q4_0 = 4 };
+// ----------------------------- Q4_K and Q6_K (W4A16) ----------------------------
+// Correctness-first scalar K-quant expert dot kernels: dequantize blocks and accumulate
+// against bf16 activation rows. AVX2/VNNI optimizations are a deliberate follow-up.
+// Reference: llama.cpp ggml-quants.c, ggml-cuda/convert.cu, models/gguf/dequant.py.
+
+// Helper: extract 6-bit scale and min from Q4_K's packed scales array.
+// Q4_K packs scales and mins using 6 bits each, packed into 12 bytes for 256 elements.
+inline void get_scale_min_k4(int j, const uint8_t* q, uint8_t& scale, uint8_t& minv) {
+  if (j < 4) {
+    scale = q[j] & 63;
+    minv = q[j + 4] & 63;
+  } else {
+    scale = (q[j + 4] & 0x0F) | ((q[j - 4] >> 6) << 4);
+    minv = (q[j + 4] >> 4) | ((q[j - 0] >> 6) << 4);
+  }
+}
+
+float q4_k_dot_f32_scalar(const uint8_t* w, const bf16_t* x, int K) {
+  // Q4_K, 256-element super-blocks of 144 bytes: half2 dm | scales[12] | qs[128].
+  // dm.x scales the quants, dm.y scales the per-sub-block minimum that is SUBTRACTED.
+  //
+  // Two things here are easy to get wrong and both produce fluent-looking garbage rather
+  // than an obvious failure, so this mirrors dequantize_block_q4_K in gguf/dequantize.cuh
+  // element for element:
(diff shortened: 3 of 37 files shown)


vcruz305:
Pushed one more commit: Q4_K and Q6_K expert kernels for the cpu and hybrid MoE backends.

Q4_0 was the only GGUF type with a CPU dot kernel, so `--moe-backend cpu` and `hybrid` refused every K-quant checkpoint and sent those users to `offload`. That is the wrong way round. The people who need the hybrid backend are the ones short of VRAM, and they are running exactly these quants.

While testing it I found the format never actually reached the executor. The GGUF offload path tags the cache `quant_format="gguf"` and carries the ggml types separately in `gguf_expert_types`, but `cpu_executor` was matching on literal strings like `"q4_0"` that nothing produces. So the CPU backend would have refused a real GGUF checkpoint even for Q4_0, which is the one type it already supported. `_resolve_gguf_format` now maps the bank types onto a single weight format, and refuses with a reason that tells you what to do:

- mixed banks (Q4_K_M puts `ffn_down_exps` in Q6_K) point at `--moe-backend offload` or a `--pure` requantization
- a uniform type with no CPU kernel points at `offload` and says so

Those are separate messages because the fix is different.

Tests are in `tests/moe/test_cpu_moe_kquant.py`, checking both kernels against the vendored CUDA dequant plus the production bf16 GPU decode on byte-identical banks, at three batch sizes. Blocks are synthesized rather than carved out of a real file so it runs in CI without a multi-GB download.

Two things the tests caught that I want to flag, because neither is visible structurally and both give you a model that loads and runs at full speed:

- Q4_K quants are unsigned 0..15 with a per-sub-block minimum. They are not Q4_0's `q - 8`. Getting that wrong reads as a perfectly plausible kernel and drops cosine similarity to 0.008.
- Within a Q4_K 64-element group, a byte's high nibble is element `l + 32`, not `l + 1`, and it carries a different scale/min pair.
- Q6_K's `qh` shift is per group. An extra term in the lane index corrupted the upper half of every group while still correlating 0.17, which looks like a tolerance problem rather than a wrong kernel.

This is also why #138 matters. Without a `default:` that raises, an unsupported type on the MMQ path produces exactly this failure mode with no error at all.

zyy0212time-del:
## Bug report + proposed fix: untied lm_head misses the prefill last-position slice -> first token sampled from position 0

**Environment**: Windows 11, RTX 5060 Laptop 8GB, this branch @ 6f6c864, `--moe-backend cpu`
**Model**: `Qwen3.6-35B-A3B-Uncensored-HauhauCS-Aggressive-Q4_K.gguf` (Q4_K, untied lm_head, `tie_word_embeddings=false`)

### Symptom

`/v1/chat/completions` produces only 2-3 tokens (e.g. `user`) then stops.
`/v1/completions` works for plain prompts but degrades for prompts containing `<|im_start|`.

### Root cause

`engine.forward_batch` slices `logits[: batch.size]` -- during prefill the LM head must return only the last-position logits per request. `ParallelLMHead` and gemma4's `GGUFTiedLMHead` do this via `batch.attn_metadata.get_last_indices(bs)`, but the qwen35 untied path installs a plain `GGUFLinear` (no slice), so the engine slice picks up **position-0 logits** and the first token is sampled from the wrong row.

Evidence (temperature=0, cross-checked against llama.cpp b10451 on the identical file and tokenization):

| prompt | expected top1 (llama.cpp) | FreeToken first token |
|---|---|---|
| `Hello! The capital city of France is` | ` Paris` (id 11751) | `,` (position-0 prediction after "Hello") |
| `<|im_start| user ... <|im_end| <|im_start| assistant` (chat template) | `<think>` (id 248068) | `user` (position-0 prediction after `<|im_start|`) |

A logits dump confirms the prefill logits themselves are correct (top-k matches llama.cpp logprobs); only the row selection is wrong. Tied-embedding models are unaffected (`GGUFTiedLMHead` already slices), so this hits every GGUF checkpoint with an untied lm_head.

### Proposed fix (`models/qwen3_5_moe/gguf.py`, end of `convert_qwen35_to_gguf`)

Replace the untied branch:

``python
else:
    swap_linear(model, "lm_head", qt(-1, "output.weight"))
``

with a slicing subclass, mirroring `GGUFTiedLMHead`:

``python
else:
    class GGUFUntiedLMHead(GGUFLinear):
        def forward(self, x: torch.Tensor) -> torch.Tensor:
            from freetoken.core import get_global_ctx

            batch = get_global_ctx().batch
            if batch.is_prefill:
                indices = batch.attn_metadata.get_last_indices(batch.size)
                x = x[indices].contiguous()
            return super().forward(x)

    lin = model.lm_head
    out_features, in_features = lin.weight.shape
    model.lm_head = GGUFUntiedLMHead(
        in_features, out_features, qt(-1, "output.weight"), has_bias=lin.bias is not None
    )
``

### Verified

With the patch applied, temperature=0:
- `Hello! The capital city of France is` -> ` Paris.` (matches llama.cpp)
- chat-template prompt -> `<think>` then normal generation
- chat completions: full answer `The capital city of France is Paris.`, `finish_reason=stop`

Happy to send this as a PR if helpful.

vcruz305:
Thank you for this. Reproduced, and it is fixed in b2f8475.

Your diagnosis is right and your read of the consequence is better than mine was. I had found the same missing slice about an hour before you posted, but I came at it from the memory side: on an 8GB card a ~1800-token prompt was killing the worker, because logits are `[tokens, vocab]` and Ornith's vocab is 248,320, so that is 486 KiB of logits per token and a single 894 MB allocation. I fixed it as a memory bug and moved on.

I missed that `forward_batch` does `logits[: batch.size]`, so on smaller prompts that never OOM you do not crash, you just sample the first token off position 0 and get a quietly wrong answer. That is the worse half of the bug and you caught it. The `Hello! The capital city of France is` case is a good repro precisely because it fits comfortably in VRAM, so nothing looks wrong.

I went with the same slice you proposed, in a slightly different place. Rather than a nested class in `convert_qwen35_to_gguf`, it is `GGUFLMHead` in `layers/gguf.py`:

```python
class GGUFLMHead(GGUFLinear):
    def forward(self, x: torch.Tensor) -> torch.Tensor:
        from freetoken.core import get_global_ctx

        batch = get_global_ctx().batch
        if batch.is_prefill:
            indices = batch.attn_metadata.get_last_indices(batch.size)
            x = x[indices].contiguous()
        return super().forward(x)
```

Two reasons for the move. `qwen3_moe/gguf.py` has the identical untied branch and the identical bug, so it needed fixing in both places. And the slice deliberately does not go into `GGUFLinear` itself, since that class is also every q/k/v/o and MLP projection in the model and must not slice.

I checked the other adapters: gemma4 always takes the tied path, so `GGUFTiedLMHead` covers it, and those two were the only untied sites.

Verified on Ornith-1.5-35B-A3B IQ3_S, untied lm_head, temperature 0:

```
Hello! The capital city of France is  ->  " Paris. Paris is located in the northern part of France"
The capital of Japan is               ->  " Tokyo, a city that has been the seat of the"
```

and on the memory side, same box, 8GB, offload backend: ~1,800 tokens went from killing the worker to 6.1s, and 4,598 tokens now prefills in 16.9s.

One thing you may hit next on an 8GB card. Past roughly 5-6k prompt tokens it still runs out of room, but somewhere else: the GDN chunked kernel's activations, `chunk_gated_delta_rule_fwd` -> `chunk_fwd_o`. That one is headroom rather than a defect, and `--max-prefill-length 2048` handles it (6,160 tokens in 23.2s, 7,784 in 27.5s on the same box). Worth knowing that the 8192 default does not chunk a typical long prompt at all, so you get one enormous forward pass.

No need to send a PR for this one, but the report was genuinely useful and I would take more of them.

max-nuv:
I really hope you can get this fix working on Windows, because after pulling your branch and fixing the build for MSVS, I got a CUDA error: invalid argument in ggml_moe_a8_vec with the first 20k tokens prompt. After running short tests on Qwen 3.6 35B A3B (NVFP4) vs. Ornith 1.5 35B A3B (IQ3_XXS), I got the following results:
| Test | Qwen | Ornith |
|--------|--------|--------|
| Logic TTFT| 1.09 sec| 0.78 sec |
| Logic speed | 26.7 tokens/sec | 40.9 tokens/sec |
| Logic quality | Hit the limit. | Correct: $100, finish=stop |
| Code TTFT | 0.89 sec | 0.69 sec |
| Code speed | 27.4 tokens/sec | 40.0 tokens/sec | 
| Code quality | Hit the limit. | Correct code, but finish=length | 

So, Ornith, with a smaller quantum, produced better and faster results, and I'm really looking forward to trying the model in real-world tasks.

P.S. 3070 8GB VRAM, context 96k tokens

vcruz305:
Found it and fixed it in 9952a39. Thank you for the report, and for building it on MSVC to get there.

It is not a Windows problem and not a 3070 problem, which is worth saying because it would have sent me looking in the wrong place. `moe_vec.cuh` carries the routed-pair index (`token * top_k + slot`) in `gridDim.z`, and CUDA caps `gridDim.z` at 65535 on every architecture. All 19 launchers computed:

```cpp
const dim3 block_nums(block_num_y, 1, tokens * top_k);
```

At top_k 8 that ceiling is **8191 tokens**. Your 20k prompt asks for z = 160000, the launch is rejected with `cudaErrorInvalidConfiguration`, and torch surfaces it as the fairly unhelpful "invalid argument". Anyone with a long enough prompt hits it on any GPU and any OS.

`fused_experts_gguf` makes two launches, gate_up as `(top_k, N)` and down as `(1, N * top_k)`, and both land on `z = N * top_k`, so they overflow together.

The launcher now issues token-aligned chunks. Moving the pair index into `x` would also have fixed it, since `gridDim.x` allows 2^31-1, but it would cost the locality the kernel is built around: with the row index in `x`, consecutive blocks walk rows of the *same* expert and reuse its weight pages. Chunking preserves that and leaves the kernel body untouched. Chunks are whole tokens so the kernel's own `token = blockIdx.z / topk` arithmetic stays valid against the offset `vy` / `dst` / `topk_ids` pointers. The 19 duplicated launcher bodies now share one `moe_vec_launch<>` helper.

Tests in `tests/kernels/test_moe_vec_grid_z.py` check two things, and the second is the one that matters:

- a launch past the ceiling succeeds (verified to z = 160000, i.e. 20k tokens at top_k 8)
- rows below the ceiling are **bit-identical** however many chunks were used

A wrong per-chunk pointer offset would still launch and still look healthy, and would quietly corrupt everything above 8191 tokens. The equality check is what catches that; a crash test would not.

Also verified the kernel still compiles for sm_75 and sm_89.

On your numbers: Ornith IQ3_XXS beating Qwen NVFP4 on both speed and quality is a real effect rather than a fluke of the quant. Ornith's expert banks come out uniform at IQ3_XXS, so the whole MoE stays on the packed `moe_vec` path with no dequantization, whereas the NVFP4 route materializes more per token. The `finish=length` on both Qwen tests also suggests it was running into the output cap rather than answering worse, so the quality column may be flattering Ornith slightly.

One thing you should hit next, at 96k context. There are two other prefill limits I have measured, neither of which is this bug:

- An untied lm_head was projecting **every** prefill position instead of the last, so logits were `[tokens, vocab]`. Fixed in b2f8475. On a large-vocab model that is hundreds of MB per prompt and it also made the first sampled token come from position 0.
- Past roughly 5-6k prompt tokens on an 8GB card the GDN chunked kernel's activations run out of room (`chunk_gated_delta_rule_fwd` -> `chunk_fwd_o`). That one is genuine headroom rather than a defect, and `--max-prefill-length 2048` handles it. Worth knowing the 8192 default does not chunk a typical long prompt at all, so you get one enormous forward pass.

If you pull again you will get all three. I would be glad of more reports like this one.

zyy0212time-del:
**Windows/CUDA follow-up: end-to-end verified + a request on merge timing / Windows wheels**

Thanks again for b2f8475 and 9952a39. Reporting back from the Windows side for completeness.

Our engine is this branch @ g30aa89115 (it predates both fix commits) with equivalent local patches for the two bugs, so this validates the diagnosis and approach rather than the exact commits:

- Qwen3.6-35B-A3B Q4_K (pure-quant, abliterated) on an 8GB RTX 5060 Laptop, `--moe-backend cpu`: ~70s cold start, ~13.4 t/s decode, chat output correct.
- Prompts around 8K tokens (agent frameworks with large system prompts) no longer crash the worker - before the chunking fix they reliably took down ggml_moe_a8_vec.

One practical concern: FreeToken Desktop on Windows keeps showing the "engine update available" prompt. The latest release (v0.1.2, published 2026-08-19, before this PR was opened) ships Linux-only wheels and has no qwen35moe GGUF support, so Windows users depending on this branch currently have no safe update path.

Is there a timeline for merging this? And could the next release include win_amd64 wheels? Happy to help test Windows builds (CUDA 13.0 / MSVC 17.14) - this machine is set up and idle.

tuxevil:
Some end-to-end Qwen3.6-35B-A3B results from #131 that may be useful, especially for the GGUF and hybrid paths.

Test system:

* Quadro RTX 4000, Turing sm_75, 8 GB
* Ryzen 9 7945HX (16C/32T)
* 96 GB DDR5-5600
* PCIe Gen3 x16, measured ~13.17 GB/s H2D/D2H
* PyTorch 2.11 / CUDA 13.3
* #131 plus the optional-backend sm_75 gating from #24

I tested several Qwen3.6 quants with a fixed benchmark covering decode, prefill, prefix caching and hybrid/offload.

### GGUF compatibility

These uniform-bank files loaded and served successfully:

* IQ3_S
* Q4_K
* Q5_K_S

I also ran into several GGUFs with mixed expert-bank types. The rejection behavior added by this PR was useful: rather than silently misinterpreting the slot layout, FreeToken identified the differing bank types and refused them. A uniform Q5_K_S worked correctly.

Q5_K_S currently remains offload-only because there is no compatible Q5_K CPU expert kernel, as expected.

### Decode vs prefill

Best Eco-mode offload results:

| quant  | best ratio | fixed 511-token decode |
| ------ | ---------: | ---------------------: |
| IQ3_S  |        .90 |            45.44 tok/s |
| Q4_K   |        .93 |            39.45 tok/s |
| Q5_K_S |        .93 |            34.53 tok/s |

However, ~2K-token prefill was:

| quant  |      prefill |
| ------ | -----------: |
| Q4_K   | ~193.6 tok/s |
| Q5_K_S | ~171.7 tok/s |
| IQ3_S  | ~143.3 tok/s |

That seems like useful real-world confirmation of the implementation detail described in this PR: IQ3_S decode is excellent, but its I-quant prefill path (dequant + matmul because there is no MMQ kernel) gives up a substantial amount of prefill throughput to K-quants using MMQ.

On this system Q4_K was ~35% faster than IQ3_S for fresh ~2K context despite IQ3_S being ~15% faster at sustained decode.

### Hybrid vs offload

I also tested Q4_K hybrid with caps 1/2/4/8, first with the 7945HX deliberately limited to ~2.5 GHz and then again with CPU boost enabled and governor=`performance`.

Best Q4_K results:

Eco:

* offload .93: 39.45 tok/s
* hybrid cap8: 23.92 tok/s

Boost/performance:

* offload .93: 39.16 tok/s
* hybrid cap8: 26.60 tok/s

So CPU boost improved the best hybrid result by ~11%, confirming that the CPU path is doing useful work, but hybrid was still ~32% slower than offload.

GPU telemetry made the bottleneck quite visible:

* offload: ~99% GPU utilization
* hybrid cap8: ~62% GPU utilization

The GPU was spending a significant amount of time waiting for the CPU side of the hybrid pipeline.

I saw the same general behavior with native NVFP4:

Boost:

* offload .93: 37.16 tok/s
* best hybrid (cap1): 25.39 tok/s

### Synthetic bandwidth vs end-to-end inference

One potentially actionable observation is that the synthetic bandwidth benchmark can make hybrid look considerably more attractive than actual inference.

For example, in Boost mode `ft bench bw` measured NVFP4 roughly as:

* CPU MoE: 41.17 GB/s
* PCIe gather: 13.27 GB/s
* ratio: 3.10
* recommendation: hybrid

but the actual Qwen3.6 inference result was:

* offload: 37.16 tok/s
* hybrid: 25.39 tok/s

So at least on this CPU + Turing combination, CPU-MoE/PCIe bandwidth ratio alone does not capture the synchronization/scheduling cost and GPU starvation of the full hybrid pipeline.

This may be worth considering if these microbench results are ever used for automatic backend selection: treating them as a viability signal rather than a definitive `hybrid > offload` prediction seems safer.

### CPU Eco vs Boost

As another data point, the best offload result was essentially insensitive to CPU boost across every model I tested:

| model  |   Eco | Boost |
| ------ | ----: | ----: |
| IQ3_S  | 45.44 | 44.96 |
| Q4_K   | 39.45 | 39.16 |
| NVFP4  | 37.21 | 37.16 |
| Q5_K_S | 34.53 | 34.22 |

That seems consistent with the winning offload configurations being overwhelmingly GPU-bound on this machine.

Happy to provide the benchmark scripts/raw summaries if any of these results would be useful for regression tests or backend-selection work.

PR #23, opened 2026-08-22
feat: add simple ROCm GPU support for RDNA3 (gfx1100-1103)

bouclem:
## ROCm support for RDNA3 GPUs

Adds initial ROCm support so FreeToken builds and runs on AMD RDNA3 cards (gfx1100-1103).

Supported GPUs:
- gfx1100 - RX 7900 XTX / XT
- gfx1101 - RX 7900 GRE
- gfx1102 - RX 7700 / XT
- gfx1103 - RX 7600 / XT

What works:
- `pinned_tensor.cpp` compiles under hipcc via a hip_compat.h shim that maps CUDA runtime API to HIP
- setup.py detects ROCm at build time, links amdhip64, passes --offload-arch for the four gfx targets
- JIT compilation routes to HIP flags on ROCm
- NVIDIA-specific arch checks (sm90/sm100) return None on ROCm so the engine falls back to Triton kernels
- nvcc version check is skipped on ROCm
- clangd generation falls back to rocm-smi

Left as TODO(ROCm) in the code:
- NCCL to RCCL migration for multi-GPU
- flashinfer/sgl_kernel have no ROCm builds (Triton fallback works)
- Triton autotune configs need RDNA3 tuning
- PDL has no ROCm equivalent
- rocm-smi auto gfx detection (currently relies on env vars)

Issue #82, opened 2026-08-23
ROCm RDNA4 gfx1201 Ported. Windows only for now. Test please


Maxritz:
https://github.com/[Maxritz/FreeToken-rocm-test](https://github.com/Maxritz/FreeToken-rocm-test/tree/main)/tree/main

## Windows ROCm Port

This fork brings FreeToken up on **Windows 11 + AMD ROCm** with no NVIDIA toolchain.
Bring-up target: RX 9070 XT (`gfx1201`). Everything below was verified live: model load,
prefill, decode (~57 tok/s bf16 3B), SSE streaming, and the bundled mini web UI.

### What works today

| Area | Status |
|---|---|
| Dense HF safetensors models (bf16/fp16), single GPU | ✅ working (Qwen2.5-3B verified; 7B fits only with `--num-pages` cap) |
| OpenAI-compatible `/v1/chat/completions`, streaming | ✅ working |
| Triton attention/norm/activation kernels on RDNA4 | ✅ working (AMD backend) |
| tvm-ffi JIT CUDA kernels (`index`, `store`, `radix`) | ✅ compiling via `hipcc` + host `clang++` |
| Web chat UI | ✅ `G:\FreeToken\webui\index.html` (serve on port 1420) |
| CUDA extensions skipped gracefully | ✅ `FREETOKEN_SKIP_CUDA_EXT=1` |
| Pinned-memory / ZMQ IPC fallbacks on Windows | ✅ TCP loopback + torch pin_memory |
| MoE expert offload (split CPU–GPU execution) | ⚠️ untested on Windows yet |
| GGUF loader | ⚠️ `gemma4` architecture only — Qwen/Laguna/DeepSeek adapters pending |
| ROCmFPX-quantized GGUFs | ❌ proprietary quant types, needs the source fork |

If you use it, I hope you will let me know how it goes. I just spent the whole night tinkering to get it to work. Thanks to #OPENCODE #OX ALPHA LLM , and #AMD (I used to love you guys, but you made me search for the ROCM 10.1 SDK really difficult . But anyway. ACCEPT ME in the AI test group or send me an invite.)
Anyway, if I don't doze off and wake up late to travel to my work location, I will try to port to RDNA2 as well. It is a 6700xt . Now will tinker with some more models and GGUF loading until I fall asleep in my chair.

hojnikb:
hopefully some madlad ports this to vulkan as well for those devices that dont support rocm (or is it broken).

[...]

PR #132, opened 2026-08-24
feat(rocm): add RDNA3 and RDNA4 runtime foundation

zihaomu:
## Summary

- add an independent ROCm source-build foundation for RDNA3 (`gfx1100`-`gfx1103`) and RDNA4 (`gfx1200`, `gfx1201`);
- support both traditional `/opt/rocm` layouts and the modular ROCm SDK shipped by the official ROCm 7.14 PyTorch image;
- build and load the pinned-memory and CPU-MoE extensions on HIP;
- make the pinned-host fast-index-copy and Triton activation/norm paths work on ROCm;
- detect the visible `gcnArchName`, with explicit environment overrides for cross compilation.

## Why this targets `main` directly

This PR is a standalone, current-`main` integration. It incorporates the useful work from #23 and the RDNA3 runtime follow-up at bouclem/FreeToken#1, then adds the RDNA4 and ROCm 7.14 changes. It therefore does not require either external feature branch to merge first.

This intentionally overlaps those open PRs so maintainers can review and merge a complete, hardware-tested ROCm foundation without being blocked by a cross-fork base chain.

## Attribution

- @bouclem authored the initial RDNA3/ROCm foundation in #23; that author information is preserved in commit `7caa62d` (upstream source commit `27c0977`).
- @nekomario28 authored the RDNA3 runtime completion in bouclem/FreeToken#1; author information is preserved across commits `af67560` through `62bb962` (upstream follow-up head `4e11ed0`).
- @zihaomu added RDNA4 discovery, ROCm 7.14 modular-SDK support, versioned HIP-runtime linking, runtime DLPack handling, tests, and installation documentation.

Thank you to both contributors for establishing and validating the earlier ROCm paths.

## Review hardening

- removed unused public gfx-family helpers rather than exposing ambiguous APIs without consumers;
- changed pinned-host allocation aliases to `hipHostMalloc` / `hipHostFree` and unified the ROCm macro guard;
- documented that cached architecture detection is bound to the first visible/current device;
- added an actual Triton compile-time/host E4M3 twin test on ROCm;
- added launch-argument tests proving ROCm activation and norm paths omit CUDA-only `launch_pdl`.

The reported E4M3 mismatch is not reproducible with AMD Triton 3.7.1: `target_info.cuda_capability_geq()` returns false for a HIP target, so both twins select emulation. The added compiled-kernel test protects that contract.

## Validation

Hardware and software:

- AMD Radeon AI PRO R9700 (`gfx1201`)
- PyTorch `2.11.0+rocm7.14.0`
- HIP `7.14.60850`
- AMD Triton `3.7.1`

Results:

- clean `python -m pip install -e . --no-build-isolation --no-deps`: PASS; both native extensions compiled and loaded;
- cold TVM-FFI fast-index-copy JIT build: PASS;
- architecture, pinned/copy, Triton attention, Gemma norm, SwiGLU, and CPU-MoE graph replay focused set: `48 passed`;
- review-hardening E4M3/launch-contract tests: `7 passed`;
- `python -m pip check`: PASS;
- `git diff --check`: PASS.

## Compatibility and merge gate

ROCm-specific code is guarded by HIP/ROCm detection and existing CUDA paths are retained. Physical NVIDIA regression testing is not available on the development host and is not claimed. Upstream CUDA CI (clean install plus E4M3, activation/norm PDL, pinned extension, and JIT smoke tests) remains required before merge.

## Follow-ups

Kept out of this foundation PR for separate review:

1. generic TVM-FFI index/store HIP portability (#133);
2. CUDA-only optional-backend gating (#134);
3. RCCL tensor-parallel routing (#135);
4. native GGUF kernels on ROCm, validated on RDNA4 (#136);
5. RDNA-specific performance tuning and full model-serving validation.


zihaomu:
Draft follow-ups are now available:

- #133 — TVM-FFI index/store JIT portability for HIP
- #134 — ROCm gating for CUDA-only optional backends
- #135 — RCCL routing for tensor-parallel communication
- #136 — native GGUF kernels on RDNA4 (also follows #133)

All four are opened as Draft PRs against `main`. Until this foundation PR merges, their GitHub “Files changed” views temporarily include the #132 foundation commits; each PR body lists its intended incremental review scope. They will remain drafts and be rebased/refreshed as their dependencies land.

Ewwgoblins:
## Independent test report: RDNA3 / gfx1100 (RX 7900 XT)

Ran this branch on an RDNA3 card, since the validation above is gfx1201. Short version: the foundation works on gfx1100, I was able to serve models end to end, and the offload architecture measures very well against llama.cpp on this hardware.

### Environment

- AMD Radeon RX 7900 XT (`gfx1100`), Ryzen 7 7800X3D, 62 GiB DDR5
- Container: `rocm/pytorch:rocm7.14_ubuntu24.04_py3.12_pytorch_release_2.11.0`
- `torch 2.11.0+rocm7.14.0`, Triton `3.7.1+git0263a6a6.rocm7.14.0`
- Branch `feat/rocm-rdna3-rdna4-foundation` @ `4541756`, **#132 only** (no #133-#136)
- `PYTORCH_ROCM_ARCH=gfx1100`, `FREETOKEN_ROCM_ARCH=gfx1100`

### Install

`python -m pip install --no-build-isolation -e .` — PASS. Both native extensions compiled and loaded. No `HSA_OVERRIDE_GFX_VERSION` needed. The widened `triton>=3.6,<3.8` pin on this branch correctly preserves the image's AMD Triton 3.7.1.

### Test suite

`pytest tests -q -m "not slow"` → **1331 passed, 20 failed, 24 skipped, 11 deselected** (211.60 s).

- **16 × `tests/kernels/test_dsv4_sparse_attn.py`** (the whole file) — `RuntimeError: PassManager::run failed` at `triton/backends/amd/compiler.py:273`. AMD Triton cannot compile that kernel on gfx1100. This file is not in the validation list above, so I believe this is new. Happy to open a separate issue if you would prefer it tracked.
- **3 × flashinfer absent** — `test_cache_budget.py` (×2), `test_offload.py::test_adjust_config_converts_moe_cache_rate_to_cache_size`. Expected, installed without `[accel]`; presumably #134.
- **1 × `test_gemma4_fused_ops.py::test_gemma4_router_uses_sgl_kernel_topk_softmax_semantics`** — `Tensor-likes are not equal!`. May also be sgl_kernel absence; not chased.

### `ft bench bw` on gfx1100

```
ceilings: CPU STREAM read 47.1 | PCIe linear H2D 28.42  D2H 28.56 GB/s  (threshold 2.0x)
  bf16   47.4/28.6 = 1.66x -> offload     nvfp4  44.2/28.6 = 1.55x -> offload
  mxfp4  35.9/28.6 = 1.26x -> offload     ds_fp4 45.6/28.6 = 1.59x -> offload
  fp8    n/a (CPU MoE has no fp8_block weight path)
```

### Serving — works, and the offload numbers are the interesting part

**openai/gpt-oss-20b (12.8 GiB safetensors, MXFP4).** This model FITS in 20 GB, which lets me compare all four configurations on one card. Method: HTTP API, 512 tokens, `--ignore-eos`, warmup discarded, page cache pre-warmed, n=3, sub-1% variance throughout.

| engine / backend | decode tok/s |
|---|---|
| llama.cpp (Vulkan) `-ngl 99 -fa on` | **193.5** |
| llama.cpp (Vulkan) `--cpu-moe` | **25.5** |
| FreeToken `--moe-backend fused` | **100.8** |
| FreeToken `--moe-backend offload` | **91.2** |
| FreeToken `--moe-backend hybrid` (forced) | 71.8 |

Two things stand out:

1. **The offload penalty is the story.** llama.cpp loses 86.8% moving experts off the card (193.5 → 25.5). FreeToken loses 9.5% (100.8 → 91.2). In the offloaded regime FreeToken is **3.58× faster** than llama.cpp on the same model and card.
2. **`ft bench bw`'s hybrid heuristic is correct here, and I verified it by overriding it.** It recommended `offload` (1.65×, under the 2.0 threshold); forcing `hybrid` costs 21%. Also worth recording: hybrid *runs* on gfx1100, so that path is functional on ROCm.

**Qwen/Qwen3.6-35B-A3B-FP8 (35 GB, does NOT fit 20 GB)** — `moe_backend='offload'`, `cache_type='hybrid_radix'`, VRAM 19.90 → 2.02 GiB, ~3 min load: **37.8 tok/s**. For scale, my own earlier llama.cpp measurements of Qwen3.6-35B-A3B on this card were 115.9 tok/s for Q4_K_M resident and 28.25 tok/s for Q4_K_M with `--n-cpu-moe 32`. So FreeToken serves a 75% larger, higher-precision checkpoint 34% faster than llama.cpp manages on the smaller Q4 once it has to offload. Different quantisations, so treat that one as indicative; the gpt-oss-20b table above is exact.

Since "full model-serving validation" is an open follow-up here and #136 says serving is out of scope, I hope this is useful.

### One design observation

`auto` selected `offload` for gpt-oss-20b even though it fits the card, costing 9.5% for nothing. Per `docs/models.md` that is intended (dense → fused, MoE → offload always), and `fused` is documented as "never auto-selected" — but on a card where the MoE fits, `auto` picks the slower path. A fit check before defaulting to offload might be worth considering.

### Host-RAM ceiling, for anyone sizing a box

Offload pins expert memory, so host RAM is the real limit. On 62 GiB total / 45 GiB available, a 35 GB model consumed ~31 GB pinned and dropped available RAM to 14 GB — at which point sshd could no longer fork, while the box stayed up and port 22 stayed open. Extrapolating, ~50 GB models need ~46 GB pinned and are out of reach on this machine. Not a bug, just a sizing note.

### Not tested (deliberately)

The **GGUF** path. On #132 alone it fails in `kernel/gguf.py`: `--expt-relaxed-constexpr` and `-ccbin` are passed unconditionally in `extra_cuda_cflags` and reach `hipcc`/`clang++`, which reject them. After locally gating those on `torch.version.hip` the next failure is `thrust/complex.h` not found — the modular ROCm SDK wheels in that image ship no rocThrust. Both look like exactly what #136 covers, and #136 depends on #133 which I did not apply, so I did not pursue it and am not filing anything for it.

---

Happy to re-run any of this, or to test the follow-up PRs on gfx1100 if a second architecture is useful to you. Thanks for doing this work — the RDNA3 scope in #132 is what made it worth trying. (and I Hope i did this correctly! its my first comment!)

zihaomu:
Hi @Ewwgoblins Thank you for your detailed report, it was very helpful.

thedandano:
## Independent test report: RDNA3 / gfx1101 (RX 7800 XT) on ROCm 7.14 and 10.0

Second RDNA3 data point, following @Ewwgoblins' gfx1100 report above, and the first data I'm aware of on the new ROCm 10.0 images, which hit a Triton regression that 7.14 doesn't. Tested at 4541756; I re-checked the diff to c0713e4 and everything below still applies to the current head.

### System

| | |
|---|---|
| GPU | AMD Radeon RX 7800 XT, 16 GB (Navi 32, `gfx1101`) |
| CPU / RAM | Ryzen 7 7800X3D / 30 GB |
| Containers | `rocm/pytorch:rocm7.14_ubuntu24.04_py3.13_pytorch_release_2.11.0` and `rocm10.0_..._2.11.0` |
| Stack | torch 2.11.0+rocm{7.14,10.0}, Triton 3.7.1 (7.14) / 3.8.0 (10.0), Python 3.13 |

### What works

- Build and JIT: both extensions compile, and a cold JIT build emits real `gfx1101` code objects (checked in the built `.so` bundles).
- ROCm 7.14: `tests/engine/ tests/moe/` runs 186 passed, 9 skipped, 3 failed. The 3 failures (`test_cache_budget.py`, `test_offload.py`) are config-resolution tests that depend on NVIDIA-only optional deps being absent, not RDNA3 problems.
- End-to-end serving on ROCm 10.0, with the fix from finding 1 applied: gpt-oss-20b (MXFP4) serves and answers coherently. Load-to-ready is 69 s from safetensors. Converting to FTW with `ft checkpoint` took 65 s and kept `mxfp4_triton`; loading from FTW takes 41 s. Both formats produce the same completions.

### Finding 1: on ROCm 10.0 / Triton 3.8.0, the `moe_align_block_size` Triton kernels miscompile on gfx1101 and crash the GPU

On the 10.0 image, `pytest tests/engine/ tests/moe/` aborts (exit 134):

```
Memory access fault by GPU node-1 ... Reason: Page not present or supervisor privilege.
```

Root cause: Triton 3.8.0's AMD backend miscompiles the two alignment kernels in `python/freetoken/kernel/triton/moe_align.py` on gfx1101. They emit garbage `expert_ids`, and the even-K branch of `fused_moe_kernel` (`python/freetoken/kernel/triton/fused_moe.py:276`, unmasked `tl.load(b_ptrs)`) turns a garbage id into a wild pointer. Triton 3.7.1 on the 7.14 image compiles the same kernels correctly, so this only bites the 10.0 line. Zeroing the buffers hides the crash but leaves wrong routing (roughly 70% mismatched outputs), so it's a genuine miscompile rather than uninitialized memory.

What worked here: replace the alignment step on ROCm with a static-shape pure-PyTorch implementation (`scatter_add_` histogram, `searchsorted`, stable `argsort`), keeping the Triton path for CUDA. It's safe under HIP graph capture (verified with capture/replay against mutated inputs). With it, the same suite goes from abort to 186 passed on ROCm 10.0, and serving works as described above. It costs about 13% on the MoE layer versus the Triton kernels, so it's a fallback, not a replacement. Happy to open a PR against `feat/rocm-rdna3-rdna4-foundation` if useful. It may also be worth reporting the miscompile to Triton upstream.

### Finding 2: dsv4 sparse attention is broken on 7.14's Triton and fixed on 10.0's

The mirror image of finding 1. On 7.14 (Triton 3.7.1), 16 dsv4 tests in `tests/kernels/` fail with a Triton compiler crash:

```
Pipeline failed while executing [`TritonAMDGPUCanonicalizePointers` on 'tt.func' operation: @_sparse_attn_paged_kernel]
```

On 10.0 (Triton 3.8.0) all of them pass. So neither image currently runs everything: 7.14 loses dsv4 sparse attention, 10.0 loses MoE alignment. With the finding 1 fix, ROCm 10.0 is the fully working combination on this card.

### Finding 3: on the ROCm 10.0 image, the documented install silently replaces ROCm torch with the CUDA wheel

Running `pip install --no-build-isolation -e .` from docs/install.md on the 10.0 image: the image's Triton 3.8.0 violates `triton>=3.6,<3.8` in pyproject.toml, so the resolver swaps in PyPI's CUDA-only Triton. That invalidates ROCm torch's own `triton==3.8.0` dependency, which pulls in PyPI's CUDA `torch-2.11.0` plus about 20 nvidia-* packages. At that point torch is the CUDA build, so the GPU is unavailable and everything silently runs on CPU. The documented 7.14 image is fine (its Triton 3.7.1 satisfies the pin), but the very next ROCm image line breaks with no error. My workaround: `pip install --no-build-isolation --no-deps -e .`, then the remaining deps minus torch/triton. Raising the ceiling to `<3.9` on ROCm would resolve it, and 3.8 is also what fixes finding 2.

### Finding 4: the arch table in hip_compat.h mislabels the RDNA3 cards

`python/freetoken/kernel/csrc/include/freetoken/hip_compat.h:12` says `gfx1101 — RX 7900 GRE`. The 7900 GRE is Navi 31, which is gfx1100. gfx1101 is Navi 32 (RX 7800 XT / 7700 XT); this 7800 XT reports `gcnArchName == gfx1101` live. Likewise gfx1102 is Navi 33 (RX 7600/XT), and gfx1103 is the Phoenix APU iGPU, not a discrete card.

### Finding 5: the gemma4 router "mismatch" on RDNA3 is bf16 tie ordering, not a bug

One router test compares top-k expert selection against a reference. On this card the bf16 router logits contain exact ties (row 2's top scores include two identical `2.140625` values), so torch's top-k can legitimately order them differently than the reference does. A tolerance or tie-aware comparison would be more robust than treating it as an RDNA3 bug.

### NVFP4 runs end to end on RDNA3, including the MoE offload backend

Since this PR carries the e4m3 compat layer, I tried the hardest model I could fit: Qwen3.6-35B-A3B in NVFP4, about 21 GB of weights against this card's 16 GB. It worked without modification: the Triton NVFP4 dequant path compiled and ran on gfx1101, and the auto-selected `offload` MoE backend (experts in system RAM, hot set cached in VRAM) served it correctly. I haven't seen NVFP4 inference on RDNA3 reported anywhere yet, so this run at least confirms the compat layer holds up under a real workload.

Rough performance, single stream: 29.6 tok/s decode average over a 12-round streaming benchmark (256 tokens per round, 2 warmup rounds discarded), 1% low 24.0, TTFT 0.71 s. As a sanity check, a llama.cpp HIP build (b10729) serving the same model on the same card with expert layers on CPU reports 33.7 tok/s by its own server-side timing, so the offload path lands about 12% below that. The two engines offload differently (llama.cpp computes experts on the CPU; FreeToken streams them to the GPU) and FreeToken here still runs finding 1's fallback, so treat this as a plausibility check rather than a comparison.

---

I've kept full logs from both images and the benchmarks, and can re-run anything on gfx1101, test follow-up patches, or send the moe_align fix with its regression tests (10 tests, including graph-capture coverage) as a PR.

lukascechovic:
FreeToken serves `Qwen3.8-Flash-Next` on **AMD gfx1201 under Linux**, in production, and has done for
several days. As far as we can find as of 2026-09-02 this is the first Linux gfx1201 serving report —
#82's reports are Windows, and the ROCm branches here validate kernels and collectives but explicitly
stop short of serving. Happy to be corrected if someone got there first.

### What runs

- **2 x AMD Radeon AI PRO R9700** (`gfx1201`, 32 GB each) — ⚠️ **one card is used**; `qwen4_exp` is
  gated to TP=1
- Linux, PyTorch 2.11.0+rocm7.14.0, HIP 7.14.60850
- FreeToken `main@4b94bdc3` + PR **#132** + **#133** + **#134**, plus one local patch (below)
- `RadixArk/Qwen3.8-Flash-Next-NVFP4`, `--moe-backend offload --expert-load parallel`
- 122 GiB host RAM; the model pins **111.4 GiB** of it

⭐ The **#132 + #133 + #134** composition had never been compiled by anyone — #133 and #134 are
siblings off #132, not a chain. It compiles clean and it serves.

### Three flags are mandatory or it does not serve

Recording these because each one cost us a debugging session:

- `--expert-load parallel` — the serial path is unusably slow here
- `PYTORCH_ALLOC_CONF=expandable_segments:False` — ⛔ `expandable_segments:True` produces a
  `rocdevice.cpp` **Memory Fault Error** on gfx1201, not an OOM, and NCCL broken-pipe noise hides the
  real line
- `HIP_VISIBLE_DEVICES=0`

### One patch is required

Stock, the server answers prompts of **at most 15 tokens** — a 16-token prompt kills the worker. The
QSA prefill kernel asks for 65,792 B of LDS against gfx1201's 65,536 B. Filed separately with the
arithmetic, a reproducer and a 10-line HIP-gated clamp. ⛔ Every prefill number below therefore
carries *"`BLOCK_N` clamped to 32"*, and is **not** comparable to a CUDA run of the same engine, which
would use the GB300-tuned tile.

### It loads and it is stable

```
--moe-cache-auto resolved moe_cache_size=5960 num_pages=1720 (prefill_overlap=True)
Allocating 110080 tokens for KV cache, K + V = 2.60 GiB
```

CUDA-graph capture works at `[1]` and `[1,2,4]`. ⭐ Notably the graph-replay MoE crash that #82 reports
on Windows **does not reproduce on Linux**: captured decode measured 27.952 tok/s against eager
18.459 — a **1.51x** win that Windows users are currently paying for.

Cold load is **88–191 s** and the first request after a cold swap returns 503 by design.

### Performance, against llama.cpp on the same box

The comparison people will actually want. llama.cpp is running the same model as Q4_K_XL with
`--n-cpu-moe`, on **both** cards; FreeToken is NVFP4 on **one**:

| ctx | FreeToken TTFT | llama.cpp TTFT | FreeToken decode | llama.cpp decode |
|---|---:|---:|---:|---:|
| 10k | **9.42 s** | 37.19 s | **26.87 tok/s** | 16.18 tok/s |
| 50k | **45.44 s** | 240.14 s | **26.92** | 12.13 |
| 100k | **89.20 s** | 593.40 s | **26.60** | 9.72 |

⭐ FreeToken wins **both axes at every context measured, from one card against two**. And the decode
is **flat** — x1.010 over 10x of context, where llama.cpp decays x1.665 — so the margin widens as the
prompt grows. That flatness held all the way to 262k in a later run (x1.009 over x25.7 of context).

⚠️ Caveats we would rather state than have found: these are half-box numbers on clamped tiles;
different quantisations (NVFP4 vs Q4_K_XL) so this is a **row-versus-row** comparison, not a kernel
one; and we have no fidelity instrument, so **nothing here is a quality claim in either direction**.

### Two other measurements, in case they are useful

- `ft bench bw`'s `pcie_linear_h2d = 28.7 GB/s` reproduces our board's x8 gen5 ceiling **exactly**, as
  measured by two unrelated instruments. The tool is telling the truth on AMD.
- A 6,399-token prefill takes **3.83 s** and a 23,999-token prefill 17.6 s. ⚠️ The rate is **not
  monotonic** across our three points (639 → 1,669 → 1,365 tok/s) and we do not explain it; do not fit
  a single prefill throughput to three points.

### Kernel suite

**1471 passed, 27 failed, 78 skipped** on gfx1201. The failures are classified in a separate issue;
one of them is a real silent-correctness bug in `CpuMoeExecutor` under graph replay, which we have
filed. No `qwen4_exp` wrong-answer failure was found on four independent probes.

zihaomu:
Hi @lukascechovic, that's result looks great! Thx for your report.

lukascechovic:
Following up on the gfx1201 datapoint I left above: the tree those numbers came from is now public and reproducible — https://github.com/lukascechovic/FreeToken/tree/rocm-gfx1201. There's a Dockerfile.gfx1201 that builds it from a clean clone of the branch, no other inputs.

Not proposing it as-is; just so there's something concrete behind the RDNA4 numbers if anyone wants to reproduce them.

zihaomu:
Rebase refresh (2026-09-16):

- rebased the 9 foundation commits onto current upstream `main` at `cac247a`;
- resolved the only textual conflict in `pyproject.toml` by retaining upstream `torchvision` and Transformers 5.16 constraints while preserving the ROCm-compatible Triton `>=3.6,<3.8` range;
- verified `git diff --check`, Python AST parsing, and TOML parsing;
- ran the focused foundation tests in a PyTorch 2.11 / ROCm 7.14 container: `6 passed, 17 skipped` (GPU-required cases skipped because no device was attached);
- built and linked `_pinned_tensor`, `_cpu_moe`, and `_ple_store` successfully against the ROCm 7.14 modular SDK from a clean temporary source copy.

The refreshed head is `c2af4c6`. GitHub now reports the PR as mergeable/clean. The earlier physical R9700 results in the PR description were not rerun during this conflict-only refresh.

python/freetoken/kernel/csrc/jit/fast_index_copy.cuh
```diff
@@ -34,48 +34,72 @@ inline constexpr auto get_mem_package() {
...
+#if FREETOKEN_USE_ROCM
+    *dst = value;
+#else
     uint32_t tmp0 = value.x;
     uint32_t tmp1 = value.y;
     uint32_t tmp2 = value.z;
     uint32_t tmp3 = value.w;
     asm volatile("st.global.wt.v4.b32 [%0],{%1,%2,%3,%4};" ::"l"(dst), "r"(tmp0), "r"(tmp1), "r"(tmp2), "r"(tmp3));
+#endif
 }
```
jason-fxz, line 95: Perf only, not blocking.

The ROCm branch drops the streaming hint entirely. sglang's `hicache.cuh` already has the ROCm version via `__builtin_nontemporal_load/store` over `ext_vector_type`, e.g.:

```cpp
typedef uint32_t native_uint4 __attribute__((ext_vector_type(4)));
__builtin_nontemporal_store(__builtin_bit_cast(native_uint4, value), reinterpret_cast<native_uint4*>(dst));
```

Ref: https://github.com/sgl-project/sglang/blob/main/python/sglang/kernels/jit/csrc/kvcacheio/hicache.cuh

docs/install.md
```diff
@@ -2,7 +2,9 @@
 
 ## Requirements
```
jason-fxz, line 4: Please move this into a separate docs/install_amd.md (marked WIP, linking the AMD Support roadmap) and leave a one-line pointer at the top of install.md.

python/freetoken/kernel/utils.py
```diff
@@ -63,6 +72,69 @@ def _rank(a: str) -> int:
         cc = max(arch_list, key=_rank).rstrip("a").replace(".", "")
         flags = flags + [f"-gencode=arch=compute_{cc},code=compute_{cc}"]
     return flags
+
+
+def _hip_cflags(extra: List[str]) -> List[str]:
+    """HIP flags for a kernel build on ROCm."""
+    # TODO(ROCm): Triton autotune configs need RDNA-specific tuning (wave count, LDS size).
+    flags = DEFAULT_HIP_CFLAGS + extra
+    raw_arches = os.getenv("FREETOKEN_ROCM_ARCH") or os.getenv("PYTORCH_ROCM_ARCH", "")
```
jason-fxz, line 81: The requirement here is the same as #521 on the CUDA side: compile for the GPU this process is bound to and nothing else, with the arch in the build hash so different cards never share a cached .so. Right now the arch comes from FREETOKEN_ROCM_ARCH / PYTORCH_ROCM_ARCH (the latter is torch's build-time variable, typically preset to a multi-arch list in the ROCm images) or from current_device(), and tvm-ffi's own ocm_agent_enumerator probe adds every agent on the machine on top. Suggest resolving the arch once from the bound GPU's gcnArchName and pinning it through TVM_FFI_ROCM_ARCH_LIST (as _pin_tvm_ffi_arch_ctx does for CUDA) so tvm-ffi stops probing; keep the --offload-arch in extra_cuda_cflags, since tvm-ffi only hashes the extra flags and that is what keeps the arch in the cache key.

jason-fxz (changes requested):

python/freetoken/kernel/csrc/jit/fast_index_copy.cuh
jason-fxz: also provide output of `benchmarks/bench_offload_cache_copy.py`, `bw_GBps` should achieve 90%+ of `ft bench bw` PCIe ceilings

zihaomu: Hi @jason-fxz, thanks for code review. @LZ-QWQ Please take a look.

python/freetoken/kernel/utils.py
zihaomu: Fixed at new commit https://github.com/FlashML-org/FreeToken/pull/132/commits/dc4eebf34a775c13ce139baf6b123bd2d3346062.

python/freetoken/kernel/csrc/jit/fast_index_copy.cuh
LZ-QWQ: We updated the ROCm `load_nc/store_nc` implementation in [c6be4af](https://github.com/FlashML-org/FreeToken/pull/132/commits/c6be4afb541b34f36fb811cd2f6efc01784e9e5d) as requested, using Clang non-temporal builtins. 
On W7900D, `bench_offload_cache_copy.py` achieved 93.39–99.99% of the 27.06 GB/s PCIe ceiling, while `measure_pcie_gather_bw` reached 27.14 GB/s (fused=true). 
For completeness, we also noted two implementation details:
1. Clang’s `__builtin_nontemporal_load/store` SGLang’s [L1 no-allocate](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#id746) intent rather than FreeToken’s [write-through](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#id745) semantics.
2. `measure_pcie_gather_bw` selects `fast_index_copy_multi_jit` when `fused=true`; its `fast_index_copy_multi` kernel uses direct uint4 loads/stores and does not call load_nc or store_nc.

PR #134, opened 2026-08-24
fix(rocm): gate CUDA-only optional backends on ROCm

zihaomu:
> **Draft follow-up to #132.** This branch is based on the #132 head. Until #132 merges, GitHub Files changed also includes the foundation diff; the incremental review scope is three files.

## Summary

- never select `flashinfer`, `sgl_kernel`, OpenAI `triton_kernels`, Marlin, or other NVIDIA donor binaries on ROCm;
- explicitly route NVFP4 `auto`/`triton` requests to Triton on ROCm;
- fail clearly for forced `marlin`/`flashinfer` requests without probing or importing donor packages;
- avoid querying CUDA driver/capability paths on an AMD runtime;
- remove the unused `driver_hip_version()` helper;
- retain existing Triton/PyTorch fallback paths and CUDA selection behavior.

## Incremental review scope

Relative to `zihaomu:feat/rocm-rdna3-rdna4-foundation`:

- `python/freetoken/kernel/backend.py`
- `python/freetoken/moe/nvfp4_backends.py`
- `tests/kernels/test_backend.py`

Diff size: 107 additions, 11 deletions.

## Dependency

- Depends on #132.
- Independent of the JIT and RCCL follow-ups.

## Validation

- backend-selection tests: `8 passed`
- ROCm NVFP4 tests cover `auto`, `triton`, `marlin`, and `flashinfer`;
- tests assert ROCm never executes package probes or donor imports;
- gfx1201 Triton attention, norm, and activation fallback tests: PASS.

## Compatibility

The CUDA path continues to probe and select the same optional packages. This remains Draft until #132 establishes the shared ROCm foundation.

PR #133, opened 2026-08-24
fix(rocm): make TVM-FFI index and store JIT kernels portable to HIP

zihaomu:
> **Draft follow-up to #132.** This branch is based on the #132 head. Until #132 merges, GitHub Files changed also includes the foundation diff; the incremental review scope is the four files listed below.

## Summary

- disable NVIDIA programmatic-dependent-launch instructions and launch attributes on ROCm while preserving the CUDA path;
- accept `kDLROCM` tensors in the generic index and store TVM-FFI kernels;
- keep the CUDA PTX implementation unchanged and use the HIP-compatible common launch path;
- use `kWarpThreads` as the single source for wave-size assertions;
- cover split=1/2/4 row copies, masked vocabulary misses, both index widths, and both store index widths.

## Incremental review scope

Relative to `zihaomu:feat/rocm-rdna3-rdna4-foundation`:

- `python/freetoken/kernel/csrc/include/freetoken/utils.cuh`
- `python/freetoken/kernel/csrc/jit/index.cu`
- `python/freetoken/kernel/csrc/jit/store.cu`
- `tests/kernels/test_jit_index_store.py`

Diff size: 88 additions, 14 deletions.

## Dependency

- Depends on #132.
- Does not include optional-backend gating, RCCL routing, or GGUF support.

## Validation

- GPU: AMD Radeon AI PRO R9700 (`gfx1201`)
- PyTorch `2.11.0+rocm7.14.0`, HIP `7.14.60850`
- cold JIT plus pinned-memory focused run: `14 passed`
- warm `tests/kernels/test_jit_index_store.py`: `6 passed`
- covered fp32 D=256 (split=2), D=512 (split=4), lower/upper masked reset, int64 indexing, and int32/int64 store indices.

## Compatibility

CUDA branches remain behind the non-ROCm preprocessor path. Physical NVIDIA regression testing is not claimed; this PR remains Draft until #132 and CUDA CI establish the final base.

PR #135, opened 2026-08-24
fix(rocm): route tensor parallel communication through RCCL

zihaomu:
> **Draft follow-up to #132.** This branch is based on the #132 head. Until #132 merges, GitHub Files changed also includes the foundation diff; the incremental review scope is three files.

## Summary

- disable the NVIDIA-only custom PyNCCL communicator for multi-GPU ROCm runs;
- initialize PyTorch distributed with backend `nccl`, which the ROCm wheel implements through RCCL;
- retain the existing PyNCCL path on CUDA;
- make direct PyNCCL module loading fail immediately on ROCm instead of reaching a misleading `-lnccl` linker failure;
- validate an actual two-rank collective through `Engine._init_communication()`.

PyNCCL itself is deliberately bypassed, not ported to RCCL.

## Incremental review scope

Relative to `zihaomu:feat/rocm-rdna3-rdna4-foundation`:

- `python/freetoken/engine/engine.py`
- `python/freetoken/kernel/pynccl.py`
- `tests/engine/test_rocm_communication.py`

Diff size: 151 additions, 2 deletions.

## Dependency

- Depends on #132.
- Independent of the JIT and optional-backend follow-ups.

## Validation

- two AMD Radeon AI PRO R9700 (`gfx1201`) GPUs, PyTorch `2.11.0+rocm7.14.0`, HIP `7.14.60850`;
- `tests/engine/test_rocm_communication.py`: `4 passed` with both GPUs visible;
- actual `Engine._init_communication()` selected the default `nccl` backend (RCCL), created the CPU `gloo` subgroup, and produced all-reduce result `3.0` on both ranks;
- unit coverage also proves direct PyNCCL loading fails before native compilation/linking on ROCm.

## Known limitation

Full multi-GPU model serving with real weights is not claimed yet. The PR remains Draft until #132 establishes the common ROCm foundation.


lukascechovic:
Confirming this from a second pair of R9700s (`gfx1201`, Linux), and adding the number the PR does not
claim: what the torch/RCCL fallback actually costs at a real model's decode shape.

We hit the pynccl break independently while trying to measure the collective through the engine's own
transport, and our reading of it matches yours exactly:

```python
# freetoken/kernel/pynccl.py:29-31
@functools.cache
def _load_nccl_module() -> Module:
    # TODO(ROCm): NCCL -> RCCL migration for multi-GPU tensor parallelism on AMD.
    return load_aot("pynccl", cuda_files=["pynccl.cu"], extra_ldflags=["-lnccl"])
```

Two independent breaks in that one call: the vendored `csrc/include/freetoken/nccl227.h` is NVIDIA's
own NCCL 2.27 header with **zero** `__HIP`/`hip_runtime` guards, so the build dies at
`fatal error: 'cuda_runtime.h' file not found`; and `-lnccl` has nothing to link against either,
because a ROCm image ships `librccl.so.1`, not `libnccl.so`.

⭐ Worth adding to the PR's rationale: the reason this sat unreported is
`freetoken/distributed/impl.py:79-80` — `enable_pynccl_distributed` returns before touching the
builder when `tp_info.size == 1`. So a TP=1 server parses `--use-pynccl`, records `use_pynccl=True` in
`ServerArgs`, logs it, and never trips on it. Our deployed row does exactly that. The failure only
fires at `tp_size = 2`, which on AMD nobody had reached. Your "fail immediately on ROCm instead of
reaching a misleading `-lnccl` linker failure" is the right call — the current error points at a
missing library rather than at an unported header.

### The fallback performs fine — measured

Since the PR routes through `TorchDistributedImpl` / PyTorch's bundled RCCL, here is what that path
costs on two R9700s at the shape that matters. `Qwen3.8-Flash-Next` decode is one token through 48
layers, so the collective is `[1, 2560]` bf16 = **5,120 bytes** per call:

| | µs/call | x48 layers |
|---|---:|---:|
| back-to-back, 15 banked readings | **28.6 – 41.9** | **1.37 – 2.01 ms/token** |
| host-synced either side (upper bound) | 72.4 | 3.48 ms/token |

The back-to-back figure is the honest one — the host-synced arm charges ~30 µs of synchronisation the
engine never pays. A tuning sweep (`NCCL_PROTO` LL/LL128/Simple, `NCCL_ALGO` Tree/Ring,
`NCCL_P2P_DISABLE`) moved nothing: every setting landed inside the same 28–42 µs band, which is
run-to-run jitter, not a tuning win.

RCCL 2.30.4 chose direct peer-to-peer with no host staging, which we did not expect on this board:

```
Channel 00/0 : 0[3000] -> 1[7000] via P2P/IPC comm 0x17c0dbe0 nRanks 02
```

⛔ These two cards have **no XGMI** — they are two hops apart through the AM5 root complex, on x8 gen5
links. `torch.cuda.can_device_access_peer(0, 1)` is `True` and RCCL takes it anyway.

⇒ the fallback is not a degraded mode to apologise for. It is fast enough that a per-layer all-reduce
is a real design option on AMD.

### One caveat on our numbers

⚠️ 15 of our 16 readings have a banked artifact and span 28.6–41.9 µs. The 27.8 µs low comes from the
tuning table, whose per-setting raw was overwritten; we quote the range rather than the minimum.

### Where we are, in case it is useful to your step 5

We serve `qwen4_exp` (`RadixArk/Qwen3.8-Flash-Next-NVFP4`, `--moe-backend offload`) on **one** of these
two cards under Linux, in production, at a 262,144-token context. So we have some ground truth for the
"full model-serving validation" you have staged as step 5 of #132, and we are glad to run things
against your branches.

⚠️ We have also taken TP=2 for this model as far as it goes today, and it does not load — for reasons
below your layer, not in it. Details in a separate report so as not to derail this PR; the short
version is that `qwen4_exp`'s own tensor-parallel path is unfinished, and separately that replicated
offload banks want 2 × 111.4 GiB of host RAM. Neither is anything a working RCCL collective can fix,
and your PR removes a real obstacle that sits in front of both.

### Platform

AMD Radeon AI PRO R9700 x2 (`gfx1201`), Linux · PyTorch 2.11.0+rocm7.14.0, HIP 7.14.60850, RCCL 2.30.4
· FreeToken `main@4b94bdc3` + #132 + #133 + #134 · gfx1201 `BLOCK_N` clamped to 32 in the QSA attend
kernel (see our separate LDS issue), which does not touch this measurement.

zihaomu:
@lukascechovic Thanks for the detailed independent validation. Your observation about TP=1 masking the issue is correct: use_pynccl defaults to true, but enable_pynccl_distributed returns before loading or building PyNCCL when the TP size is one.

I added an explicit regression test for that contract in f75a553. It verifies that the single-rank path stays on Gloo and does not probe the ROCm branch, initialize PyNCCL, or create a second process group.

I also rechecked the communication tests locally on one R9700 (gfx1201) with PyTorch 2.13 nightly and ROCm 7.14:

- tests/engine/test_rocm_communication.py: 4 passed, 1 skipped
- the TP=1 engine path initialized Gloo without touching the PyNCCL loader
- a real single-rank PyTorch nccl process group initialized on ROCm and completed all_reduce successfully
- direct PyNCCL loading failed early with the intended NVIDIA-only error

The skipped case is the existing real two-rank RCCL test. I was limited to one GPU for this run, so I did not claim an independent two-GPU reproduction here.

I agree that changing only -lnccl to -lrccl would not be a sufficient port: the current custom module also depends on NVIDIA headers, CUDA runtime APIs, and NCCL-specific APIs. Keeping this PR on the PyTorch nccl process-group path backed by RCCL is therefore the appropriately scoped fix.

Your R9700 latency measurements are useful supporting microbenchmark evidence. I will keep the performance claim scoped accordingly: they demonstrate that the RCCL route is viable on that platform, but are not yet an end-to-end TP2 serving result until the model TP path itself is complete.

lukascechovic:
Following up on the scoping you gave here — you're right that this thread's numbers were transport-only.
The model TP path finished a few hours after your reply, and it serves: **+34.1% decode / −39.5% TTFT**
at 10k context (+31.5% / −40.4% at 100k) against the same box's TP=1 row, on two R9700s. Full numbers,
the mechanism behind it (your PR #311 disk-PLE was what actually unblocked the host-RAM budget — 111
GiB pinned → 64 GiB), and two things we have **not** closed (a greedy-text divergence we can't yet
diagnose without `logprobs`, and an 18-20 min cold load) are on
[#62](https://github.com/FlashML-org/FreeToken/issues/62#issuecomment-5523382144).

No quality claim attached — we don't have a fidelity instrument on this box, which is exactly why the
divergence item is still open. Wanted this on your radar since you're the one who scoped the gap it
fills.

### Platform

Same as above: AMD Radeon AI PRO R9700 x2 (`gfx1201`), Linux, 122 GiB host RAM · PyTorch
2.11.0+rocm7.14.0, HIP 7.14.60850, RCCL 2.30.4 · FreeToken `main@4b94bdc3` + #132/#133/#134 + #311 + our
sharding work.

zihaomu:
Dependency refresh (2026-09-16): rebased the three RCCL-specific commits onto the conflict-resolved #132 head `c2af4c6`. `git range-diff` reports all three commits as patch-equivalent. Focused communication tests in the ROCm 7.14 / PyTorch 2.11 container: `4 passed, 1 skipped`; the skipped test is the real two-rank RCCL GPU test because no devices were attached. New head: `500763c`; GitHub reports clean/mergeable.

Issue #123, opened 2026-08-24
Request hang (silent, zero-log) with hybrid_radix cache + moe-backend offload + nvfp4 triton backend — reproduced on 2 independent models, not explained by #110


hardwork9047:
## Summary

A ~2,400-token chat completion request hangs forever (no error, no log
line — not even "Prefill batch") with `cache_type='hybrid_radix'` +
`--moe-backend offload` + `nvfp4_backend='triton'`, on plenty of free
VRAM. The engine process pins one CPU thread at ~190% while GPU shows
100% utilization at only ~90W (an idle-spin signature, not active
compute), and the frontend never recovers — subsequent requests, even
trivial ones, also hang forever and the process must be `kill -9`'d.

I initially suspected this was the OOM-crash-then-zombie mechanism from
#110, so I tested mitigating it with `--memory-ratio` (0.9 → 0.8 → 0.5,
raising free-after-init VRAM from 1.82 GiB up to 10.08 GiB — roughly half
the GPU's total free memory). **The hang reproduced identically at every
memory-ratio setting**, which rules out #110's OOM-driven mechanism as the
cause here (10 GiB free is far beyond any plausible workspace-allocation
OOM for a ~2,400-token prompt). I then reproduced the exact same hang on a
**second, independent model** with the same architecture family
(hybrid attention + mamba/linear-state layers, `hybrid_radix` cache),
which points at something specific to this cache-type/backend combination
rather than either model's weights.

## Environment

- FreeToken `0.1.2` (PyPI install), torch `2.11.0+cu130`
- NVIDIA driver `595.71.05`, 2× RTX 4090 24GB (sm_89, no NVLink), only
  `CUDA_VISIBLE_DEVICES=0` used
- Ubuntu 22.04, kernel `6.8.0-124-generic`
- Models tested (both hang identically):
  - `nvidia/Qwen3.6-35B-A3B-NVFP4`
  - `ornith-ai/Ornith-1.5-35B-A3B-NVFP4`
- Launch flags (both models):
  ```
  ft serve --model-path <local snapshot> \
    --max-running-requests 1 \
    --max-seq-len-override 32768 \
    --moe-backend offload \
    --moe-cache-size 1024 \
    --host 0.0.0.0 --port 1920
  ```
- Resolved config (both models, identical): `moe_backend='offload'`,
  `attention_backend='fi'`, `cache_type='hybrid_radix'`,
[...]

Issue #72, opened 2026-08-23
Prefill hangs on RTX 4060 Ti (SM89) for prompts beyond warmup length


2stacks:
## Summary
 
On an RTX 4060 Ti 16GB (Ada, **SM89**), a ~180-token prompt sent to a freshly started server hangs indefinitely. An ~11-token prompt works and decodes at 50–66 tok/s.
 
While hung: no `Prefill batch` line, no access-log line, connection stays open. GPU sits flat at 100% utilization / ~35W of 165W. Scheduler thread spins at ~100% CPU. SIGINT/SIGTERM are ignored; `kill -9` required.
 
**Is SM89 desktop Ada a tested configuration?** The paper benchmarks an RTX 4060 laptop, RTX 5090, and RTX PRO 6000.
 
## Environment
 
- FreeToken `0.1.2+g0ab982f10` (Desktop installer engine wheel)
- RTX 4060 Ti 16GB (SM89), driver 590.48.01, Ubuntu 24.04
- CUDA 13.0 toolkit installed, `nvcc` on PATH and confirmed in use
- `torch==2.11.0+cu130`, `flashinfer-python==0.6.17`, `sglang-kernel==0.4.5+cu130`, `freetoken-kernel-cache==0.1.2+cu130.g0ab982f10`
- Model: `nvidia/Qwen3.6-35B-A3B-NVFP4` (FTW)
## Reproduction
 
Start a fresh server:
 
```
ft serve --model ~/.freetoken/models/Qwen3.6-35B-A3B-NVFP4 \
         --served-model-name qwen3.6-35b-a3b
```
 
Send a ~180-token prompt as the **first** request:
 
```bash
python3 -c 'import json;p="The quick brown fox jumps over the lazy dog. "*20;print(json.dumps({"model":"qwen3.6-35b-a3b","messages":[{"role":"user","content":p}],"max_tokens":32}))' > /tmp/p.json
 
curl http://127.0.0.1:1919/v1/chat/completions -H 'Content-Type: application/json' -d @/tmp/p.json
```
 
Never returns. Restart required.
 
For contrast, an ~11-token prompt (`"hi"`) as the first request completes normally.
 
Startup logs `Prefill warmup complete for lengths [80, 128]`.
 
## Request history affects the outcome
 
[...]

PR #169, opened 2026-08-25
fix(engine): load prefill triton kernels at startup, not mid-request

jason-fxz:
Two changes, one goal: no triton compile / cuModuleLoadData on the request path.

1. `fla/l2norm`: `do_not_specialize=["T"]`, `T` was `tl.constexpr`: a new module per unseen prompt length, forever.
2. `_warmup_prefill` runs on every attention backend (was `triton`-only, so `fi`/`fa` had no prefill warmup), over a config-derived ladder covering every in-repo size bucket instead of `[80, 128]`. New `--skip-prefill-warmup`; `FREETOKEN_WARMUP_MAX_LEN` caps the ladder.

Suspected trigger of the sm_89 silent hangs (#123, #72; #31 has a py-spy stack inside cuModuleLoadData): a first-launch module load mid-forward on the scheduler thread.

Tested on H100 (driver 580.95.05, torch 2.11.0+cu130), `nvidia/Qwen3.6-35B-A3B-NVFP4`, offload + fi: l2norm bitwise equal, same latency; a 11..2560-token sweep loads zero in-repo kernels at request time (main: 4-10 per new size); first-request TTFT on a cold cache 15.1 s -> 0.6 s; warmup adds 4.1 s to startup on a warm cache.

The hang does not reproduce on sm_90. This PR removes the trigger; confirming the fix needs
an sm_89 card.

Refs #123, #72.

### DIFF
--- a/python/freetoken/engine/config.py
+++ b/python/freetoken/engine/config.py
@@ -42,6 +42,8 @@ class EngineConfig:
     # CPU MoE backend (--moe-backend cpu): number of CPU worker threads computing
     # the decode experts. 0 = auto (physical cores). Ignored by other backends.
     moe_cpu_threads: int = 0
+    # skip the startup prefill warmup forwards; first requests then pay the kernel module loads (and the full JIT compile on a cold start)
+    prefill_warmup: bool = True
     # Hybrid CPU/GPU decode (--moe-backend offload only): which MoE layers decode on
     # the CPU executor instead of the GPU offload/PCIe path. Spec is an explicit id
     # list ("3,7,11"), a count ("8" -> 8 layers evenly strided across depth), or a
--- a/python/freetoken/engine/engine.py
+++ b/python/freetoken/engine/engine.py
@@ -10,6 +10,7 @@ import torch
 from freetoken.attention import AttnType, attention_backend_info, create_attention_backend
 from freetoken.core import Batch, Context, Req, set_global_ctx
 from freetoken.distributed import destroy_distributed, enable_pynccl_distributed, set_tp_info
+from freetoken.env import ENV
 from freetoken.gpu_select import gpu_identity
 from freetoken.layers import set_rope_device
 from freetoken.models import create_model, load_weight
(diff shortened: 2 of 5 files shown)


gdevenyi:
Tested on sm_89 (2x RTX 6000 Ada), which is the card this PR most wants a data point from:

|---|---|---|
| first request after boot, ~1k prompt, time to first token | 0.50 s | 0.49 s |
| single-stream / 8-conc decode | 90.8 / 326.0 tok/s | 90.1 / 326.9 tok/s |
| TTFT, ~1k prompt (warm) | 0.81 s | 0.81 s |

Model `RadixArk/Qwen3.8-Flash-Next-NVFP4` (qwen4_exp, `qsa_sparse` attention backend, TP=2, flags in the setup note below). The l2norm change is bitwise-neutral here as claimed, and the wider prefill warmup adds no visible boot time at this size. The cold first-request stall does not exist on this path to begin with: the QSA backend's kernels are already warmed at startup, so 0.5 s cold vs 0.8 s warm-with-load is just the empty machine. About the sm_89 hang you reference (#123, #72): this session booted the same model about 30 times on these Ada cards (TP=1 and TP=2, offload and hybrid) without a single hang, on main and with several PRs applied, so whatever triggers it needs more than sm_89 + first-launch JIT. Applies to main only with a 3-way merge (the branch is off an older base).

---

On this model the QSA sparse backend got no prefill warmup at all before this PR (the call was gated on the triton backend), so the first request of each size bucket compiled on the request path after a Triton-cache-cold start. With this PR: `Prefill warmup over 10 lengths: [9, 16, 17, 32, 33, 48, 65, 80, 103, 4096]`, 16.6 s at startup at TP=2 (offload, 8 running), then no compiles during the benchmark; serving numbers unchanged (99.7-102 tok/s single-stream, 344-355 at 8 concurrent). Merge note: it conflicts with #231 in engine.py only because both add a method at the same spot. The fla/l2norm despecialization is already on main, so only the warmup part applied here.

PR #184, opened 2026-08-25
fix(deps): use PyTorch 2.13 to fix SM89 hangs

### LINKED ISSUE #182: Bug in PyTorch 2.11 causes SM89 hangs
## Summary

On an RTX 4070 (SM89), `Qwen3.6-35B-A3B-NVFP4` can hang indefinitely during prefill with PyTorch 2.11 but works with PyTorch 2.13.

While hung, GPU utilization remains near 100% and the request never completes.

I suspect this to be the cause of the following existing issues:
- https://github.com/FlashML-org/FreeToken/issues/72
- https://github.com/FlashML-org/FreeToken/issues/123

An early clue was that the same workload completed when FreeToken was run with:

```bash
CUDA_LAUNCH_BLOCKING=1 uv run ft serve ...
```

I isolated the cause to a row-wise FP8 `torch._scaled_mm` call in the fused QKV projection:

```text
(2010, 2048) @ (2048, 9216) -> (2010, 9216)
FP8 E4M3 inputs, BF16 output
scale_a=(2010, 1)
scale_b=(1, 9216)
```

This matches an upstream PyTorch SM89 bug fixed by [pytorch/pytorch@252bb4a](https://github.com/pytorch/pytorch/commit/252bb4a398db09ccd254dbb51672af311acabc46)

## Environment

- GPU: RTX 4070, SM89
- Driver: `580.142`
- CUDA: `13.0`
- Affected PyTorch: `2.11.0+cu130`
- Model: `Qwen3.6-35B-A3B-NVFP4`

## Reproduction scripts

I made two reproduction scripts

<details>
<summary>Standalone PyTorch reproducer</summary>

```python
#!/usr/bin/env python3
"""Minimal reproducer for an SM89 row-wise FP8 torch._scaled_mm hang."""

import os
import time

import torch


# M is the number of rows, K the input size, and N the output size.
M, N, K = 2010, 9216, 2048

# Repetition makes the intermittent failure reliable:
# 5/5 locally at 128 calls on an RTX 4070.
OPS = 128
TIMEOUT_SECONDS = 10

torch.manual_seed(0)
a = (
    torch.randn((M, K))
    .mul_(10)
    .clamp_(-448, 448)
    .to(device="cuda", dtype=torch.float8_e4m3fn)
)

torch.cuda.manual_seed_all(1)
b = (
    torch.randn((N, K), device="cuda")
    .mul_(32)
    .clamp_(-448, 448)
    .to(torch.float8_e4m3fn)
)

# These scale shapes select the row-wise scaling path.
scale_a = torch.ones((M, 1), device="cuda")
scale_b = torch.ones((1, N), device="cuda")

ready = torch.cuda.Event()
ready.record()
while not ready.query():
    time.sleep(0.01)

stream = torch.cuda.Stream()
events = [torch.cuda.Event() for _ in range(OPS)]

with torch.cuda.stream(stream):
    for event in events:
        out = torch._scaled_mm(
            a,
            b.t(),
            scale_a=scale_a,
            scale_b=scale_b,
            out_dtype=torch.bfloat16,
        )
        event.record(stream)

deadline = time.monotonic() + TIMEOUT_SECONDS
completed = 0

while time.monotonic() < deadline:
    while completed < OPS and events[completed].query():
        completed += 1

    if completed == OPS:
        print(f"completed all {OPS} calls")
        raise SystemExit(0)

    time.sleep(0.1)

print(
    f"GPU stalled: {completed}/{OPS} calls completed; "
    f"call #{completed + 1} did not complete within "
    f"{TIMEOUT_SECONDS} seconds",
    flush=True,
)

# Normal interpreter shutdown would wait for the stuck CUDA operation.
os._exit(124)
```

</details>

<details>
<summary>FreeToken reproducer</summary>

Start a fresh server:

```bash
uv run ft serve \
    --model /path/to/Qwen3.6-35B-A3B-NVFP4 \
    --host 127.0.0.1 \
    --port 1919 \
    --attention-backend triton,fi \
    --moe-backend offload \
    --disable-moe-prefill-overlap \
    --cuda-graph-max-bs 0
```

Then run:

```python
import json
import urllib.request


fragment = (
    "CUDA stream ordering and expert cache lifetime must remain correct "
    "across asynchronous prefill execution. "
)

body = json.dumps(
    {
        "model": "Qwen3.6-35B-A3B-NVFP4",
        "prompt": (
            "freetoken-prefill-hang-fp8-ab-fixed-0 unique-prefix 17\n"
            + fragment * 124
        ),
        "max_tokens": 1,
        "temperature": 0,
    }
).encode()

request = urllib.request.Request(
    "http://127.0.0.1:1919/v1/completions",
    data=body,
    headers={"Content-Type": "application/json"},
)

print(urllib.request.urlopen(request, timeout=60).read().decode())
```

With PyTorch 2.11, the request can time out while the GPU remains near 100% utilization.

</details>

## Results

```text
PyTorch 2.11: standalone reproducer hung in 5/5 runs
PyTorch 2.12: 0/5 hangs
PyTorch 2.13: 0/5 hangs
```

In FreeToken, the fused row-wise QKV operation hung in 5/5 fresh-server trials. Splitting it into tensor-wise Q, K, and V operations completed in 5/5 trials.

The split implementation was only a diagnostic control, not a proposed fix.

## Proposed fix

Update FreeToken dependency stack to:

```text
torch>=2.13,<2.14
triton==3.7.1
sglang-kernel==0.4.6.post1
```

This includes the upstream fix and aligns with [SGLang 0.5.18](https://github.com/sgl-project/sglang/blob/v0.5.18/python/pyproject.toml), which uses PyTorch 2.13 and `sglang-kernel==0.4.6.post1`

endenis:
Fixes #182

Possibly related to #72 and #123

## Summary

PyTorch 2.11 can hang indefinitely during prefill on SM89 GPUs.

I reproduced this with `Qwen3.6-35B-A3B-NVFP4` on an RTX 4070: the request stops making progress while GPU utilization remains near 100%.

This PR updates FreeToken to PyTorch 2.13 and along with these dependencies:

```text
torch>=2.13,<2.14
triton==3.7.1
sglang-kernel==0.4.6.post1
```

## Root cause

I isolated the non-completing GPU submission to the fused QKV projection’s row-wise FP8 `torch._scaled_mm` call:

```text
(2010, 2048) @ (2048, 9216) -> (2010, 9216)
FP8 E4M3 inputs, BF16 output
scale_a=(2010, 1)
scale_b=(1, 9216)
```

- CUDA events showed that the GPU reached this call but never completed it.
- `CUDA_LAUNCH_BLOCKING=1` avoided the hang but only by changing the timing (not a real fix).
- Splitting the projection into separate Q, K, and V calls also avoided the hang. This suggests the problem is in the row-wise FP8 path, not FP8 matrix multiplication in general.

The behavior matches an upstream PyTorch SM89 bug fixed by [pytorch/pytorch@252bb4a](https://github.com/pytorch/pytorch/commit/252bb4a398db09ccd254dbb51672af311acabc46)
The fix is present in PyTorch 2.12 and 2.13.

I made two reproduction scripts:
1. Standalone PyTorch reproducer
2. FreeToken reproducer

You can find them in the description of the #182 issue.

## Changes

- Require PyTorch 2.13 for FreeToken at runtime and during extension builds.
- Update Triton to `3.7.1` (required by PyTorch 2.13).
- Update `sglang-kernel` to `0.4.6.post1` which is built against the PyTorch 2.13.
- Apply the same PyTorch constraint to `freetoken-kernel-cache`.

There are no kernel or synchronization workarounds in this change.

## Testing

Tested on:

```text
GPU: RTX 4070 (Ada, SM89)
Driver: 580.142
CUDA: 13.0
Model: Qwen3.6-35B-A3B-NVFP4
```

Standalone `_scaled_mm` reproducer (attached in the  #182 issue):

```text
PyTorch 2.11: 5/5 runs hung
PyTorch 2.13: 0/5 runs hung
```

## Notes
I only have access to an RTX 4070. It would be useful to test on other cards as well.
Don't hesitate to reach out if you need more details


jason-fxz:
Closing: #243 fixes the sm_89 hang without bumping torch, so this is no longer needed for #182. Thanks for isolating the root cause.

PR #185, opened 2026-08-25
fix(kernel): launch GGUF MoE GEMV token axis on grid.x, not grid.z

### LINKED ISSUE #186: GGUF MoE GEMV crashes with cudaErrorInvalidValue once tokens*top_k exceeds 65535 (grid.z limit)
## Summary

`moe_vec_*_q8_1_cuda` (`python/freetoken/kernel/csrc/gguf/moe_vec.cuh`) launches the flat
`(token, top-k)` index as **grid.z**:

```c
const dim3 block_nums(block_num_y, 1, tokens * top_k);   // x19 launchers
```

`maxGridDimZ` is 65535 on every CUDA compute capability, so any call with
`tokens * top_k > 65535` fails the launch with `cudaErrorInvalidValue`.

For a GGUF MoE model with `top_k = 8` this caps a prefill batch at **8191 tokens** — and
`--max-extend-tokens` defaults to exactly **8192**. Serving `gemma-4-26B-A4B-it-Q4_0` therefore
dies on the first prompt long enough to fill one prefill chunk. It is not limited to a single
long prompt: the scheduler's token budget is per *batch*, so several concurrent medium prompts
packed into one batch hit it identically.

Proposed fix in #185: moves the flat `(token, top-k)` index to `grid.x` (limit 2³¹−1), adds a launch return-code check at each of the 19 launch sites, and adds a regression test that fails on `main` and passes with the change.

## Reproduction

Model-free — synthetic Q4_0 banks, so it needs no checkpoint download and runs in seconds.
Threshold is exact and deterministic: 8191 tokens works, 8192 fails.

```python
"""Minimal repro: ggml_moe_a8_vec fails once tokens*top_k exceeds 65535 (maxGridDimZ)."""
import ctypes, numpy as np, torch
from freetoken.kernel.gguf import ggml_moe_a8_vec
from freetoken.models.gguf.dequant import GGML_Q4_0

rt = ctypes.CDLL("libcudart.so")
rt.cudaDeviceGetAttribute.argtypes = [ctypes.POINTER(ctypes.c_int), ctypes.c_int, ctypes.c_int]
def attr(a):
    v = ctypes.c_int(0); rt.cudaDeviceGetAttribute(ctypes.byref(v), a, 0); return v.value
print(f"GPU={torch.cuda.get_device_name(0)}  maxGridDimX={attr(5)} Y={attr(6)} Z={attr(7)}")

E, H, NROWS, TOPK = 4, 256, 64, 8
rng = np.random.default_rng(0)
blocks = np.zeros((E, NROWS, H // 32, 18), dtype=np.uint8)      # Q4_0: {half d; uint8 qs[16]}
d = np.array([0.01], dtype=np.float16).view(np.uint8)
blocks[..., 0], blocks[..., 1] = d[0], d[1]
blocks[..., 2:] = rng.integers(0, 256, blocks[..., 2:].shape, dtype=np.uint8)
w = torch.from_numpy(blocks.reshape(E, NROWS, H // 32 * 18)).cuda()

for tokens in (8191, 8192):
    x = torch.randn((tokens, H), dtype=torch.bfloat16, device="cuda")
    ids = torch.randint(0, E, (tokens, TOPK), dtype=torch.int32, device="cuda")
    print(f"\n--- tokens={tokens}  grid.z = tokens*top_k = {tokens*TOPK} "
          f"({'<=' if tokens*TOPK <= 65535 else '>'} 65535) ---")
    y = ggml_moe_a8_vec(x, w, ids, TOPK, int(GGML_Q4_0), NROWS, tokens)
    torch.cuda.synchronize()
    print(f"    ok, out={tuple(y.shape)} finite={bool(torch.isfinite(y.float()).all())}")
```

### Full output, on a clean checkout of `main` @ `2757bb5`

```
GPU=Tesla T4  maxGridDimX=2147483647 Y=65535 Z=65535

--- tokens=8191  grid.z = tokens*top_k = 65528 (<= 65535) ---
    ok, out=(65528, 64) finite=True

--- tokens=8192  grid.z = tokens*top_k = 65536 (> 65535) ---
Traceback (most recent call last):
  File "/t.py", line 27, in <module>
    print(f"    ok, out={tuple(y.shape)} finite={bool(torch.isfinite(y.float()).all())}")
                                                                     ^^^^^^^^^
torch.AcceleratorError: CUDA error: invalid argument
Search for `cudaErrorInvalidValue' in https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html for more information.
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.
```

## The reported location is never the real one

None of the 19 launch sites in `moe_vec.cuh` checks the launch return code. A rejected launch
only *sets* the error flag, so it is reported by whatever CUDA call runs next. In the trace
above that is a `torch.isfinite` on the following line. While debugging a live server I saw the
same single fault attributed to three different places:

- `flashinfer.activation.gelu_tanh_and_mul` — the next CUDA call, via `act_fn` in `moe/fused_q4_0.py:44`
- the following `ggml_moe_a8_vec` at `moe/fused_q4_0.py:46`, once flashinfer was gated off
- a bare `torch.zeros` in an unrelated allocation

`CUDA_LAUNCH_BLOCKING=1` does **not** help, because the failure is in the launch *configuration*
rather than in execution. This cost me a lot of time and is why #185 also adds a check at each
launch site.

## Environment

| | |
|---|---|
| FreeToken | `0.1.2`, source build, tested at `2757bb5` (`git rev-parse --short HEAD`) |
| GPU | Tesla T4, 16 GiB, sm_75 |
| Driver / CUDA | 580.178.04 / CUDA 13.0.88, torch 2.11.0+cu130 |
| CPU / RAM | AMD EPYC 7V12 (4 vCPU) / 27 GiB |
| OS | Ubuntu 24.04.4 LTS, kernel 6.17.0-1022-azure (Azure VM), no WSL, no proxy |
| Checkpoint | `ggml-org/gemma-4-26B-A4B-it-GGUF` → `gemma-4-26B-A4B-it-Q4_0.gguf` (128 experts/layer, `top_k=8`, 30 layers) |
| Other GPUs | none |

Serving command under which this first showed up:

```
ft serve --model=<path>/gemma-4-26B-A4B-it-Q4_0.gguf --host=0.0.0.0 --port=1919 \
         --attention-backend=triton --memory-ratio=0.55 --max-running-requests=3 \
         --max-seq-len-override=16384 --num-tokens=49152 --moe-prefill-hit-d2d
```

Any prompt long enough to produce a full 8192-token prefill chunk killed the engine
(`Backend worker is gone and cannot be restarted`).

**One caveat, stated plainly:** the repro above is on an unmodified `main`, but the *serving*
path is not — a stock `main` cannot load this checkpoint on a T4 for two unrelated reasons (the
gemma4 GGUF loader hardcodes the token-embedding quant as Q6_K where this file stores Q8_0, and
the Triton decode attention tile overflows Turing's 64 KiB shared memory at `head_dim=512`). I
carry local patches for both. Neither touches `moe_vec.cuh` or anything on the path in the repro
script, and the script itself runs on a completely clean tree. Happy to open separate issues for
those two if useful.

## Not a duplicate of

- **#110 / #20** — those are `torch.OutOfMemoryError` in the prefill expert workspace. This is
  `cudaErrorInvalidValue` from a grid dimension over the device limit, at a fixed token count,
  with VRAM to spare.
- **#123** — that is a silent hang; this is a hard crash with a deterministic threshold.

## Related, but not reachable here

`moe.cuh` puts `tokens_post_padded / mmq_x` in grid.y and `mmvq.cuh` puts `nvecs` in grid.y.
Both can overflow the same 65535 cap in principle, though neither is reachable on my config
(`ggml_moe_a8` has no Python callers, and `nvecs` is bounded by `max_seq_len`). Note
`quantize_row_q8_1_cuda` in `gguf_kernel.cu` already tiles its y axis at 65535 — so the limit is
handled a few lines away, `moe_vec.cuh` just missed it. I left those alone to keep #185 to one change.

erichstuntebeck:
Fixes #186.

## Problem

`moe_vec_*_q8_1_cuda` launches the flat `(token, top-k)` index as **grid.z**:

```c
const dim3 block_nums(block_num_y, 1, tokens * top_k);   // moe_vec.cuh, x19 launchers
```

`maxGridDimZ` is 65535 on every CUDA compute capability, so a batch with
`tokens * top_k > 65535` fails the launch with `cudaErrorInvalidValue`.

With `top_k = 8` that caps a prefill batch at **8191 tokens** — and `--max-extend-tokens`
defaults to exactly **8192**. So a GGUF MoE model reliably dies on the first prompt long
enough to fill one chunk. It is not limited to one long prompt either: the scheduler's token
budget is per *batch*, so a few concurrent medium prompts packed together trip it identically
(observed here as `Prefill batch, #new-seq: 3, #new-token: 7299` — three streams, one batch).

This cost me a long time to find, because **none of the 19 launch sites check the launch
return code**. A rejected launch only sets the error flag, so it was reported by whatever
unrelated CUDA call ran next. I saw the same single fault blamed on three different places:

- `flashinfer.activation.gelu_tanh_and_mul` (the very next CUDA call, via `act_fn` in `fused_q4_0.py`)
- the following `ggml_moe_a8_vec` (once flashinfer was gated off)
- a bare `torch.zeros` in an unrelated allocation

`CUDA_LAUNCH_BLOCKING=1` does not help here — the failure is in the launch *configuration*,
not in execution.

## Fix

Move the flat index to `grid.x` (limit 2³¹−1) and rows to `grid.y`, whose extent is
`ceil(nrows / GGML_CUDA_MMV_Y)` and stays far below the cap. `quantize_row_q8_1_cuda` next
door already tiles its y axis at 65535 for the same reason, so the limit is known in this file's
neighbourhood — `moe_vec.cuh` just missed it.

Also adds `FT_MOE_VEC_LAUNCH_CHECK()` after each of the 19 launches, so a rejected
configuration reports itself with the geometry that caused it instead of latching.

In short, three things: the flat `(token, top-k)` index moves to `grid.x`; every launch site
gains a return-code check; and a regression test is added that fails on `main` and passes here.

## Test

`tests/kernels/test_gguf_moe_vec.py` runs a batch one token over the old cap
(`8192 * 8 = 65536`) and compares against the same batch split in half, each half under the
old limit — so it checks the result is *correct*, not merely that nothing raised.

```
# on main (2757bb5)
FAILED tests/kernels/test_gguf_moe_vec.py::test_moe_vec_above_grid_z_limit_matches_split_batches
E   torch.AcceleratorError: CUDA error: invalid argument
1 failed in 92.68s

# on this branch
1 passed in 89.20s
```

Both runs used a **fresh JIT build cache**. Worth flagging for reviewers: ninja compares
mtimes, so swapping `moe_vec.cuh` in place over a warm `~/.cache/torch_extensions` silently
reuses the old `.so` and the test passes when it should fail.

## Tested on

- **GPU**: Tesla T4 (sm_75, 16 GiB), driver 580.178.04, CUDA 13.0.88, torch 2.11.0+cu130
- **OS/toolchain**: Ubuntu 24.04 container, g++ 13.3, `-std=c++17` (built clean)
- **Checkpoint**: `ggml-org/gemma-4-26B-A4B-it-GGUF` → `gemma-4-26B-A4B-it-Q4_0.gguf`
  (128 experts/layer, `top_k=8`, 30 layers), served via `--moe-backend offload`
- **Command**:
  ```
  ft serve --model=<ckpt> --host=0.0.0.0 --port=1919 --attention-backend=triton \
           --memory-ratio=0.55 --max-running-requests=3 --max-seq-len-override=16384 \
           --num-tokens=49152 --moe-prefill-hit-d2d
  ```
- **End to end**: a 13,962-token prompt that killed the engine on every prior build now
  completes, and a 3-needle recall test passes 3/3 at 14,329 tokens. Server logs show
  `Prefill batch, #new-token: 8192` succeeding — the exact geometry that used to fail.

## Performance

Not the goal, but it measures faster rather than slower. Isolated kernel benchmark, median of
10 after 3 warmups, two independent runs per arm, each arm built from a clean cache
(`nrows=1408`, `hidden=2816`, `top_k=8`):

| tokens | main | this branch | delta |
|---|---|---|---|
| 2048 | 187.1 / 193.5 ms | 174.0 / 176.5 ms | ≈ −8% |
| 4096 | 380.0 / 394.3 ms | 346.1 / 355.1 ms | ≈ −9% |
| 8191 | 768.6 / 808.3 ms | 705.8 / 711.6 ms | ≈ −10% |

The mechanism is not root-caused, and this PR claims none. Swapping the axes changes which
blocks are co-scheduled, but the per-token expert varies, so the obvious weight-reuse story does
not straightforwardly hold. The numbers are recorded here as a measurement, not as a
justification for the change — the fix stands on the correctness bug alone.

## Notes

- Filed as #186, which carries the model-free repro script and full environment details.
- Only `q4_0` is exercised on my hardware, but all 19 launchers in the file shared the
  identical grid expression and are changed identically.
- Sibling kernels were checked for the same mistake: `moe.cuh` puts `tokens/mmq_x` in grid.y
  and `mmvq.cuh` puts `nvecs` in grid.y, both of which can overflow in principle but neither
  is reachable on my config. Left alone to keep this to one change.

### DIFF
--- a/python/freetoken/kernel/csrc/gguf/moe_vec.cuh
+++ b/python/freetoken/kernel/csrc/gguf/moe_vec.cuh
@@ -2,6 +2,23 @@
 // https://github.com/vllm-project/vllm/blob/4492e3a55428e161ca8db381edc28263e5da4c8d/csrc/quantization/gguf/moe_vec.cuh
 // copied and adapted from
 // https://github.com/ggerganov/llama.cpp/blob/b2899/ggml-cuda/mmvq.cu
+
+// A kernel launch rejected for its *configuration* (a grid dimension over the device
+// limit, an oversized block) does not raise -- it only sets the error flag. Nothing here
+// checked it, so the flag latched and was reported by whatever unrelated CUDA call ran
+// next: the grid.z overflow fixed below surfaced as a failure inside flashinfer's
+// activation kernel, inside the next ggml_moe_a8_vec, and even inside a torch::zeros.
+#ifndef FT_MOE_VEC_LAUNCH_CHECK
+#define FT_MOE_VEC_LAUNCH_CHECK()                                                        \
+  do {                                                                                   \
+    cudaError_t err_ = cudaGetLastError();                                               \
+    TORCH_CHECK(                                                                         \
+        err_ == cudaSuccess,                                                             \
+        "moe_vec launch failed: ", cudaGetErrorString(err_),                             \
+        " (grid=", block_nums.x, ",", block_nums.y, ",", block_nums.z,                   \
+        " block=", block_dims.x, ",", block_dims.y, ",", block_dims.z, ")");             \
+  } while (0)
+#endif
 template <typename scalar_t, int qk, int qi, typename block_q_t, int vdr, vec_dot_q_cuda_t vec_dot_q_cuda>
 static __global__ void moe_vec_q(
     const void* __restrict__ vx,
@@ -12,10 +29,13 @@ static __global__ void moe_vec_q(
     const int ncols,
     const int nrows,
     const int token_stride) {
-  const auto row = blockIdx.x * blockDim.y + threadIdx.y;
+  // The flat (token, top-k) index rides blockIdx.x, not .z: it reaches tokens*top_k,
+  // which exceeds the 65535 cap on grid.y/.z (grid.x allows 2^31-1). Rows move to .y,
+  // whose extent is ceil(nrows / GGML_CUDA_MMV_Y) and stays far below the cap.
+  const auto row = blockIdx.y * blockDim.y + threadIdx.y;
 
-  const auto token = blockIdx.z / topk;
-  const auto expert = (topk_ids)[blockIdx.z];
+  const auto token = blockIdx.x / topk;
+  const auto expert = (topk_ids)[blockIdx.x];
 
   if (row >= nrows) {
     return;
@@ -47,7 +67,7 @@ static __global__ void moe_vec_q(
   }
 
   if (threadIdx.x == 0) {
-    dst[blockIdx.z * nrows + row] = tmp;
+    dst[blockIdx.x * nrows + row] = tmp;
   }
 }
 
@@ -64,10 +84,11 @@ static void moe_vec_q4_0_q8_1_cuda(
     const int token_stride,
     cudaStream_t stream) {
   const int block_num_y = (nrows + GGML_CUDA_MMV_Y - 1) / GGML_CUDA_MMV_Y;
-  const dim3 block_nums(block_num_y, 1, tokens * top_k);
+  const dim3 block_nums(tokens * top_k, block_num_y, 1);
   const dim3 block_dims(WARP_SIZE, GGML_CUDA_MMV_Y, 1);
   moe_vec_q<scalar_t, QK4_0, QI4_0, block_q4_0, VDR_Q4_0_Q8_1_MMVQ, vec_dot_q4_0_q8_1>
       <<<block_nums, block_dims, 0, stream>>>(vx, vy, dst, topk_ids, top_k, ncols, nrows, token_stride);
+  FT_MOE_VEC_LAUNCH_CHECK();
 }
 
 template <typename scalar_t>
@@ -83,10 +104,11 @@ static void moe_vec_q4_1_q8_1_cuda(
     const int token_stride,
     cudaStream_t stream) {
   const int block_num_y = (nrows + GGML_CUDA_MMV_Y - 1) / GGML_CUDA_MMV_Y;
-  const dim3 block_nums(block_num_y, 1, tokens * top_k);
+  const dim3 block_nums(tokens * top_k, block_num_y, 1);
   const dim3 block_dims(WARP_SIZE, GGML_CUDA_MMV_Y, 1);
   moe_vec_q<scalar_t, QK4_0, QI4_1, block_q4_1, VDR_Q4_1_Q8_1_MMVQ, vec_dot_q4_1_q8_1>
       <<<block_nums, block_dims, 0, stream>>>(vx, vy, dst, topk_ids, top_k, ncols, nrows, token_stride);
+  FT_MOE_VEC_LAUNCH_CHECK();
 }
 
 template <typename scalar_t>
@@ -102,10 +124,11 @@ static void moe_vec_q5_0_q8_1_cuda(
     const int token_stride,
     cudaStream_t stream) {
   const int block_num_y = (nrows + GGML_CUDA_MMV_Y - 1) / GGML_CUDA_MMV_Y;
-  const dim3 block_nums(block_num_y, 1, tokens * top_k);
+  const dim3 block_nums(tokens * top_k, block_num_y, 1);
   const dim3 block_dims(WARP_SIZE, GGML_CUDA_MMV_Y, 1);
   moe_vec_q<scalar_t, QK5_0, QI5_0, block_q5_0, VDR_Q5_0_Q8_1_MMVQ, vec_dot_q5_0_q8_1>
       <<<block_nums, block_dims, 0, stream>>>(vx, vy, dst, topk_ids, top_k, ncols, nrows, token_stride);
+  FT_MOE_VEC_LAUNCH_CHECK();
 }
 
 template <typename scalar_t>
@@ -121,10 +144,11 @@ static void moe_vec_q5_1_q8_1_cuda(
     const int token_stride,
     cudaStream_t stream) {
   const int block_num_y = (nrows + GGML_CUDA_MMV_Y - 1) / GGML_CUDA_MMV_Y;
-  const dim3 block_nums(block_num_y, 1, tokens * top_k);
+  const dim3 block_nums(tokens * top_k, block_num_y, 1);
   const dim3 block_dims(WARP_SIZE, GGML_CUDA_MMV_Y, 1);
   moe_vec_q<scalar_t, QK5_1, QI5_1, block_q5_1, VDR_Q5_1_Q8_1_MMVQ, vec_dot_q5_1_q8_1>
       <<<block_nums, block_dims, 0, stream>>>(vx, vy, dst, topk_ids, top_k, ncols, nrows, token_stride);
+  FT_MOE_VEC_LAUNCH_CHECK();
 }
 
 template <typename scalar_t>
@@ -140,10 +164,11 @@ static void moe_vec_q8_0_q8_1_cuda(
     const int token_stride,
     cudaStream_t stream) {
   const int block_num_y = (nrows + GGML_CUDA_MMV_Y - 1) / GGML_CUDA_MMV_Y;
-  const dim3 block_nums(block_num_y, 1, tokens * top_k);
+  const dim3 block_nums(tokens * top_k, block_num_y, 1);
   const dim3 block_dims(WARP_SIZE, GGML_CUDA_MMV_Y, 1);
   moe_vec_q<scalar_t, QK8_0, QI8_0, block_q8_0, VDR_Q8_0_Q8_1_MMVQ, vec_dot_q8_0_q8_1>
       <<<block_nums, block_dims, 0, stream>>>(vx, vy, dst, topk_ids, top_k, ncols, nrows, token_stride);
+  FT_MOE_VEC_LAUNCH_CHECK();
 }
 
 template <typename scalar_t>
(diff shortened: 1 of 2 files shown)


salekseev:
Independent confirmation on Ada (sm_89) with a different model, plus a data point on the
misattribution you describe.

## Before the patch

FreeToken 0.1.2 (PyPI wheel), RTX 4080 SUPER 16 GB, driver 610.57.04 / CUDA UMD 13.3,
serving `unsloth/gemma-4-26B-A4B-it-qat-GGUF` (Q4_0, `top_k = 8`, 30 layers x 128 experts).

Any prompt long enough to fill a default 8192-token prefill chunk killed the worker, and it
never came back:

```
RuntimeError: Check failed: (err == cudaSuccess) is false: Failed to launch kernel: invalid argument
  __tvm_ffi_gelu_tanh_and_mul -> gelu_tanh_and_mul.cu:44
Backend supervisor: backend worker freetoken-TP0-scheduler exited
Backend worker is gone and cannot be restarted; stopping the API server
```

I spent an afternoon on this before finding #186, and landed in exactly the trap your PR
description calls out: the fault is reported by the *next* CUDA call, so I had it filed in my
own notes as "a bad launch geometry in the AOT `gelu_tanh_and_mul` kernel" — the innocent
bystander. Your explanation (no launch-return check at any of the 19 sites, so the flag
latches) is the missing piece. The `FT_MOE_VEC_LAUNCH_CHECK()` addition is arguably the more
valuable half of this PR for anyone who hits it next.

Client-visible behaviour is also worth noting: the request does not error, it hangs — the
process exits underneath an open connection, so a caller sees a 300 s timeout, and
`docker ps` still reports the container up while `nvidia-smi` shows the VRAM gone.

Measured boundary, one prompt per server since the first failure is terminal:

| prompt tokens | default 8192 chunk | `--max-prefill-length 4096` |
|---|---|---|
| 32215 | OK, 4072 tok/s | — |
| 35415 | **worker died** | OK |
| 61175 | — | OK, 4437 tok/s |

`4096 * 8 = 32768` stays under the 65535 cap, which is why halving the chunk was a complete
workaround — consistent with your `tokens * top_k` analysis.

## After the patch

Applied to the installed tree (the `.cuh` is JIT-compiled, so no wheel rebuild is needed —
though the `TORCH_EXTENSIONS_DIR` cache has to be busted, since ninja can otherwise keep a
`.so` built from the pre-patch header). `--max-prefill-length` back to its 8192 default:

| prompt tokens | result | prefill tok/s |
|---|---|---|
| 35415 | OK | 773 |
| 51543 | OK | 1958 |
| 77303 | OK | 1570 |
| 103063 | OK | 1786 |

Engine still `{"status":"ok"}` afterwards, uptime 14180 s. Decode unaffected (150 tok/s at
400 tokens, 142 at 1200). So on this box the patch removes the cap outright rather than
moving it.

Caveat on attribution: that run also carried #103 (`--kv-cache-dtype q8_0`) because I was
testing both, and a 131072-token KV pool. Neither touches the launch geometry, and the
before/after boundary above was measured on bf16 KV with a 65536 pool, but flagging it so
nobody reads more into the numbers than they support.

Also worth flagging for anyone hitting this on a GGUF MoE: the practical trigger is not
"someone sent a huge prompt". `--max-extend-tokens` defaults to 8192 and `top_k = 8` is
ordinary, so **the default configuration of a GGUF MoE model is one long prompt away from a
dead server** — and agent traffic (system prompt plus tool schemas) clears 8k routinely.

salekseev:
Follow-up to remove the attribution caveat in my comment above: I re-ran with **only this
patch active** — `--kv-cache-dtype` left at `auto` (bf16), `--kv-reserve-tokens 65536`
(pool: 65560 tokens / 5.43 GiB), `--max-prefill-length` at its 8192 default. That is
byte-for-byte the configuration where 35415 prompt tokens killed the worker before.

| prompt tokens | pre-patch | post-patch |
|---|---|---|
| 32215 | OK | OK |
| 35415 | worker died, server shut down | **OK, 819 tok/s** |
| 51543 | untestable (server already gone) | **OK, 2282 tok/s** |

Engine `{"status":"ok"}` afterwards. So the fix stands on its own here — #103 and the
131072-token pool in my earlier numbers were incidental.

salekseev:
Heads-up that #211 fixes this same bug with the opposite approach — keeping the routed-row
count on `z` and chunking it, on the argument that `x` scatters L2 locality for this
bandwidth-bound GEMV. I have been running this PR in production on sm_89 for a day, so I
measured the two head to head.

**#211 is faster on prefill here, consistently.** Same image, same flags, same model, only the
kernel patch swapped; separate JIT cache directories; JIT warmed before measuring.

| prompt tokens | this PR (grid.x) | #211 (z chunked) | Δ |
|---|---|---|---|
| 35415 | 770.7 tok/s | 808.6 | +4.9% |
| 51543 | 1953.1 | 2019.1 | +3.4% |
| 77303 | 1567.9 | 1627.1 | +3.8% |
| 103063 | 1787.0 | 1844.8 | +3.2% |

Decode is identical (149.11/145.56 vs 148.79/145.34 at 400/1200 tokens) — expected, since a
decode step routes only `batch * top_k` rows, so the axis hardly matters there. The gap appears
exactly where the locality argument predicts.

None of which detracts from the diagnosis in this PR, which is what let me fix a dead
production stack: correct root cause, and the write-up of the latched-error misattribution
saved me from chasing flashinfer's activation kernel any further than I already had.

**The part of this PR I would keep regardless of which axis wins:**
`FT_MOE_VEC_LAUNCH_CHECK()` after each of the 19 launches. #211 does not add it, and without it
a rejected launch configuration only sets the error flag and is reported by whatever unrelated
CUDA call runs next — which is precisely why this bug was so hard to place. That check is
orthogonal to the axis choice and worth landing on its own.

Setup: FreeToken 0.1.2, RTX 4080 SUPER (sm_89, 16 GB), driver 610.57.04 / CUDA 13.3,
`unsloth/gemma-4-26B-A4B-it-qat-GGUF` (Q4_0, `top_k = 8`),
`--kv-reserve-tokens 131072 --kv-cache-dtype q8_0 --max-running-requests 4`, default 8192
chunked prefill. Full numbers also on #211.

Issue #111, opened 2026-08-23
Requests longer than the KV pool are queued forever with no error — and --moe-cache-auto leaves only ~8k tokens of KV on a 96GB GPU


CygnusBill:
## Summary

Two related problems, one report:

1. A request whose prompt needs more KV pages than `num_pages` is **never scheduled and never rejected**: no prefill batch, no log line, no HTTP error — the client waits until its own timeout, and the engine (which is perfectly healthy) simply serves the next request after the abort.
2. With `--moe-cache-auto` on a 96GB RTX PRO 6000 serving MiniMax-M2.5-NVFP4, the joint budget solve yields `moe_cache_size=10114 num_pages=8209` (page_size=1) — an **effective context of ~8.2k tokens** for a model advertising far more, with nothing in the logs flagging it. The pool lands just above the `kv_reserve_tokens=8192` floor because the solve gives nearly all remaining budget to MoE cache slots.

Together: any agent-style workload (system prompt + tools + history routinely >8k tokens) silently hangs at defaults on high-end hardware.

## Environment

FreeToken v0.1.2 @ f0abe58, torch 2.11.0+cu130, CUDA 13.0.1, driver 595.91.07, RTX PRO 6000 Blackwell 96GB, EPYC 128c/503GB, `nvidia/MiniMax-M2.5-NVFP4`.

## Reproduction

1. `ft serve --model nvidia/MiniMax-M2.5-NVFP4 --moe-cache-auto` → log: `resolved moe_cache_size=10114 num_pages=8209 (prefill_overlap=True)`.
2. Send prompts up to ~7.1k tokens → fine (single prefill batch, ~1.5k tok/s ingest).
3. Send a prompt of ~8.8k tokens (> 8209 pages) → the request never returns. Server log shows **no prefill batch for it at all**; GPU idle; no error anywhere.
4. Abort the client; send any normal request → served immediately (the engine was healthy the whole time — this cleanly discriminates it from the engine-death issue at https://github.com/FlashML-org/FreeToken/issues/110).

Verified allocator-independent (identical under `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`) and MoE-overlap-independent (identical with `--disable-moe-prefill-overlap`). The pass/hang boundary tracks token count against `num_pages` exactly: a prompt tokenizing to 8,011 passes; 8,760 hangs. Full experiment trail in the first comment.

## Suggested fixes

[...]

PR #198, opened 2026-08-26
fix(engine): reserve explicit KV pages during MoE auto-sizing

pedro-moser:
## Summary

- include an explicit `--num-pages` KV allocation in the reserve passed to the joint MoE/KV auto-sizing solver
- also covers `--num-tokens`, which resolves to `num_page_override` after page size is finalized
- preserve the existing `kv_reserve_tokens` and model-specific minimum-reserve floors by taking the maximum of all three

## Problem

`Engine._resolve_auto_moe_cache_size()` runs before the KV pool is allocated. It previously passed only:

```python
max(config.kv_reserve_tokens, min_reserve)
```

When a user explicitly requested more KV pages/tokens, that allocation did not participate in the MoE cache budget. The solver could therefore assign those bytes to expert slots and leave startup to fail later during KV allocation.

This is the explicit-context half of the tradeoff described in #111. It does not change the default 8,192-token reserve or admission behavior for over-capacity requests.

## Test plan

Regression test on current `main`:

- configure 64 pages with a resolved page size of 16
- intercept the arguments passed to `resolve_moe_cache_auto`
- verify `kv_reserve_tokens == 64 * 16`

The test failed before the fix with `0 != 1024` and passes after it.

Executed:

```text
PYTHONPATH=$PWD/python pytest -q tests/engine/test_cache_budget.py
24 passed in 1.50s
```

`git diff --check` and Python compile checks also pass.

## Risk and exclusions

- No behavior changes when neither `--num-pages` nor `--num-tokens` is set.
- An explicit context request can now make the budget solver reject startup earlier instead of allowing a later CUDA OOM; that is intentional.
- This PR does not change the default MoE/KV split, scheduler admission, GGUF cache layout, or parser behavior.

Addresses part of #111.

### DIFF
--- a/python/freetoken/engine/engine.py
+++ b/python/freetoken/engine/engine.py
@@ -490,7 +490,11 @@ class Engine:
             num_experts=num_experts,
             total_experts=total_experts,
             prefill_overlap=config.moe_prefill_overlap,
-            kv_reserve_tokens=max(config.kv_reserve_tokens, min_reserve),
+            kv_reserve_tokens=max(
+                config.kv_reserve_tokens,
+                min_reserve,
+                (config.num_page_override or 0) * page_tokens,
+            ),
             page_size=page_tokens,
             quant_format=banks.quant_format,
         )
--- a/tests/engine/test_cache_budget.py
+++ b/tests/engine/test_cache_budget.py
@@ -265,7 +265,7 @@ def test_mha_kv_cost_simple_full_attention():
     assert fixed == 0
 
 
-def test_engine_resolve_auto_moe_cache_size_maps_kwargs():
+def test_engine_resolve_auto_moe_cache_size_maps_kwargs(monkeypatch):
     import torch
 
     from freetoken.engine.engine import Engine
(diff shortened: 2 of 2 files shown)


gdevenyi:
Tested on 2 x RTX 6000 Ada (48 GiB, sm_89) serving RadixArk/Qwen3.8-Flash-Next-NVFP4 at TP=2 (#385) with `--moe-backend offload --ple-backend pinned --num-tokens 262144 --memory-ratio 0.94 --max-running-requests 16` and the implicit `--moe-cache-auto`. On `main` this flag combination OOMs at init because the auto-sized expert cache ignores `--num-tokens` (#383).

With this PR's engine hunk applied alone (my local equivalent clamp reverted; the test hunk not applied because it collides with my own regression test), the KV pool comes up at 262,144 tokens and the expert cache plans around it:

| | single-stream tok/s | 8 concurrent tok/s | expert residency | TTFT (1k prompt) |
|---|---:|---:|---:|---:|
| main + local clamp | 90.8 | 326.0 | 94.5% | 0.81 s |
| main + #198 | 90.4 | 322.2 | 94.5% | 0.85 s |

Same behaviour and same numbers within noise, so this fixes #383 for that path.

PR #208, opened 2026-08-26
qwen3_5_moe: support unsloth/Qwen3.8-27B-NVFP4 (mixed-precision NVFP4+FP8)

chrisqianz:
## Problem

Mixed-precision compressed-tensors exports (e.g. `unsloth/Qwen3.8-27B-NVFP4`) cannot load: the dense pass assumes either pure NVFP4 (`weight_packed`) or pure bf16, but unsloth's layout stores attention linears as weight-only FP8 (per-row scale), most dense-MLP layers as NVFP4, a few MLP layers as FP8, and `lm_head` as FP8. Raw e4m3 tensors die in the bf16 fusion (`torch.cat` fp8/bf16 promotion), and bf16 dequant of `lm_head`/MLP doubles memory.

## Fix (per-module native storage — no new kernels)

Reuses the existing `Fp8PerTensor*` kernels; no shared kernel / modelopt-pass changes:

- `models/config.py`: `dense_mlp_storage` per-layer override on `ModelConfig`
- `models/loader.py`: `ShardReader.has()` for scale-sibling detection
- `qwen3_5_moe/config.py`: `_compressed_linear_storage` sniffs the safetensors index `weight_map` (order-independent per-layer classification: `weight_packed` → nvfp4, scaled → fp8, plain → bf16) + FP8 `lm_head` detection
- `qwen3_5_moe/moe.py` + `model.py`: overridden layers build shared-expert / dense-MLP / `lm_head` linears as native W8A16
- `qwen3_5_moe/weight.py`: dense pass keeps scaled FP8 linears native (fused q/k/v → `qkv_proj`, in_proj_qkv/z → `in_proj_qkvz`, gate/up → `gate_up_proj`, singletons o_proj / GDN out_proj / down_proj / lm_head with per-row fp32 scales); unscaled fp8 dequantizes to bf16 (per-tensor / per-row / block scale). Official pure-NVFP4 checkpoints are unaffected (pass is a no-op there).
- `docs/models.md`: add `unsloth/Qwen3.8-27B-NVFP4` to the known-good list

## Verification (RTX 5090 D 32GB, driver 595.84, torch 2.11 cu130)

- Before: `torch.cat` fp8/bf16 promotion crash on the mixed layout
- After: native resident **21.809 GiB** (vs ~54 GB if all dequantized to bf16)
- config tests 8/8, weight tests 14/14
- full-pass key/shape/dtype vs model state dict: 0 mismatch / 0 stray / 0 missing
- `ft serve` e2e: 21.8 GB load, KV 16K, CUDA graph capture, `/health` ok, real chat generation

Usage note: on a 32 GB card use `ft serve --model unsloth/Qwen3.8-27B-NVFP4 --num-tokens 32768 --max-prefill-length 1024` (the large-vocab logits buffer OOMs at the default 8192-token prefill).

### DIFF
--- a/docs/models.md
+++ b/docs/models.md
@@ -11,7 +11,7 @@ for them; other checkpoints of the same architectures work too.
 | GLM-4.7 | [nvidia/GLM-4.7-NVFP4](https://huggingface.co/nvidia/GLM-4.7-NVFP4) |
 | Qwen3.8-Flash-Next | [Qwen/Qwen3.8-Flash-Next-FP8](https://huggingface.co/Qwen/Qwen3.8-Flash-Next-FP8), [RadixArk/Qwen3.8-Flash-Next-NVFP4](https://huggingface.co/RadixArk/Qwen3.8-Flash-Next-NVFP4) |
 | Qwen3.6 / Qwen3.5 MoE | [Qwen/Qwen3.6-35B-A3B](https://huggingface.co/Qwen/Qwen3.6-35B-A3B) ([-FP8](https://huggingface.co/Qwen/Qwen3.6-35B-A3B-FP8)), [nvidia/Qwen3.6-35B-A3B-NVFP4](https://huggingface.co/nvidia/Qwen3.6-35B-A3B-NVFP4), [Qwen/Qwen3.5-35B-A3B](https://huggingface.co/Qwen/Qwen3.5-35B-A3B) ([-FP8](https://huggingface.co/Qwen/Qwen3.5-35B-A3B-FP8)) |
-| Qwen3.8 / Qwen3.6 dense | [Qwen/Qwen3.8-27B](https://huggingface.co/Qwen/Qwen3.8-27B) ([-FP8](https://huggingface.co/Qwen/Qwen3.8-27B-FP8)), [RadixArk/Qwen3.8-27B-NVFP4](https://huggingface.co/RadixArk/Qwen3.8-27B-NVFP4), [Qwen/Qwen3.6-27B](https://huggingface.co/Qwen/Qwen3.6-27B) ([-FP8](https://huggingface.co/Qwen/Qwen3.6-27B-FP8)), [nvidia/Qwen3.6-27B-NVFP4](https://huggingface.co/nvidia/Qwen3.6-27B-NVFP4) |
+| Qwen3.8 / Qwen3.6 dense | [Qwen/Qwen3.8-27B](https://huggingface.co/Qwen/Qwen3.8-27B) ([-FP8](https://huggingface.co/Qwen/Qwen3.8-27B-FP8)), [RadixArk/Qwen3.8-27B-NVFP4](https://huggingface.co/RadixArk/Qwen3.8-27B-NVFP4), [unsloth/Qwen3.8-27B-NVFP4](https://huggingface.co/unsloth/Qwen3.8-27B-NVFP4) (mixed NVFP4+FP8 export), [Qwen/Qwen3.6-27B](https://huggingface.co/Qwen/Qwen3.6-27B) ([-FP8](https://huggingface.co/Qwen/Qwen3.6-27B-FP8)), [nvidia/Qwen3.6-27B-NVFP4](https://huggingface.co/nvidia/Qwen3.6-27B-NVFP4) |
 | Qwen3-MoE | [Qwen/Qwen3-30B-A3B](https://huggingface.co/Qwen/Qwen3-30B-A3B) |
 | gpt-oss | [openai/gpt-oss-120b](https://huggingface.co/openai/gpt-oss-120b), [openai/gpt-oss-20b](https://huggingface.co/openai/gpt-oss-20b) |
 | Gemma-4 | [google/gemma-4-26B-A4B-it](https://huggingface.co/google/gemma-4-26B-A4B-it), [nvidia/Gemma-4-26B-A4B-NVFP4](https://huggingface.co/nvidia/Gemma-4-26B-A4B-NVFP4), [google/gemma-4-12B-it](https://huggingface.co/google/gemma-4-12B-it), [nvidia/Gemma-4-31B-IT-NVFP4](https://huggingface.co/nvidia/Gemma-4-31B-IT-NVFP4) .. |
@@ -40,3 +40,6 @@ for them; other checkpoints of the same architectures work too.
   authoritative model args are read from there.
 - Qwen3.8-Flash-Next keeps a 47.7 GiB PLE n-gram table pinned in host RAM.
 - Multimodal checkpoints are served text-only.
+- unsloth's dense Qwen3.x NVFP4 exports are mixed-precision per module (NVFP4 MLP
+  layers, FP8 attention/GDN/lm_head, bf16 residual parts); each linear loads in the
+  storage the checkpoint actually uses.
--- a/python/freetoken/models/config.py
+++ b/python/freetoken/models/config.py
@@ -267,6 +267,11 @@ class ModelConfig:
     # it bf16. Separate from dense_quant because only some NVFP4 checkpoints quantize lm_head
     # (modelopt MIXED_PRECISION does; pure NVFP4 leaves it bf16).
     lm_head_quant: str = "none"
+    # Per-layer override of the dense-MLP storage for mixed compressed-tensors exports
+    # (unsloth: most layers packed NVFP4, a few FP8 layers). Keys are layer ids; the value
+    # replaces dense_quant for that layer's dense MLP at construction ("nvfp4" native W4A16
+    # / "bf16" dequant-at-load). None = every dense-MLP layer follows dense_quant.
+    dense_mlp_storage: dict[int, str] | None = None
     shared_expert_intermediate_size: int = 0
     use_qk_norm: bool = False
     # ----- DeepSeek/GLM-style MoE extensions (default keeps other models intact) -----
--- a/python/freetoken/models/loader.py
+++ b/python/freetoken/models/loader.py
@@ -182,6 +182,9 @@ class ShardReader:
     def names_in(self, file: str) -> list[str]:
         return [name for name, shard in self._map.items() if shard == file]
 
+    def has(self, name: str) -> bool:
+        return name in self._map
+
     def get_tensor(self, name: str) -> torch.Tensor:
         import safetensors
 
--- a/python/freetoken/models/qwen3_5_moe/config.py
+++ b/python/freetoken/models/qwen3_5_moe/config.py
@@ -68,6 +68,77 @@ def _expert_quant(hf_config: Any) -> str:
 _compressed_tensors_nvfp4 = detect_compressed_tensors_nvfp4
 
 
+def _compressed_linear_storage(
+    hf_config: Any,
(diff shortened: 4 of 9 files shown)


yuyi2439:
maybe fix #238

dcutugno:
Second data point for this PR, on a different variant of the same family.

**Checkpoint:** a third-party Qwen3.8-27B **multimodal** NVFP4 export — `architectures: ["Qwen3_5ForConditionalGeneration"]`, `model_type: qwen3_5`, compressed-tensors 0.18.0, `quantization_config.format == "mixed-precision"`. Same layout family you describe, but the vision-tower variant rather than the text-only one.

**Env:** freetoken `0.1.2+g932e87e8a`, Python 3.12, Windows 11, sm_120.

Both failure modes you describe reproduce:

1. Stock: the fp8/bf16 promotion crash in the dense pass (tracked as #238).
2. Working around it by dequantizing the FP8 group to bf16 at load — the approach in #275 — clears that crash and converts cleanly, but then dies in model construction:

```
File "freetoken/kernel/triton/nvfp4_linear.py", line 867, in load_state_dict
  s = state_dict.pop(_concat_prefix(prefix, "weight_scale"))
KeyError: 'model.layers.0.linear_attn.out_proj.weight_scale'
```

because `detect_compressed_tensors_nvfp4()` is true on any NVFP4 group and `qwen3_5_moe/config.py` then forces `attn_quant = "nvfp4"`. That corroborates the per-module classification this PR introduces — no single global scalar describes the checkpoint.

**Exact per-layer split**, since the description says "most dense-MLP layers as NVFP4, a few MLP layers as FP8":

| modules | group | quant |
|---|---|---|
| MLP `gate/up/down_proj`, layers 0-55 | group_1 (168 targets) | `nvfp4-pack-quantized` |
| MLP `gate/up/down_proj`, layers 56-63 | group_0 | `float-quantized` FP8 |
| `linear_attn.{in_proj_qkv,in_proj_z,out_proj}`, 48 layers | group_0 | FP8 |
| `self_attn.{q,k,v,o}_proj`, 16 layers | group_0 | FP8 |

232 FP8 targets / 168 NVFP4. FP8 `weight_scale` is per-row BF16 `[out, 1]`; NVFP4 carries fp8 `weight_scale` plus fp32 `weight_global_scale` / `input_global_scale`. `transform_config` and `sparsity_config` are both empty.

**One variant difference worth handling:** here `lm_head` is *not* FP8. It's untied (`tie_word_embeddings: false`) and sits in the 304-entry `ignore` list, so it stays bf16. If the FP8 `lm_head` detection assumes presence, this is the checkpoint where it's absent — the "plain -> bf16" branch should cover it, but it's untested against this layout.

On footprint: dequantizing the 232 FP8 tensors takes them from 8.71 GiB to 17.42 GiB, giving a 30.06 GiB FTW from a 23 GiB source. Consistent with your 21.8 vs ~54 GB figures.

I can't run this one — 16 GB card, and 21.8 GiB resident won't fit — so I can't verify generation. Happy to test conversion and load-path behaviour on this variant if useful.

dcutugno:
Follow-up to my comment above — I applied this PR's sources onto the released build (dropping the branch `.py` files in place of the Cython-compiled modules) and tested it on the multimodal variant. Mostly good news, plus one gap.

**Working:**

- This PR's own suites: **22/22** (8 config + 14 weight).
- `ft checkpoint`: **21.35 GiB** FTW, 1027 weight tensors, 32.4 s — against 30.06 GiB when the FP8 group is dequantized to bf16 instead. Matches your 21.8 vs ~54 GB result.
- `ft serve` pointed **directly at the HF source dir**: loads and generates coherent text. So the per-module classification and the native FP8 path both work on this variant.

**The gap: the `ft checkpoint` -> `ft serve <ftw-dir>` round-trip fails.**

```
KeyError: 'model.layers.0.linear_attn.in_proj.weight'
```

`_compressed_linear_storage` classifies layers by sniffing `model.safetensors.index.json`. A converted FTW directory carries `freetoken_weight.json` instead and has no safetensors index, so the classification silently falls back:

| config read from | `attn_quant` | `dense_mlp_storage` |
|---|---|---|
| HF source dir (index present) | `fp8_pertensor` | populated |
| converted FTW dir (no index) | `nvfp4` | `None` |

With `attn_quant` back to `nvfp4`, `gdn.py` takes the `else` branch and builds the fused bf16 `in_proj`, while the FTW was written with `in_proj_qkvz` + per-row fp32 scales. Hence the missing key.

`test_index_unavailable_falls_back_to_native` covers the fallback itself, but in the FTW case the fallback is silently *wrong* rather than safe: the storage decision was already made at conversion time and then lost. Persisting the resolved `dense_mlp_storage` / attention storage into the FTW metadata (or re-deriving it from the FTW's own tensor names, where `in_proj_qkvz` vs `in_proj` is already unambiguous) would close it.

**A concrete fix, tested.** The FTW already carries what is needed: `freetoken_weight.json` records a dtype per tensor, and those dtypes describe the storage unambiguously.

```
model.layers.0.mlp.gate_up_proj.weight           uint8           -> nvfp4
model.layers.0.mlp.gate_up_proj.weight_scale     float8_e4m3fn
model.layers.56.mlp.gate_up_proj.weight          float8_e4m3fn   -> fp8
model.layers.56.mlp.gate_up_proj.weight_scale    float32
model.layers.0.linear_attn.in_proj_qkvz.weight   float8_e4m3fn   -> attention fp8
lm_head.weight                                   bfloat16        -> lm_head bf16
```

So `_compressed_linear_storage` can fall back to `freetoken_weight.json` when `model.safetensors.index.json` is absent, applying the same per-layer rule against dtypes rather than key suffixes: `uint8` -> nvfp4, `float8*` -> fp8, anything else -> bf16; the presence of `linear_attn.in_proj_qkvz` at all implies fp8 attention. It returns the same 4-tuple, so nothing downstream changes.

I ran exactly that as a monkeypatch over this branch, and a converted FTW then loads and generates correctly with no other changes.

The alternative -- persisting the resolved `(attn, dense fallback, per-layer overrides, lm_head)` tuple into the FTW metadata at conversion -- needs fewer heuristics, but it requires a metadata field and leaves already-converted FTWs broken, whereas the dtype fallback repairs those too.

Not a blocker for `ft serve` straight off the HF checkpoint, which is how the PR description verifies it — but anyone converting first hits it.

Environment note in case it is useful: this ran on a 16 GB card, which is under the ~21.4 GiB resident. It works on Windows only because WDDM spills into shared system memory, at roughly 1.3 tok/s, with `--max-running-requests 1 --cache-type naive --num-tokens 4096 --max-prefill-length 512` and `FREETOKEN_MAMBA_SSM_DTYPE=bfloat16` (the fp32 GDN recurrent-state pool wants 3.52 GiB on its own at the defaults). Not a configuration I would suggest supporting — just context for the numbers above.

chrisqianz:
Thanks a lot @dcutugno — superb catch, and the analysis checks out end to end against the source. Confirmed: `_compressed_linear_storage` sniffs `model.safetensors.index.json`, and an FTW dir (self-contained by design: `config.json` copied, `freetoken_weight.json` replaces the safetensors index) always takes the fallback branch, silently reverting `attn_quant` to `nvfp4` after the storage decision was already fixed at conversion time — hence the missing fused `in_proj` key. Our fallback test only covered the hub-id / missing-index cases, which are genuinely safe; the FTW case is silently-wrong and you found exactly that hole.

Going with your option (a): when the safetensors index is absent, re-derive the same 4-tuple from `freetoken_weight.json` tensor names + dtypes (`uint8` -> nvfp4, `float8*` -> fp8, else bf16; `linear_attn.in_proj_qkvz` presence -> fp8 attention). No downstream changes, and it repairs already-converted FTWs. Regression tests for the FTW fallback included. Fix incoming shortly.

Also noted the datapoints — released-build application, multimodal variant, 21.35 GiB FTW vs 30.06 GiB dequantized, and the WDDM-spill env context. Genuinely useful.

chrisqianz:
Fix pushed in `e1d374b` — `_compressed_linear_storage` now falls back to `_ftw_linear_storage` when `model.safetensors.index.json` is absent: re-derives the same 4-tuple from `freetoken_weight.json` per-tensor dtypes/names (`uint8` -> nvfp4, `float8*` -> fp8 restricted to `.weight`-suffixed names so scale siblings can't false-positive, split `in_proj_qkvz` registration -> fp8 attention), exact per-layer rule as you outlined. Already-converted FTWs are repaired by this too.

Seven regression tests added, including the pure-NVFP4-FTW case pinned to the *exact* previous default tuple (regression guard for layouts that loaded through the old silent fallback), three-tier dense storage in one checkpoint (nvfp4/fp8/bf16), fp8 lm_head, routed-MoE naming exclusion, and a broken-index safety net. Hub-id / missing-dir fallback untouched.

Your WDDM-spill datapoints are noted. Happy to have a round-trip test on your side against the multimodal variant when convenient — otherwise this is ready for another review pass.

chrisqianz:
Heads-up: the lm_head full-window OOM fix that briefly lived on this branch has moved to its own PR — **#342**. It is a pre-existing engine bug (present since the initial release, `forward()` untouched by this PR) and affects any qwen3_5_moe dense model at the 248,320 vocab on 32 GB cards, not just the unsloth export. This branch now contains only the mixed-precision loading work: `570c437` loader/config, `d9d05b0` docs, `71bf149` tidy, `e1d374b` FTW metadata fallback.

chrisqianz:
Closing as superseded by #418 + #438, which absorb this PR's whole problem space with a better shape: every module's storage comes from `QuantConfig.scheme_for(prefix)` instead of per-family string detection, so the mixed-precision exports this PR exists for need no bespoke path.

Mapping of this PR's mechanisms onto current `main`:

- `_compressed_linear_storage` (safetensors-index sniffing) → the string detectors are gone in #438; one `_DenseReader` fills exactly the buffers each module's scheme declares (per-tensor scales broadcast per row).
- quantized head swaps (`Fp8PerTensorLinear` / `Nvfp4LMHead` as `lm_head`) → `ParallelLMHead` builds through `quant_method`; the prefill `get_last_indices` reduction survives for every quantized head, which also retires #342.
- `e1d374b` (FTW dtype fallback) → structural: storage is decided from `quantization_config`, which lives in `config.json` and therefore in the FTW dir too; FTW dense tensors replay model-agnostically, so the conversion-time decision can no longer be lost at load.

#438's e2e matrix passes `unsloth/Qwen3.8-27B-NVFP4` (AIME, this PR's target checkpoint) and the fp8-lm_head variants, on both compressed-tensors and ModelOpt layouts — including the `ct_set` exact-name fix for `ignore` lists that enumerate containers, which this PR would otherwise have needed too.

@dcutugno your analysis held up completely — the FTW silent-fallback hole was real, the dtype-fallback was a correct fix at the time, and the per-layer table you posted is what I verified `e1d374b` against. It just lost to the structural fix: once storage comes from the quantization config, there is no index-sniffing fallback left to fall through. One note for anyone coming here from your multimodal setup: FTWs converted by older builds lack the `input_scale` tensors the model now declares — `scripts/ftw_hotfix.py --ftw <dir> --source <hf dir>` repairs them.

chrisqianz:
Superseded by #418/#438 — see comment above.

PR #223, opened 2026-08-26
fix(server): validate sampling params and rebuild timeout at the API layer

Artemowka22:
## Summary

1. **Invalid sampling silently reaches the engine.** `resolve_sampling` forwards arbitrary client floats: `temperature=-3`, `NaN`, `inf`, `top_p=0`, `top_p=7`, `top_k=0` all become a `SamplingParams` and die (or misbehave) deep in the sampler instead of failing the request. Fix: after checkpoint defaults are filled, validate `temperature` (finite, ≥ 0), `top_p` ((0, 1]), `top_k` (−1 or ≥ 1) and raise `ValueError` in the same style as the existing `max_tokens` check — every adapter already maps that to a clean HTTP 400.

2. **`CacheRebuildRequest.timeout` is unbounded.** A negative/NaN/huge timeout reaches `asyncio.wait_for` on the rebuild future; `inf` parks the handler forever. Fix: pydantic bound `0 < timeout ≤ 3600`.

The audit that surfaced these (agisota/freetoken-mlx, docs/AUDIT.md, defects 3–4) also reported a TOCTOU race on concurrent `/v1/cache/rebuild`; current upstream has already restructured that path (single synchronous section from the pre-flight checks to `maintenance_state = "rebuilding"`), so only the timeout bound is ported.

## Test plan

- [x] parametrized invalid/boundary sampling matrix on `/v1/chat/completions` in `tests/server/test_openai_api.py` — 400 with the offending parameter named; boundary values `temperature=0`, `top_p=1`, `top_k=-1`, `top_k=1` accepted
- [x] rebuild timeout bounds in `tests/server/test_rebuild_maintenance.py`
- [x] full `tests/server/` suite passes

### DIFF
--- a/python/freetoken/server/api_server.py
+++ b/python/freetoken/server/api_server.py
@@ -33,7 +33,7 @@ from freetoken.utils import (
     init_logger,
     load_generation_sampling,
 )
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 from .args import ServerArgs
 from .anthropic_api import register_anthropic_routes
@@ -494,7 +494,7 @@ class CacheRebuildRequest(BaseModel):
     # is deferred (needs the drain-gate machinery); constraining the Literal makes an
     # unsupported value fail fast with a 422 at the API layer instead of a generic 503.
     mode: Literal["if_idle"] = "if_idle"
-    timeout: float = 300.0
+    timeout: float = Field(default=300.0, gt=0, le=3600)
 
 
 async def dispatch_rebuild(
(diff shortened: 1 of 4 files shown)


gdevenyi:

Probes against the live server: `temperature=-1`, `top_p=0`, `top_k=0` each answer 400 with the message from this PR (`temperature must be a finite number >= 0, got -1.0` and so on). tests/server passes. Merged clean, in production.

Artemowka22:
Thank you for the production validation — live probes against a deployed server (and the exact 400 messages coming back as written) is better coverage than anything we could add here. Nothing to change on our side; noting for the record that this PR has now been run in production on sm_89 with TP=2 and the offload backend.

Issue #220, opened 2026-08-26
RTX 4070 Ti Super (SM89): every request hangs ~123s then worker dies (empty response) — Qwen3.6-35B-A3B-NVFP4 offload


dalu951:
## UPDATE 4: Request-#2 EOF — status after CUB/JIT fix (correlation, not yet proven root cause)

The previously observed 9 ms EOF is reproducible only under the earlier configuration and has not reproduced under the current constrained configuration. Historical evidence rules out TCP connection reuse as the cause (the failing run used fresh connections; 3/3 requests on one keep-alive connection succeed under the current config).

Historical runs for reference:
- 22:48 run (SmolLM2, --num-tokens 4096 only, cuda-graph-max-bs default, attention auto→fi): req 1 = 200, req 2 = empty/9 ms on a fresh TCP connection; server process stayed alive (no worker exit visible).
- 23:08 run (same model/flags): req 1 = 200, req 2 = 200, backend worker exited ~9 s later.
- 00:17 run (35B, --attn triton): first request triggered the flashinfer JIT compile of sampling.cu → 100 CUB errors → worker died mid-generation.

Current run (35B, --num-tokens 8192, --cuda-graph-max-bs 1, --attn triton, after CUB fix): 15/15 requests 200 (2 fresh connections, 3 on one keep-alive connection, 10 sequential), full dispatch trace for every request, zero errors, no worker exit.

The remaining suspect is CUDA-graph / request-state behavior under the default batch-size (cuda-graph-max-bs=auto) configuration — but this is NOT yet proven. A controlled A/B test (only cuda-graph-max-bs varied, everything else identical) is planned; results will be posted here.

## UPDATE 5: A/B test result — cuda-graph-max-bs is NOT the trigger

Controlled A/B experiment on the Qwen3.6-35B-A3B (everything identical except one flag):
- A: --cuda-graph-max-bs 1 -> 5/5 requests 200 (fresh TCP connection per request)
- B: cuda-graph-max-bs default/auto -> 5/5 requests 200 (fresh TCP connection per request)

Both runs: full dispatch trace for every request, zero errors, no worker exit.

The default batch-size (CUDA-graph) configuration alone does not reproduce the 9 ms EOF.
The earlier EOF remains tied to the pre-fix first-request JIT/rebuild instability
(flashinfer JIT compile of sampling.cu failing with the bundled CUB headers).
If it reappears under any configuration, a trace-based reproducer will be posted here.

[...]

PR #225, opened 2026-08-26
fix(server): resolve in-flight generation requests when the backend worker dies

dalu951:
## Problem

When the backend worker (scheduler) dies **while generation requests are in flight** (e.g. OOM / segfault during prefill on a 16 GB consumer GPU, RTX 4070 Ti SUPER / SM89), the server:

1. never resolves the waiting generation futures (`wait_for_ack` loops hang forever — no timeout),
2. leaves the request slots / ack state behind (state leak),
3. then shuts down after `BACKEND_DEATH_EXIT_GRACE_S` — clients see an **empty response after ~133 s** instead of a clean error.

## Fix

Add `fail_pending_generations()` to `FrontendManager` — the generation-side analogue of the existing `fail_pending_rebuilds()`: on worker death, inject an error `UserReply` (`error_code="backend_died"`) into every pending `event_map` entry via `loop.call_soon_threadsafe`, so all waiting handlers raise `GenerationError` and clean up.

Call it from `_on_failure` right next to `fail_pending_rebuilds`.

## Impact

- Clients get a proper error response instead of an empty/hanging request.
- No leaked ack/event state on worker death.
- Verified: supervisor detects the death, module imports cleanly, patch applies to current HEAD (9ef3651).

## Repro context

Full background incl. the KV-cache size bug (`--num-tokens` workaround) and missing `chat_template` in the unsloth mirror: https://github.com/FlashML-org/FreeToken/issues/220

### DIFF
--- a/python/freetoken/server/api_server.py
+++ b/python/freetoken/server/api_server.py
@@ -288,6 +288,34 @@ class FrontendManager:
             return
         self.maintenance_state = "failed" if msg.status == "failed" else "serving"
 
+
+    def fail_pending_generations(self, message: str) -> None:
+        """Resolve every in-flight generation as failed (worker death): inject an error ack
+        so wait_for_ack yields it and handlers raise GenerationError instead of hanging
+        forever with the request slot leaked."""
+        loop = self._loop
+        if loop is None:
+            return  # listener never started -> no pending generations
+
+        def _resolve_all() -> None:
+            for uid in list(self.event_map):
+                self.ack_map[uid].append(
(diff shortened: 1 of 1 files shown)


dalu951:
Closing for now: after deeper instrumentation the root cause is different (flashinfer JIT sampling-kernel fails to compile with nvcc 12 / CUB 12 on SM89 -> worker dies during first request -> maintenance state 'failed'). Reopening with a proper regression test once the trace is complete.

Issue #227, opened 2026-08-26
[Bug]: RuntimeError: Rowwise scaling is not currently supported on your device


voidhaze:
Running Win 11 with a 4070 Super, was trying to run Qwen3.6-35B-A3B NVFP4; when the model is loading / starting it gets to about 60% and then crashes out with the error:

"RuntimeError: Rowwise scaling is not currently supported on your device"

Logs attached

[crash_log.txt](https://github.com/user-attachments/files/31488909/crash_log.txt)

Buccolieri96:
Same

Thomaswang0822:
Same here, Qwen3.8-27B-NVFP4 on 5090

oro872gioioso:
Same with 5060Ti 16GB. Still no solution?

Usigned:
same, Qwen3.6-35B-A3B-NVFP4 on 4090D

qualitymaterial:
Same on a 5070, and I think this is the cause — it isn't NVFP4- or model-specific.

Rowwise `_scaled_mm` is a Hopper-only CUTLASS kernel, restricted to `sm_90` in [pytorch/pytorch#145728](https://github.com/pytorch/pytorch/pull/145728). But `fp8_pertensor_linear()` dispatches to it on `e4m3_native()` (`cuda_capability_geq(8, 9)`) — that's the fp8 *tensor core* capability, not the *rowwise* capability. So every consumer card fails, sm_89 and sm_120 alike, which matches the 4070S/4090D/5060Ti/5090 reports above. Probed on my 5070 (torch 2.11.0+cu130): tensorwise OK, rowwise raises.

Which weights hit it is decided in `Fp8PerTensorLinear.load_state_dict`:

```python
self._uniform_scale = bool((scale == scale[0]).all().item())
```

A real per-tensor weight is uniform and takes the safe tensorwise path. A **fused** projection is piecewise-constant, so it's never uniform and always takes rowwise — hence GDN `in_proj_qkvz` (`qwen3_5_moe/gdn.py:165`). Only that and `glm_moe_dsa/attention.py` can trigger it. It fires during CUDA graph capture, so it kills the scheduler process instead of failing a request — the daemon stays up and the app reports it as the engine failing to start.

Fix: the W8A16 triton fallback already in that file handles this correctly, so it just needs the fused case routed to it when rowwise is unavailable —

```diff
-    elif input_scale is not None and e4m3_native():
+    elif input_scale is not None and e4m3_native() and (
[...]

PR #243, opened 2026-08-28
fix(kernel): avoid row-wise _scaled_mm stall on sm_89 with torch<2.12

### LINKED ISSUE #182: Bug in PyTorch 2.11 causes SM89 hangs

bernimccoy:
Fixes #182. Same root cause as #72 and the prefill hangs in #220; also covers #227.
Complements #184 (torch 2.13): this keeps the current pin and is a no-op on torch >= 2.12.

## Problem

On sm_89, torch < 2.12 runs row-wise `_scaled_mm` through a CUTLASS kernel whose launch
ignores the current stream (pytorch/pytorch#177651, fixed in 2.12 by pytorch/pytorch@252bb4a;
isolated by @endenis in #182). `fp8_pertensor_linear` takes that path for fused projections
(q/k/v, GDN qkv|z: one scalar per part) from a side stream. At M >= 256 it stalls the GPU:
no `Prefill batch` line, 100 % util, `kill -9` required; decode-sized M completes, so short
prompts work and long ones hang. Windows torch builds have no row-wise kernel at all (#227).

## Fix

Tensor-wise scaling is unaffected. Where row-wise is unsafe, a fused
projection runs one tensor-wise `_scaled_mm` per part over its row slice (`weight[s:e].t()`,
still stride-only) and concatenates. Same W8A8 scheme (rel ~7e-4 to row-wise, accumulation order), one extra launch per part.

- `rowwise_scaled_mm_ok()`: False on sm_89 + torch < 2.12; otherwise a one-off row-wise probe
  on the default stream (safe there), so a build without the kernel fails at load. Warmed in
  `load_state_dict`, never under graph capture.
- `weight_scale_segments()`: the parts' row ranges from the load-time `weight_scale`
  run-lengths, next to the existing `_uniform_scale` check. A genuine per-row scale (> 8
  segments or not 16-row aligned) stays on the W8A16 triton path.
- `FREETOKEN_FP8_ROWWISE_MM=0/1` forces either path.

Unchanged: uniform-scale weights, no `input_scale`, sm_90+/sm_120, sm_89 on torch >= 2.12.

## Tested on

RTX 4070 SUPER 12 GB (sm_89), driver 591.86, WSL2, torch 2.11.0+cu130, triton 3.6.0,
flashinfer 0.6.17, sglang-kernel 0.4.5. `nvidia/Qwen3.6-35B-A3B-NVFP4`, HF safetensors:

```
ft serve --model ~/models/Qwen3.6-35B-A3B-NVFP4 --port 1919
```

| | `main` (= `FREETOKEN_FP8_ROWWISE_MM=1`) | this branch |
|---|---|---|
| 18-token prompt | ok | ok |
| 626-token prompt, `max_tokens=32` | never returns (> 90 s, GPU 100 %, `kill -9`) | 3 tokens, 4.4 s |
| streamed decode, 299 tokens, bs=1 | n/a | 26.4 tok/s |

Sweep of the fused [8192, 512, 512] x 2048 projection on a side stream: row-wise completes at
M <= 128, stalls at M >= 256; the per-part path completes at every M from 1 to 2010.

`tests/kernels/test_fp8_pertensor_linear.py`: side-stream test at the #182 shape, in a
subprocess so a stall fails instead of hanging pytest; fails on `main` (`rc=124`), passes
here. Per-part path vs row-wise on the same fused input at M=1/4/64/300: rel 6.9e-4 to 8.5e-4.

Not tested on sm_90+/sm_120, torch >= 2.12, or Windows (#227 relies on the probe).

## Caveats

Pre-existing on this GPU, unrelated: 3 `test_w8a8_matches_w8a8_reference` cases miss the 1e-2
tolerance (rel 0.0103-0.0107) on `main` and here alike, incl. the untouched uniform case.

Assisted by: [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/python/freetoken/kernel/triton/fp8_pertensor_linear.py
+++ b/python/freetoken/kernel/triton/fp8_pertensor_linear.py
@@ -19,7 +19,9 @@ accurate than the previous ``weight.to(bf16) * scale`` materialization, which it
 
 from __future__ import annotations
 
+import functools
 import os
+import re
 
 import torch
 import triton
@@ -40,6 +42,64 @@ _TL_DTYPE = {torch.bfloat16: tl.bfloat16, torch.float16: tl.float16, torch.float
 # Escape hatch: FREETOKEN_DEBUG_FP8_REF=1 swaps the triton kernels for a pure-torch dequant matmul
 # (numeric reference / A-B debugging). Evaluated once; the kernels are the default.
 _USE_REF = os.environ.get("FREETOKEN_DEBUG_FP8_REF") == "1"
+
+
+# Row-wise _scaled_mm on sm_89 with torch < 2.12 launches its CUTLASS stream-K kernel off the
+# current stream (pytorch/pytorch#177651, fixed by pytorch/pytorch@252bb4a; #182/#72/#220), and
+# some builds (Windows) have no row-wise kernel (#227). Fallback: one tensor-wise GEMM per part.
+def _torch_version() -> tuple[int, int]:
+    m = re.match(r"(\d+)\.(\d+)", torch.__version__)
+    return (int(m.group(1)), int(m.group(2))) if m else (0, 0)
+
+
+@functools.cache
+def rowwise_scaled_mm_ok() -> bool:
+    """Whether row-wise ``torch._scaled_mm`` may be issued from a side stream on this GPU.
+    Decided once per process, at load (never under graph capture). ``FREETOKEN_FP8_ROWWISE_MM=0/1``
+    forces the answer."""
+    forced = os.environ.get("FREETOKEN_FP8_ROWWISE_MM")
+    if forced in ("0", "1"):
+        return forced == "1"
+    if not torch.cuda.is_available():
+        return True
+    from freetoken.gpu_select import assigned_visible_gpu
+
+    idx = assigned_visible_gpu()
+    dev = torch.device("cuda", torch.cuda.current_device() if idx is None else idx)
+    if torch.cuda.get_device_capability(dev) == (8, 9) and _torch_version() < (2, 12):
+        return False
+    # Probe on the default stream (safe even where the launch ignores the current stream); a
+    # build without the row-wise kernel raises here instead of at the first forward.
+    try:
+        with torch.cuda.device(dev), torch.cuda.stream(torch.cuda.default_stream(dev)):
+            a = torch.zeros(16, 32, dtype=FP8, device=dev)
+            b = torch.zeros(32, 32, dtype=FP8, device=dev)
+            torch._scaled_mm(
+                a, b.t(), scale_a=torch.ones(16, 1, device=dev),
+                scale_b=torch.ones(1, 32, device=dev), out_dtype=torch.bfloat16,
+            )
+            torch.cuda.synchronize(dev)
+    except RuntimeError:
+        return False
+    return True
+
+
+def weight_scale_segments(weight_scale: torch.Tensor) -> list[tuple[int, int]]:
+    """``[start, end)`` row ranges over which ``weight_scale`` is constant (the fused parts).
+    Syncs; call at load."""
+    s = weight_scale.detach().reshape(-1).float().cpu()
+    change = (torch.nonzero(s[1:] != s[:-1]).flatten() + 1).tolist()
+    bounds = [0, *change, s.numel()]
+    return list(zip(bounds[:-1], bounds[1:]))
+
+
+_MAX_SEGMENTS = 8  # q/k/v = 3, GDN qkv|z = 2; a genuine per-row scale stays W8A16 instead
+
+
+def _segments_w8a8_ok(segments: list[tuple[int, int]]) -> bool:
+    """cuBLASLt needs 16-row aligned fp8 operands; more parts than a fused projection has
+    means a genuine per-row scale."""
+    return 0 < len(segments) <= _MAX_SEGMENTS and all((e - s) % 16 == 0 for s, e in segments)
 
 
 # ======================================================================================
@@ -222,6 +282,7 @@ def _static_quant(a: torch.Tensor, input_scale: torch.Tensor) -> torch.Tensor:
 def _scaled_mm(
     a: torch.Tensor, weight: torch.Tensor, weight_scale: torch.Tensor,
     input_scale: torch.Tensor, uniform_scale: bool, out_dtype: torch.dtype,
+    scale_segments: list[tuple[int, int]] | None = None,
 ) -> torch.Tensor:
     """``a @ (weight_fp8 * weight_scale)^T`` as a W8A8 cuBLASLt GEMM.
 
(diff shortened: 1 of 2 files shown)


jason-fxz:
Verified on RTX 4060 Laptop (sm_89, torch 2.11.0+cu130). The fix itself LGTM. Three requests before merge:

**Comments**

Keep them short, and only where the code isn't self-explanatory — the header block can be 3 lines. Three statements need fixing:

- Header: "every other FP8-capable arch goes to cuBLASLt" is wrong. In torch 2.11 row-wise goes to CUTLASS on Ada and Blackwell; only Hopper (cuBLAS ≥ 12.9) uses cuBLASLt. The sm_89 gate is still right — only the sm89 kernel uses stream-K, whose `initialize` memsets a workspace on the wrong stream.
- `_scaled_mm` docstring: "exact, same W8A8 numerics" → same scheme, not bit-identical (accumulation order differs; ~7% of elements, rel ~7e-4).
- `rowwise_scaled_mm_ok` docstring: "Probe on the default stream" — the probe runs on the current stream. Drop the claim or actually switch to the default stream.

**Tests**

One test is enough: run the same fused input through the row-wise and per-part paths (monkeypatch `rowwise_scaled_mm_ok`) and compare them directly at a few M, rel < 2e-3 (measured ~7e-4). This replaces the three new tests that go through the W8A8 reference or check internal plumbing.

**Commit message**

Use `Assisted-by: <assistant name>` instead of `Co-Authored-By:`.

bernimccoy:
Thanks for the review and the 4060 run. All three addressed in the amended commit:

- Comments: header cut to 3 lines and the cuBLASLt claim dropped (now names the stream-K launch); `_scaled_mm` docstring says same scheme, not bit-identical; the probe now actually runs on the default stream (`torch.cuda.stream(torch.cuda.default_stream(dev))`) and the docstring matches.
- Tests: the three reference/plumbing tests replaced by `test_per_part_path_matches_rowwise`, which monkeypatches `rowwise_scaled_mm_ok` both ways on the same fused input at M=1/4/64/300 and asserts rel < 2e-3 (measured 6.9e-4 to 8.5e-4 on the 4070 SUPER). The side-stream subprocess test stays.
- Commit trailer: `Assisted-by:`.

PR body updated to match.

jason-fxz:
Nice work, thanks. Merging.

PR #254, opened 2026-08-28
feat(bench): pass ft serve options through '--' in bench_decode_moe

yuyi2439:
Replace the script-side mirrors of server flags (--cache, --cache-rate, --hybrid-fetch, --mem-ratio, --gpu, --no-graph, --num-tokens, --kv-reserve-tokens) with a bare '--' passthrough: everything after the separator is forwarded verbatim to ft serve. The bench's own defaults (--moe-cache-auto, --cuda-graph-max-bs 1, --max-running-requests 1, --max-seq-len-override) are applied only when the passthrough does not set them.

Example:
```sh
python benchmarks/bench_decode_moe.py --model ~/model/Qwen3.6-35B-A3B-NVFP4 \
  --backend offload --num-tokens 131072 --kv-reserve-tokens 131072 \
  --moe-cpu-layers 0
```
( pass each extra argument through manually )

→

```sh
python benchmarks/bench_decode_moe.py --model ~/model/Qwen3.6-35B-A3B-NVFP4 \
  --backend offload \
  -- --num-tokens 131072 --kv-reserve-tokens 131072 --moe-cpu-layers 0
```


yuyi2439:
#212 provides `--extra-args`, which implements similar feature

yuyi2439:
@jason-fxz look this please

yuyi2439:
I reserved original `--backend` and `--model`. We can also merge these two to use `--`

PR #266, opened 2026-08-29
fix(server): hide qwen transport markers from semantic responses

### LINKED ISSUE #265: Qwen semantic responses can leak transport special-token markers
## Problem

Qwen-family decoding needs protocol tags available long enough for the reasoning and tool-call parsers, but that can also leave transport-only markers such as `<|im_start|>`, `<|im_end|>`, and multimodal padding sentinels in semantic chat-style response content.

These markers are protocol framing, not assistant-visible content. They should be suppressed at the protocol-neutral semantic generation boundary, after reasoning and tool parsing have consumed their tags.

## Expected behavior

- OpenAI chat, Anthropic messages, and Responses API outputs do not expose Qwen transport/internal multimodal markers.
- Streaming and buffered semantic responses agree even when a marker is split across decoder chunks.
- Reasoning tags, tool-call tags, and grounding markers remain available to their parsers or clients as appropriate.
- A deliberately quoted marker spelling remains visible as literal assistant content.
- Raw `/generate`, legacy completions, and offline token-ID behavior remain unchanged.

## Scope

A focused fix can live in the shared semantic generation layer (`python/freetoken/server/generation.py`) with request-local, chunk-safe filtering and targeted server tests. No tokenizer or raw-generation behavior needs to change.

HaileyStorm:
## Summary

- suppress Qwen transport and internal multimodal markers at the protocol-neutral semantic generation boundary
- run suppression only after reasoning and tool parsing, keeping parser tags, grounding markers, FIM markers, and tool arguments intact
- preserve matching quoted/code literals while making marker matching chunk-safe and event-order-safe across streaming tool calls
- leave raw `/generate`, legacy completions, non-Qwen semantic output, and offline token IDs unchanged

Closes #265.

## Why this layer

The Qwen detokenizer must keep protocol tags visible long enough for reasoning and tool parsers to consume them. Filtering in the tokenizer would also alter raw-generation and offline compatibility surfaces. `generation.py` is the shared semantic waist used by OpenAI chat, Anthropic Messages, and Responses, so the cleanup happens once after parsing and before those adapters format their wires.

The filter is request-local. It holds only an ambiguous marker prefix or the suffix of a quote after a complete hidden marker. Structured tool events encountered during that ambiguity are retained in the same sequence, preventing both marker leaks and content/tool reordering. An unmatched quote is sanitized at end-of-stream; a matching quote proves the marker spelling was deliberate literal content.

## Validation

Original symptom: observed with a local Qwen3.8 Flash-Next derivative on an NVIDIA RTX 5090, then minimized to deterministic CPU-only server tests. This PR makes no performance claim and does not require model weights or GPU execution.

Ubuntu Linux, GPU visibility explicitly disabled:

```bash
PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=python CUDA_VISIBLE_DEVICES='' HIP_VISIBLE_DEVICES='' \
/home/hailey/AI/llm/src/freetoken-flash-next-9ef36513/.venv/bin/python -m pytest -q -p no:cacheprovider \
tests/server/test_qwen_semantic_special_tokens.py \
tests/server/test_streaming_model_matrix.py \
tests/server/test_generation_accounting.py \
tests/server/test_openai_api.py
```

Result: `198 passed, 1 warning in 2.89s`. The warning is the existing FastAPI/Starlette TestClient deprecation warning.

Also passed:

- `py_compile` for both changed Python files
- `git diff --check origin/main...HEAD`
- independent closure review, including 317,449 bounded text/chunk/event probes

## Compatibility tests

The regression suite explicitly covers:

- all hidden markers split at every character boundary
- stream/buffer parity for reasoning, content, and structured tool calls
- overlapping and disproved marker prefixes without event reordering
- closed, escaped, typographic, and unterminated quotes
- preserved grounding and FIM markers
- non-Qwen semantic output
- raw `/generate`, legacy completion output, and offline token IDs

### DIFF
--- a/python/freetoken/server/generation.py
+++ b/python/freetoken/server/generation.py
@@ -42,6 +42,250 @@ from .reasoning_parser import (
 )
 
 
+# Qwen's decoder intentionally keeps protocol markers visible so reasoning and
+# tool-call parsers can consume their (non-special-token) tags.  At the semantic
+# API waist, after those parsers, only transport/internal multimodal markers are
+# suppressed.  Grounding markers (object_ref/box/quad), think tags, tool tags,
+# and FIM/repository markers are deliberately absent from this list.
+_QWEN_SEMANTIC_HIDDEN_TOKENS = (
+    "<|endoftext|>",
+    "<|im_start|>",
+    "<|im_end|>",
+    "<|vision_start|>",
+    "<|vision_end|>",
+    "<|vision_pad|>",
+    "<|image_pad|>",
+    "<|video_pad|>",
+    "<|audio_start|>",
+    "<|audio_end|>",
+    "<|audio_pad|>",
+    "<tts_pad>",
+    "<tts_text_bos>",
+    "<tts_text_eod>",
+    "<tts_text_bos_single>",
+)
+
+
+def _uses_qwen_semantic_protocol(state: Any) -> bool:
+    """Whether this server is configured for a Qwen semantic output dialect.
+
+    Parser selection is the stable frontend-visible family signal available in
+    ``ServerArgs``.  Either half is enough because users may explicitly disable
+    reasoning or omit tools for an individual request.
+    """
+    config = state.config
+    return (
+        getattr(config, "reasoning_parser", None) == "qwen3"
+        or getattr(config, "tool_call_parser", None) in {"qwen", "qwen25", "qwen3_coder"}
+    )
+
+
+class _SemanticSpecialTokenFilter:
+    """Request-local, chunk-safe suppression of semantic protocol markers.
+
+    A proper marker prefix is retained until the next detokenizer delta, so a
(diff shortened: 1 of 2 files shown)


gdevenyi:

The symptom is real on this deployment: asked to repeat `<|im_start|>user hello<|im_end|>`, the chat content came back with the `<|im_start|>` marker in it; with this PR the content is `user hello`. Greedy outputs on three unrelated prompts are word-identical to the unfiltered build, and the 8-question probe is unchanged. Merged clean, in production.

One thing outside this PR's scope that the same probe showed: the tokenizer parses control tokens typed by a user in a message (a `<|vision_pad|>` in user text vanished from the prompt), so a typed `<|im_end|>` ends the turn early. Input side, not this PR.

PR #269, opened 2026-08-29
fix(server): report why a tokenizer/detokenizer worker died during load

sime2408:
_run_scheduler wraps startup and pushes an ("error", reason) ack before dying, so the supervisor can report the real cause. _run_tokenize_worker did not, so every tokenizer-side startup failure reached the parent only as a dead process and surfaced as the generic "backend worker freetoken-detokenizer-0 exited during load" — the actionable reason (unreadable tokenizer.json, missing checkpoint file, ZMQ bind clash) was lost.

Wrap tokenize_worker in the same try/report/re-raise the scheduler already uses. The traceback still prints and the process still exits non-zero; only the ack is added. No ack_queue (direct call) stays a plain propagate.

Adds tests/server/test_launch_worker_errors.py, which fails before this change.

### DIFF
--- a/python/freetoken/server/launch.py
+++ b/python/freetoken/server/launch.py
@@ -52,7 +52,18 @@ def _run_tokenize_worker(detach: bool, **kwargs) -> None:
         _detach_process_group()
     from freetoken.tokenizer import tokenize_worker
 
-    tokenize_worker(**kwargs)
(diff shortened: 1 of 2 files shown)


gdevenyi:

Merged clean; the two unit tests pass here. No tokenizer-side startup failure to show on this box, so the ack path itself is only covered by the tests.

Issue #252, opened 2026-08-28
[Feature Request] Support compressed-tensors per-channel FP8 checkpoints for Qwen3.5 MoE


yongjer:
## Summary

I would like to request support for loading `compressed-tensors` FP8 checkpoints that use per-channel weight scaling, specifically for models such as:

`ornith-ai/Ornith-1.5-35B-A3B-FP8`

The model architecture itself appears to already be supported by FreeToken through the existing `Qwen3_5MoeForConditionalGeneration` / `qwen3_5_moe` implementation.

The missing part seems to be support for this checkpoint's FP8 serialization and quantization scheme.

## Model

Hugging Face model:

`ornith-ai/Ornith-1.5-35B-A3B-FP8`

Architecture:

```json
"architectures": [
  "Qwen3_5MoeForConditionalGeneration"
],
"model_type": "qwen3_5_moe"
```

The checkpoint uses `compressed-tensors` with FP8 weights:

```json
"quant_method": "compressed-tensors",
"format": "float-quantized"
```

The relevant quantization scheme is approximately:

- FP8 weights
- static per-channel weight scaling
- dynamic per-token activation scaling
- selected modules excluded from quantization

This differs from the FP8 formats currently handled by the Qwen3.5 MoE loader, such as block FP8 and ModelOpt-style FP8.

## Current behavior

FreeToken already has support for the Qwen3.5 MoE model architecture, so no new model implementation should be required.

However, the current quantization detection and weight-loading paths do not appear to recognize `compressed-tensors` per-channel FP8 checkpoints.

As a result, this checkpoint cannot be loaded using its intended FP8 representation.

## Requested support

It would be useful to add a dedicated loading path for `compressed-tensors` FP8 checkpoints with per-channel weight scales.

A possible initial scope could be:

1. Detect `compressed-tensors` FP8 checkpoints with `float-quantized` format.

2. Support FP8 tensors accompanied by per-channel `weight_scale` tensors.

3. Correctly handle scale broadcasting according to the stored tensor shape rather than assuming per-tensor or block scaling.

4. Support Qwen3.5 MoE expert weights, including stacked expert banks.

5. Respect the checkpoint's quantization ignore/exclusion rules.

6. Support weight and scale tensors located in different safetensors shards.

## Possible incremental implementation

A first implementation does not necessarily need native per-channel FP8 MoE execution.

For correctness-first support, FreeToken could initially load:

```text
FP8 weight + per-channel scale
              |
              v
[...]

PR #275, opened 2026-08-29
fix: [BUG] qwen3_5 dense: mixed-precision NVFP4 crashes in ct_bf16_fuse (Float8 × BFloat16 promotion)

### LINKED ISSUE #238: [BUG] qwen3_5 dense: mixed-precision NVFP4 crashes in ct_bf16_fuse (Float8 × BFloat16 promotion)
# [BUG] qwen3_5 dense checkpoints: mixed-precision NVFP4 (FP8 attention + NVFP4 MLP) crashes in `ct_bf16_fuse` — Float8_e4m3fn × BFloat16 promotion

## Environment

- FreeToken **0.1.2** (PyPI wheel `freetoken-0.1.2-cp312`, kernel-cache `0.1.2+cu130`)
- Python 3.12.13 (uv-managed venv), Ubuntu 22.04, driver 580.119.02
- GPU: 1× NVIDIA RTX 4090 (also reproduced logic on a 2×4090 box; TP=1)
- Checkpoint: `unsloth/Qwen3.8-27B-NVFP4` (llm-compressor export, `quantization_config.format == "mixed-precision"`)

## Summary

Serving a Qwen3.8-27B dense checkpoint (qwen3_5 family) quantized with the standard NVFP4 recipe — **attention/GDN projections FP8, rest NVFP4** — crashes during weight loading. Both `ft serve` and `ft checkpoint` fail identically, which suggests the shared compressed-tensors walker lacks a dequant branch for **Float8-quantized linears** when they land in a bf16 fuse group.

## Reproduction

```bash
ft serve --model unsloth/Qwen3.8-27B-NVFP4 --model-source modelscope
# same failure with:
ft serve --model unsloth/Qwen3.8-27B-NVFP4 --model-source modelscope --nvfp4-backend flashinfer
ft checkpoint --model <snapshot-dir> --out <out-dir>
```

## Error

```
RuntimeError: Promotion for Float8 Types is not supported, attempted to promote Float8_e4m3fn and BFloat16
```

Key frames (identical for serve and checkpoint):

```
freetoken/engine/engine.py(459)        _load_weight_state_dict
freetoken/models/weight.py(251)        load_weight
freetoken/models/qwen3_5_moe/weight.py(648)  _iter_weights_compressed_tensors
freetoken/models/qwen3_5_moe/weight.py(589)  _emit_bf16_weight
freetoken/models/loader.py(251)        ct_bf16_fuse  →  torch.cat([...], dim=0)
```

## Analysis

- The checkpoint's `config.json` reports `quantization_config.format == "mixed-precision"` with 303 ignored modules — the usual llm-compressor NVFP4 recipe (FP8 for attention/GDN linears, NVFP4 for the rest).
- The NVFP4 (`weight_packed` 4-bit) parts dequantize fine — loading progresses ~5s deep before failing.
- The crash happens when a **fuse group** for GDN `in_proj` is concatenated: one member tensor is still **raw Float8_e4m3fn** (an FP8-quantized linear that the compressed-tensors walker passed through unconverted), and `torch.cat` refuses to promote Float8 against BFloat16.
- `--nvfp4-backend triton|flashinfer` makes no difference (failure is upstream of kernel backends).

## Expected behavior

Either (a) Float8-quantized linears in compressed-tensors checkpoints are dequantized to the compute dtype before fusion/emit, or (b) a clear early error naming the unsupported mixed format.

## Notes

- A single-recipe export (`Qwen/Qwen3.8-27B-FP8`, pure FP8) presumably takes the mainline FP8 path and is unaffected — not yet tested.
- Side note: `docs/models.md` does not list Qwen3.8 yet, though the Desktop catalog does; might be worth a docs refresh alongside the fix.
- Happy to test a patch.

yuyi2439:
Fixes #238

### DIFF
--- a/python/freetoken/models/loader.py
+++ b/python/freetoken/models/loader.py
@@ -182,6 +182,9 @@ class ShardReader:
     def names_in(self, file: str) -> list[str]:
         return [name for name, shard in self._map.items() if shard == file]
 
+    def __contains__(self, name: str) -> bool:
+        return name in self._map
+
     def get_tensor(self, name: str) -> torch.Tensor:
         import safetensors
 
--- a/python/freetoken/models/qwen3_5_moe/weight.py
+++ b/python/freetoken/models/qwen3_5_moe/weight.py
@@ -645,7 +645,11 @@ def _iter_weights_compressed_tensors(
                     continue
 
                 if name.endswith(".weight"):
-                    yield from _emit_bf16_weight(name, reader.get_tensor(raw_name))
+                    # Some exports (e.g. GDN in_proj_{qkv,z}) quantize a bare ``.weight`` to
+                    # per-tensor FP8 (no weight_packed); ``reader`` supports ``in`` across all
+                    # shards, so a sibling scale in a different shard is still found.
+                    tensor = _load_maybe_quantized(reader, raw_name, reader)
+                    yield from _emit_bf16_weight(name, tensor)
                     continue
 
                 # A_log / dt_bias (kept fp32 by the model; the load downcast exempts them).


dcutugno:
Tested this approach out-of-tree — confirms it fixes #238 on a second checkpoint variant and a second platform.

I'm on a compiled build, so I couldn't apply the diff literally. Instead I patched one level down at `ShardReader.get_tensor` — the exact call this PR replaces — reusing the engine's own `_dequant_fp8_weight` for the bare-FP8 `.weight` case. Semantics should match the PR: verified byte-identical to a manual `w.to(bf16) * scale`, with `weight_scale`, `weight_packed` and plain BF16 tensors passing through untouched.

**Env:** freetoken `0.1.2+g932e87e8a`, Python 3.12, Windows 11, sm_120, 16 GB, TP=1.
**Checkpoint:** a third-party Qwen3.8-27B multimodal NVFP4 export (`Qwen3_5ForConditionalGeneration`, compressed-tensors 0.18.0, `format: "mixed-precision"` — 232 `float-quantized` FP8 targets + 168 `nvfp4-pack-quantized`).

`ft checkpoint` was failing at ~2 s; it now completes:

```
wrote FTW checkpoint
  tensors: 875 weight + 0 experts_bank
  FTW: 30.06 GiB across 4 shard(s) (<= 8.0 GiB each)
  converted in 41.6s
[patch] dequantized 232 FP8 tensors: 8.71 GiB -> 17.42 GiB
```

`ft serve` pointed at the source dir: the Float8 promotion crash is gone. It now loads for ~38 s and stops on VRAM, which is expected for a dense 27B on a 16 GB card and unrelated to this PR:

```
Free memory before loading model: 14.62 GiB
OutOfMemoryError: CUDA out of memory. Tried to allocate 2.37 GiB.
GPU 0 has a total capacity of 15.89 GiB of which 0 bytes is free.
Of the allocated memory 23.13 GiB is allocated by PyTorch
```

Two caveats worth stating plainly:

- **Numerical correctness is unverified.** I could not run a token through the model on this hardware, so this confirms the crash is fixed and the pipeline completes — nothing more. Relatedly, the output carries 875 weight tensors where a pure-FP8 export of the same model produces 915. That is plausibly just different fusion between the two quant layouts, but I can't confirm it from here.
- **Footprint.** Dequantizing to bf16 takes those 232 tensors from 8.71 GiB to 17.42 GiB, so the resulting FTW (30.06 GiB) is larger than the 23 GiB source checkpoint. Keeping them FP8, as the `quant_method: "fp8"` path already does, would avoid that. Possibly out of scope for this fix, but it does make the mixed-precision path meaningfully more expensive than the pure-FP8 one.

One question: #238's traceback lands at `loader.py:251` inside `torch.cat`, while mine lands at `loader.py:256`. That may just be line drift between the PyPI wheel and my dev build, but if `ct_bf16_fuse` genuinely has two Float8-reachable sites, it would be worth confirming this change covers both.

dcutugno:
Correction to my report above — I spoke too soon. The conversion completing is not the end of the story: **the resulting FTW does not load.**

Serving it fails downstream, in model construction rather than in the weight walker:

```
File "freetoken/kernel/triton/nvfp4_linear.py", line 867, in load_state_dict
  s = state_dict.pop(_concat_prefix(prefix, "weight_scale"))
KeyError: 'model.layers.0.linear_attn.out_proj.weight_scale'
```

Root cause, and I don't think it is in scope for this PR: `detect_compressed_tensors_nvfp4()` in `models/config.py` returns `True` when *any* `config_groups` entry is NVFP4, and its docstring states the assumption it makes — "All `targets: ["Linear"]` are NVFP4 except the per-module `ignore` list". That does not hold for `format: "mixed-precision"`. `qwen3_5_moe/config.py` then forces `attn_quant = "nvfp4"`, so `linear_attn.out_proj` is built as `Nvfp4DenseLinear` and demands an NVFP4-form `weight_scale`. Dequantizing that tensor to bf16 (this PR's approach, and what I emulated) removes the scale entirely, so the KeyError is the direct consequence.

The deeper issue is that quantization is resolved into three global scalars (`expert_quant` / `attn_quant` / `dense_quant`), and this checkpoint cannot be described by any single assignment:

| modules | quant |
|---|---|
| attention + GDN projections, all 64 layers | per-channel FP8 |
| MLP, layers 0-55 | NVFP4 |
| MLP, layers 56-63 | FP8 |

Fixing `attn_quant` alone would just move the failure to layers 56-63 via `dense_quant`.

So my earlier conclusion needs revising: this change removes the Float8 promotion crash, but on a mixed-precision checkpoint it produces an FTW the engine cannot consume. It converts, then dies at load.

Worth noting #208 appears to target exactly this, with per-module native storage reusing the existing `Fp8PerTensor*` kernels rather than dequantizing — and reports 21.8 GiB resident vs ~54 GB fully dequantized, with working generation. That also addresses the footprint point I raised above: the FP8 group here is per-row scaled, which `Fp8PerTensorLinear` already stores natively (`weight` fp8 + per-row fp32 `weight_scale [out]`), so the bf16 expansion is avoidable rather than inherent.

Is this PR still wanted as a narrower stopgap for checkpoints that only trip the loader path, or is it superseded by #208? Happy to test either.

yuyi2439:
Thanks for the detailed test. To clarify: I have not tested the FTW conversion path. I advocate on-the-fly dequantization, because this path saves VRAM.

#343 implements mixed-precision detection, but only for experts loading. (https://github.com/FlashML-org/FreeToken/issues/238#issuecomment-5514703766)

salekseev:
We got the same failure with a different checkpoint, and this patch corrects it.

Our checkpoint is `unsloth/Qwen3.6-35B-A3B-NVFP4-Fast`, which is an MoE model. #238 reports a dense 27B model. The error text is the same:

```
RuntimeError: Promotion for Float8 Types is not supported, attempted to promote Float8_e4m3fn and BFloat16
```

The failure happens about 5 seconds into the weight load, as #238 describes.

Environment: one RTX 4080 SUPER, 16 GiB, sm_89, CUDA 13.3 image, TP=1. This card uses the same architecture as the RTX 4090 in #238, so we do not add architecture coverage. We add a second checkpoint, and a second family member: MoE in place of dense.

The patch applies to `main` af71ba4 without conflict. After the patch, the Float8 promotion failure stops.

The load then stops at a later problem, which is separate from this pull request. We describe that problem in #252.

Issue #151, opened 2026-08-24
`ft bench bw` picks `hybrid` on Ampere, but `offload` is 8.3x faster (DeepSeek-V4-Flash, 2x RTX 3090)


GlidLov:
# `ft bench bw` picks `hybrid` on Ampere, but `offload` is 8.3x faster (DeepSeek-V4-Flash, 2x RTX 3090)

## Summary

On a 2x RTX 3090 (SM86) host, `ft bench bw` measures the CPU MoE kernel at **56.2 GB/s** vs **12.3 GB/s** for PCIe, reports a 4.57x ratio, and selects `hybrid`. In actual serving, `hybrid` delivers **0.67 tok/s** while forcing `--moe-backend offload` on the same machine and model delivers **5.58 tok/s**.

The auto-selected backend is **8.3x slower** than the one it rejected.

The calibrated CPU number does not survive contact with serving: at 0.67 tok/s and ~3.07 GB of expert weights per token, the effective CPU-MoE rate is roughly **2 GB/s**, about **28x below** what `ft bench bw` measured for the same kernel, dtype and workload minutes earlier.

## Environment

| | |
|---|---|
| FreeToken | 0.1.2 (PyPI, `freetoken[accel]`) |
| Model | `deepseek-ai/DeepSeek-V4-Flash-0731` (fp8 dense, `expert_dtype: fp4`) |
| GPU | 2x RTX 3090 24GB, SM86, driver 595.84 |
| CPU | Intel Xeon Gold 6252 @ 2.10GHz, avx512f, 46 cores available to the container |
| RAM | 503 GB |
| CUDA / torch | 13.0 / 2.11.0+cu130 |
| OS | Linux x86_64 (container) |

## Reproduction

```bash
ft bench bw --model dsv4 -o bwprofile.json

ft serve --model-path /path/to/DeepSeek-V4-Flash-0731 \
  --tp-size 2 --moe-backend hybrid --moe-cache-auto \
  --host 0.0.0.0 --port 1919
# -> 0.67 tok/s

ft serve --model-path /path/to/DeepSeek-V4-Flash-0731 \
  --tp-size 2 --moe-backend offload --moe-cache-auto \
  --host 0.0.0.0 --port 1919
# -> 5.58 tok/s
```

Decode measured over a streamed 60-token completion, first token excluded, after a warmup request.

## `ft bench bw` output

```
[...]

PR #278, opened 2026-08-29
fix(moe): match the benchbw verdict to the served model's expert size

bernimccoy:
Fixes the 3x decode regression of `--moe-backend auto` on small-expert NVFP4 models
(Qwen3.6-35B-A3B-NVFP4 on an RTX 4070 SUPER: auto -> hybrid 27 tok/s, offload 84 tok/s).

## Problem

`load_backend_recommendation` keys the profile on the expert format only and prefers the
`ft bench bw --dtype` verdict, measured on one geometry per format (nvfp4: 3072x1536, 7.6 MiB
experts). The CPU MoE kernel does not scale down to small experts: the same box benches
89.7 GB/s there and 32.2 GB/s at Qwen3.6-35B-A3B's 2048x512 (1.7 MiB), below the 2x PCIe
threshold. So auto picks hybrid and decodes 3x slower than offload. `ft bench bw --model
qwen3.6-moe` already records offload for this geometry, but the reader consults the
per-format `dtypes` entry first and never reaches the per-model one.

## Fix

`load_backend_recommendation` / `load_hybrid_fetch_fraction` take the served model's
`geometry` (H, I, E, top_k) and `expert_bytes`:

- a per-model workload entry benched at that geometry wins;
- otherwise the dtype entry applies only if its `expert_bytes` is within
  `EXPERT_BYTES_TOLERANCE` (2x) of the model's; other workload entries are filtered the same
  way; a skipped entry logs once with the `ft bench bw --model` remedy;
- callers that pass no geometry, and profiles written before `expert_bytes` was recorded,
  get today's format-only match.

The engine passes both at the backend pick and at `--moe-hybrid-max-fetch auto`. No profile
format change.

## Tested on

RTX 4070 SUPER 12 GB (sm_89), driver 591.86, WSL2, torch 2.11.0+cu130, 24-core host, 108 GB
RAM. `nvidia/Qwen3.6-35B-A3B-NVFP4`, `--moe-cache-auto` (3831 slots), bs=1.

`benchmarks/bench_decode_moe.py --backend offload,cpu,hybrid` (256 decode tokens):

| backend | decode tok/s |
|---|---|
| offload | 83.7 |
| cpu | 25.9 |
| hybrid (what auto picked on `main`) | 27.2 |

`ft bench bw`: dtype nvfp4 89.7 GB/s CPU vs 25.3 PCIe (hybrid); `--model qwen3.6-moe`: 32.2
vs 25.9 (offload). With this change and the same profile, `ft serve` (auto) logs
`benchbw profile: the 'nvfp4' entry was benched on 7.61 MB experts, this model's are 1.69 MB
(4.5x apart); not applying its verdict`, picks offload, and streams 91.0 tok/s on the request
that gave 26.4 tok/s under auto -> hybrid.

`tests/moe/test_hybrid_fetch.py`: new case with the measured profile numbers (exact-geometry
entry wins; a 4.5x-off dtype verdict is skipped; a 1.3x-off one still applies; no-geometry
callers unchanged); 6/6 pass. bf16 Qwen3-30B-A3B (9.4 MiB experts) is within 2x of the bf16
dtype geometry, so its verdict is unchanged.

## Caveats

The 2x tolerance is a judgment call from one box (the kernel ratio here was 2.8x between the
two geometries); a per-model bench stays the precise answer and the warning points at it.

Assisted by: [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/python/freetoken/engine/engine.py
+++ b/python/freetoken/engine/engine.py
@@ -652,8 +652,11 @@ class Engine:
         from freetoken.moe.bench_profile import load_hybrid_fetch_fraction
 
         gpu_name, gpu_uuid = _profile_gpu(self.device.index)
+        model_config = config.model_config
         fraction = load_hybrid_fetch_fraction(
-            cache.quant_format, gpu_name=gpu_name, gpu_uuid=gpu_uuid
+            cache.quant_format, gpu_name=gpu_name, gpu_uuid=gpu_uuid,
+            expert_bytes=_model_expert_bytes(cache.quant_format, model_config),
+            geometry=_model_geometry(model_config),
         )
         if fraction is None:
             cache.hybrid_max_fetch = 1
@@ -995,6 +998,32 @@ class Engine:
         self.graph_runner.destroy_cuda_graphs()
         torch.distributed.destroy_process_group()
         destroy_distributed()
+
+
+def _model_geometry(model_config) -> dict | None:
+    """The served model's MoE geometry in benchbw's workload terms, or None for a dense model."""
+    try:
+        geometry = {
+            "hidden": int(model_config.hidden_size),
+            "inter": int(model_config.moe_intermediate_size),
+            "experts": int(model_config.num_experts),
+            "top_k": int(model_config.num_experts_per_tok),
+        }
+    except (AttributeError, TypeError, ValueError):
+        return None
+    return geometry if geometry["experts"] > 0 and geometry["inter"] > 0 else None
+
+
+def _model_expert_bytes(bench_fmt: str, model_config) -> int | None:
+    """Per-expert offload-bank bytes for this model in ``bench_fmt`` (the same per-format
+    sizing ``bank_bytes_estimate`` uses for the pin budget), or None for a format without one."""
+    from freetoken.moe.offload_cache import _BANK_BYTES_PER_EXPERT
+
+    geometry = _model_geometry(model_config)
+    per_expert = _BANK_BYTES_PER_EXPERT.get(bench_fmt)
+    if geometry is None or per_expert is None:
+        return None
+    return int(per_expert(geometry["hidden"], geometry["inter"]))
 
 
 def _profile_gpu(index: "int | None" = None) -> Tuple[str | None, str | None]:
@@ -1371,7 +1400,15 @@ def _adjust_config(config: EngineConfig):
         from freetoken.moe.bench_profile import load_backend_recommendation
 
         gpu_name, gpu_uuid = _profile_gpu()
-        if load_backend_recommendation(bench_fmt, gpu_name=gpu_name, gpu_uuid=gpu_uuid) == "hybrid":
+        # The profile is keyed on the expert format; the model's geometry / expert size picks the
+        # entry that was benched on comparable experts (a small-expert model must not inherit a
+        # verdict measured on 4x larger ones).
+        recommendation = load_backend_recommendation(
+            bench_fmt, gpu_name=gpu_name, gpu_uuid=gpu_uuid,
+            expert_bytes=_model_expert_bytes(bench_fmt, model_config),
+            geometry=_model_geometry(model_config),
+        )
+        if recommendation == "hybrid":
             from freetoken.moe.cpu_executor import compiled_extension_supports
 
             _act = getattr(model_config, "hidden_act", "silu")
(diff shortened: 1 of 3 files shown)


gdevenyi:
This fixes the wrong auto pick on my box too, and it is exactly the mechanism described: the `nvfp4` profile entry here was benched on 7.61 MB experts while Qwen3.8-Flash-Next's are 2.64 MB.

|---|---|---|
| `--moe-backend auto` resolves to | hybrid (measured 28.9 tok/s single-stream at TP=1, vs 53-59 for offload) | **offload** |
| log | (silent) | `benchbw profile: the 'nvfp4' entry was benched on 7.61 MB experts, this model's are 2.64 MB (2.9x apart); not applying its verdict. Run ft bench bw --model ...` |
| auto, TP=2 decode single / 8-conc | n/a (hybrid) | 90.7 / 320.5 tok/s, same as explicit offload (90.8 / 326.0) |

Applies cleanly to main. The warning text is the right call: it tells the operator what to run instead of silently picking. One suggestion: on a model whose experts are this far from the benched geometry, `auto` could also skip hybrid outright when `top_k * expert_bytes` per layer is below the PCIe-vs-CPU crossover the profile already knows, so the fallback is correct even without a per-model bench.

A-XIANGQAQ:
Confirming this fixes our case too, with one gap for this model family.

**Box:** RTX 4090 48 GB (sm_89), 38-core AVX512-VNNI host, `RadixArk/Qwen3.8-Flash-Next-NVFP4`. The `nvfp4` dtype entry here is benched on 7.61 MB experts; this model's are **2.64 MB** (2560×640×3, NVFP4 + FP8 block scales) — 2.9x apart, same as @gdevenyi. `auto` currently resolves to hybrid; serving measures **hybrid 46.6 tok/s vs offload 62.6–63.6** (27% slower), so the guard this PR adds is exactly what we need.

**Gap:** the remedy the fix logs (`run ft bench bw --model ...`) has no matching entry for this model. The `--model` list is dsv4, gemma4-26b, glm4.7-nvfp4, glm5.3-flash-nvfp4, gpt-oss-120b, gpt-oss-20b, minimax-m2.5, qwen3-30b, qwen3.6-moe — Qwen3.8-Flash-Next isn’t in it, and `--formats` only overrides the format, not the geometry. So here the outcome would be “offload + a warning we cannot act on”.

Would it make sense to (a) name the nearest geometry in the warning, or (b) add a `qwen3.8-flash-next` workload entry (H=2560, I=640, E=512, top_k=10)? Happy to run the bench and post the numbers if you want the entry.

Reference numbers from our box: dtype `nvfp4` → hybrid at 5.36x (CPU 135.6 vs PCIe 25.3 GB/s, `avx512f+avx512vnni`); `qwen3.6-moe` geometry → hybrid at 3.74x (CPU 94.5 vs PCIe 25.3).

A-XIANGQAQ:
Follow-up with a data point, because the gap I mentioned can be worked around in one line: adding

```python
"qwen3.8-flash-next": Workload("qwen3.8-flash-next", 2560, 640, 512, 10, ("nvfp4",)),
```

to `WORKLOADS` makes `ft bench bw --model qwen3.8-flash-next` run. With the geometry now matching exactly (2.64 MB experts), the verdict on this box is **still hybrid**:

```
ceilings: CPU STREAM read 158.9 | PCIe linear H2D 25.3  D2H 26.4 GB/s  (threshold 2.0x)
qwen3.8-flash-next  H=2560 I=640 E=512 top_k=10
  nvfp4  2.64 MB   CPU-MoE 112.7 GB/s   PCIe-gather 25.2 GB/s   4.46x  -> hybrid
         overlapped: CPU-MoE 95.3 + PCIe 19.7 GB/s -> hybrid fetches 17.1% of misses
```

CPU kernel rate by geometry on this box: **135.6 GB/s** at the 7.61 MB dtype geometry → **112.7** at this model's 2.64 MB → **94.5** at `qwen3.6-moe`'s 1.69 MB. The scaling is real, but nothing here crosses the 2x threshold.

Serving disagrees: offload **62.6–63.6 tok/s** vs hybrid **46.6** (256 output tokens, streamed, client-side timing, idle server) — 27% slower.

What the utilization says about why: under hybrid the GPU sits at 73–100% and the CPU never exceeds 1–18%. Neither side is saturated, so the 27% is not bandwidth — it looks like the per-step sync/latency of splitting each layer's 1–2 token GEMV between the CPU pool and the GPU step. That is also what #151 measured when the benched CPU rate did not survive serving.

So on this box the geometry guard alone would still resolve to hybrid. If it is useful for the design: what separates the two verdicts here is the **decode batch size**, not the expert size — the bench measures the CPU kernel at a synthetic batch, while serving runs bs≈1–2.

Data point only. Happy to re-run if a bs=1 variant of the kernel bench would help settle it.

PR #257, opened 2026-08-28
feat(qwen4_exp): support Qwen3.8-Flash-Next

jason-fxz:
## What

Support [Qwen3.8-Flash-Next](https://huggingface.co/Qwen/Qwen3.8-Flash-Next) (`qwen4_exp`), text-only.

Model structure:
- 36 GDN (gated delta net) + 12 QSA (compressed block-sparse attention) layers
- 4-stream hyper-connection residual
- PLE n-gram embedding (47.7 GiB table, kept in pinned host memory, UVA gather)
- 512 routed experts top-10 + gated shared expert; NVFP4 ([RadixArk/Qwen3.8-Flash-Next-NVFP4](https://huggingface.co/RadixArk/Qwen3.8-Flash-Next-NVFP4)) and official block-FP8 checkpoints

Engine support:
- CUDA graph decode (QSA selection, PLE gather and n-gram hash all capture-safe)
- hybrid radix cache: prefix reuse with GDN/PLE state snapshots at page boundaries
- MoE expert offload / hybrid CPU co-execution
- vendored vLLM/SGLang triton kernels + an original radix block top-k

## How to run

```bash
# bs=1, 50k tokens reserved for KV, rest of VRAM to the expert cache
ft serve --model RadixArk/Qwen3.8-Flash-Next-NVFP4 \
    --moe-backend offload --moe-cache-auto \
    --max-running-requests 1 --kv-reserve-tokens 50000
```

## Performance

bs=1 decode, AIME25, with the command above. Host RAM: ~111 GiB pinned (63 GiB NVFP4 expert banks + 48 GiB PLE table), 128 GB recommended.

| GPU | decode |
|---|---|
| RTX 5090 (32 GB) | 65 tok/s |
| RTX 4090 (24 GB) | 36 tok/s |

PR #283, opened 2026-08-30
fix(models): load NVFP4 expert banks from checkpoints without a safetensors index

paulp83:
## Problem

Launching Qwen3.8-Flash-Next (the #257 model) with the offload MoE backend crashes at startup:

```
FileNotFoundError: [Errno 2] No such file or directory: '.../RadixArk--Qwen3.8-Flash-Next-NVFP4/snapshots/.../model.safetensors.index.json'
  File "python/freetoken/models/nvfp4_banks.py", line 222, in load_nvfp4_expert_source_banks_parallel
```

Reproduced with:

```
ft serve --model RadixArk/Qwen3.8-Flash-Next-NVFP4 \
    --moe-backend offload --moe-cache-auto \
    --max-running-requests 1 --kv-reserve-tokens 50000
```

The dense weights load fine ("Loading weights: 206/206"), then the scheduler dies when the offload cache builds the expert banks.

## Root cause

The RadixArk `Qwen3.8-Flash-Next-NVFP4` checkpoint ships **no `model.safetensors.index.json`**. Its weights are split across three shard families:

- `model-bf16-*.safetensors` — dense bf16 weights
- `model-plefp8-*.safetensors` — the n-gram PLE table
- `layer-{L}-experts-{A}-{B}.safetensors` — routed NVFP4 experts (48 layers × 512 experts)

The dense path (`iter_weights` → `iter_weight_files`) and the PLE path (`_ple_table_files`) both already handle a missing index by globbing `*.safetensors` and reading shard headers — which is why the dense load succeeds. But both NVFP4 expert-bank loaders in `models/nvfp4_banks.py` (serial and parallel) hard-opened the index file to build their tensor→shard map, so they raise `FileNotFoundError` as soon as the bank build starts. (`iter_expert_tensors_parallel`, which the parallel path delegates to, already has the no-index fallback — but it's only reached *after* the outer function crashes on the index read.)

## Fix

One helper in `models/nvfp4_banks.py`, used by both loaders:

- `_weight_map(folder)` — returns the name→shard map from `model.safetensors.index.json` when present, else builds it from each `*.safetensors` shard's header (the same fallback `iter_expert_tensors_parallel` already uses).

Checkpoints that do carry an index keep using it unchanged; this only adds the fallback.

## Tests

Added two tests in `tests/models/qwen4_exp/test_weight.py` (serial + parallel) that build a no-index checkpoint whose only tensors are routed NVFP4 experts — split across two shards named like the real `layer-*-experts-*` files — and assert the full 6-bank placement (gate/up fused on the output-row axis, down separate, scalar `weight_scale_2` broadcast into the per-row global banks). Both fail with the launch crash's exact `FileNotFoundError` before the fix, pass after.

- `tests/models/qwen4_exp/test_weight.py`: 24/24 pass.
- `tests/models/` + `tests/moe/`: identical results before/after (the pre-existing failures in this environment are all `FileNotFoundError: 'ninja'` from JIT-compiling CUDA extensions; 56 failed / 192 passed on baseline → 56 failed / 194 passed with the fix, +2 = the new tests).

## Validation against the real checkpoint

`_weight_map` on the actual snapshot dir maps all 296,475 tensors across the 206 shards (matching the dense loader's 206/206), and the NVFP4 expert spec regex then finds exactly 48 × 512 × 3 × 3 = 221,184 expert tensors — every tensor the bank loaders need.

### DIFF
--- a/python/freetoken/models/nvfp4_banks.py
+++ b/python/freetoken/models/nvfp4_banks.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
 import collections
+import glob
 import json
 import os
 import re
+import struct
 from dataclasses import dataclass
 from typing import Callable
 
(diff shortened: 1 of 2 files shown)


jason-fxz:
`RadixArk/Qwen3.8-Flash-Next-NVFP4` does ship `model.safetensors.index.json` (34 MB, 296,475 entries — the same count your `_weight_map` rebuilds from the headers). What was missing was the download: `download_hf_weight` never fetched it (#124), now fixed in #336. Closing.

Issue #281, opened 2026-08-29
Proposal: reproducible end-to-end benchmark / regression harness


tuxevil:
While testing Qwen3.6 GGUF support in #131 together with Turing support from #24, I ended up building a small end-to-end benchmark harness to answer questions that the existing kernel/microbenchmarks don't fully capture.

Before cleaning it up into a PR, I'd like to check whether something along these lines would be useful upstream.

## Motivation

FreeToken has excellent low-level benchmarks and tests, but comparing real serving configurations currently requires users to assemble their own methodology.

This makes results from different hardware, quantizations and MoE backends difficult to compare.

For example, on my test machine:

* Quadro RTX 4000 (sm_75, 8 GB)
* Ryzen 9 7945HX
* 96 GB DDR5
* PCIe Gen3 x16 (~13.17 GB/s measured H2D/D2H)
* Qwen3.6-35B-A3B

`ft bench bw` considered hybrid viable/recommended for several formats, but actual fixed-prompt inference consistently favored offload.

For NVFP4 in CPU Boost/performance mode:

```text
ft bench bw:
CPU MoE            41.17 GB/s
PCIe gather        13.27 GB/s
ratio               3.10
recommended         hybrid

actual Qwen3.6:
offload .93        37.16 tok/s
best hybrid        25.39 tok/s
```

Q4_K showed the same general pattern:

```text
offload .93        39.16 tok/s
hybrid cap8        26.60 tok/s
```

The microbenchmark is still useful as a viability signal, but the full pipeline adds scheduling, synchronization, expert routing and GPU starvation that aren't represented by bandwidth alone.

## Proposed performance suite

A model/backend-independent CLI or script that runs a reproducible set of serving tests against a FreeToken server:

### Fixed decode

* deterministic prompt
* fixed output budget
* multiple repetitions
* wall-clock end-to-end tok/s
* optional decode-window telemetry

### Prefill/context sweep

For example:

* ~512 tokens
* ~1K
* ~2K
* ~4K

Record TTFT and effective prompt throughput.

### Prefix cache

Measure:

* cold prefix
* identical cached prefix
* cached prefix + small suffix

### MoE backend comparison

Where supported:

* offload
* hybrid with configurable caps
* optionally cpu/fused

### Cache/residency sweep

Optionally test several `memory-ratio` / expert-cache configurations rather than relying on one default.

### Hardware metadata

Automatically record enough information to make community results comparable:

* FreeToken commit/version
* model/checkpoint
[...]

PR #287, opened 2026-08-30
fix(scheduler): hybrid radix hit corruption — GDN snapshot copy-on-donate + admission barriers

Cerynitius:
## Problem

Under the default `--cache-type radix`, hybrid (GDN) models intermittently serve corrupted answers on cache-**HIT** requests that are co-batched with in-flight decode. Measured with deterministic ground-truth probes on a production deployment: **~10% of hit requests wrong under concurrent load** — empty or garbled output, special-token leaks into content (`<|assistant|>` mid-text), and with a multimodal tower, answers describing a blank image. Cold prefill is always clean. Failures cluster in sticky per-branch windows (a poisoned branch keeps failing for a while, then self-heals when re-donated).

Every hybrid GDN model on the `hybrid_radix` / `linear_state` pool path is exposed; nothing about the trigger is model-specific.

## Root structure

`_cache_req_hybrid` donates the request's snapshot slot into the radix tree by **ownership transfer** (chunk commit donates the frozen ping-pong slot; finish donates the live slot "zero-copy"). That leaves one linear-state slot id shared between the tree and the request's still-in-flight pipeline. A late write through any retained alias poisons every subsequent hit of that branch — which is exactly the observed sticky-window signature.

Stream-order alone does not close the window: adding full-device syncs at the donate and restore points shrank the failure rate from ~10% to ~3% but not to zero. Severing the alias does.

## Fix

- **copy-on-donate**: the tree receives a *private clone* of the snapshot slot (`_clone_slot_for_tree`), written once behind the donate barrier and read-only for its tree lifetime. The request keeps its own slots — no replacement alloc at chunk commit; both ping-pong slots and the live slot are freed at finish.
- **Barriers** before hit-admission restores (`_restore_linear_states`) and donation bookkeeping (`_cache_req_hybrid`): request-level events, sub-millisecond, and they guarantee the clone's source is quiescent.

Cost: one ~MB slot copy per donation plus one device sync per prefill commit / hit admission. Decode throughput unchanged in before/after measurement.

## Validation

On a v0.1.2-based deployment serving GLM-5.3-Flash-NVFP4 (34 KDA layers riding the same `hybrid_radix`/`linear_state` path as qwen3.5 GDN; support PR #270) on an RTX PRO 6000 Blackwell 96 GB:

- A deterministic hammer alternating identical prompts with abandoned/saturating concurrent streams reproduced the corruption at 3/30 (image probes) and 2/24 (pure text) before the fix; after the fix: **0/60 image, 48/48 text rounds with zero corrupted answers**.
- An overnight ground-truth soak (~2,200 probes: exact-match arithmetic, needle retrieval, multi-turn recall, media A/B alternation with per-request content forensics) measured the pre-fix hit corruption at ~10% and confirmed request payloads and cache keys correct on every failure — isolating the fault to the hybrid hit path. Cold-path controls: 0/30.
- Post-fix regression battery: prefix-hit latencies unchanged (repeat/turn-2/long-document hits 0.8–1.1 s), LRU eviction under pool-overflow pressure clean, page-accounting integrity checks clean throughout.

The hammer scripts are small and self-contained; happy to attach them or port them into `tests/` if useful (they pair naturally with the regression-harness proposal in #281).

### DIFF
--- a/python/freetoken/scheduler/cache.py
+++ b/python/freetoken/scheduler/cache.py
@@ -338,14 +338,35 @@ class CacheManager:
             req.cache_handle = new_handle
             self.lock(new_handle)
 
+    def _clone_slot_for_tree(self, src: int) -> int:
+        """copy-on-donate: give the tree a PRIVATE clone of a GDN snapshot slot instead of
+        transferring ownership of the request's slot. Ownership transfer shares one slot id
+        between the still-in-flight request pipeline and the tree; a late write through any
+        retained alias then poisons every future prefix hit of that branch (observed as
+        ~10% corrupted hit answers under concurrent load, in sticky per-branch windows
+        that self-heal on re-donation). A private clone is written exactly once -- here,
+        after the donate barrier's full device sync, so the source is quiescent -- and is
+        read-only for the rest of its tree lifetime. Cost: one ~MB slot copy per donation."""
+        self.ensure_mamba_slots(1)
+        dst = self.linear_state_pool.alloc(1)[0]
+        self.linear_state_pool.copy_from(src, dst)
+        return dst
+
     def _cache_req_hybrid(self, req: Req, *, finished: bool) -> None:
         """Hybrid (GDN) cache_req: commit KV like radix AND manage the GDN state snapshot.
-        Prefill chunk commit: DONATE the frozen ping-pong slot (the snapshot the forward wrote
-        at the tracked ×64 boundary mamba_last_track_seqlen) into the tree; replace it with a
-        fresh slot if the tree took it (dedup keeps it for reuse). Finish: donate the live slot
-        (final full-sequence state, zero-copy since the req is done) and free the req's slots."""
+        Prefill chunk commit: donate a PRIVATE CLONE of the frozen ping-pong slot (the
+        snapshot the forward wrote at the tracked ×64 boundary mamba_last_track_seqlen)
+        into the tree; the request keeps its own slots. Finish: donate a clone of the live
+        slot (final full-sequence state) and free all of the req's slots."""
         from freetoken.kvcache.hybrid_radix_cache import HybridCacheHandle
 
+        # Donate barrier: settle all in-flight device work before snapshot donation and the
+        # dedup/re-point bookkeeping below. Stream-level wait_stream ordering alone leaves a
+        # window in which a hit admitted alongside in-flight decode restores or donates
+        # corrupted GDN state (reproduced with deterministic ground-truth probes: ~10% wrong
+        # answers under saturation, cold prefill always clean). Fires once per prefill
+        # commit / request finish -- request-level, sub-millisecond.
+        torch.cuda.synchronize(self.device)
         pool = self.linear_state_pool
         old_handle = req.cache_handle
         page_indices = self.page_table[req.table_idx, : req.cached_len]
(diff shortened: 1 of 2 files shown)


Cerynitius:
**Correction after deeper forensics — the corruption evidence in this PR's description does not implicate upstream.**

Slot-level checksum instrumentation overturned the ownership-aliasing theory. The true root cause of every corrupted hit we observed was in **our downstream model integration**: our KDA attention op (GLM-5.3 support, #270) never implemented the ×64-boundary track-snapshot write that the hybrid-radix design expects (`_write_track_snapshot` exists in the Qwen GDN op and is a per-op responsibility). Every reuse point our deployment donated therefore carried an **unwritten, all-zero recurrent state**; hits restored zero state and ran the prefix on the full-attention layers alone — usually survivable, intermittently wrong. That also explains why the failure rate seemed to respond to unrelated synchronization: it never did; the rate was constant model-derailment noise on top of a deterministic zero-state restore.

With the writer implemented on our side, restore-side checksums show real boundary state on every hit and 90 adversarial hammer rounds run clean — with or without the changes in this PR.

What remains of this PR on its own merits, with no corruption claim attached:
- **copy-on-donate** severs the slot-id sharing between a request's in-flight pipeline and the tree (structural hardening; cost ≈ one ~MB copy per donation);
- the admission/donate barriers are likely unnecessary given stream ordering and can be dropped;
- a `gdn_track_snapshots`-style guard (not in this PR) may be worth more than either: an op that lacks the snapshot writer currently donates unwritten state silently — a loud gate would have caught our integration gap immediately.

Happy to re-scope the PR to the guard + optional copy-on-donate, or close it — maintainers' call. Apologies for the initial misattribution; the ground-truth probes and checksum forensics that led here are documented in the linked repository.

gdevenyi:
Ran this on a Qwen3.8-Flash-Next TP=2 deployment (36 GDN layers on the same `hybrid_radix` / linear-state path). Nothing to report either way, which is worth having on record:

|---|---|---|
| radix-hit probe: a fixed deterministic question re-asked 24 times as a prefix-cache hit while 4 x 256-token generations run concurrently (8 rounds) | 0 / 24 wrong | 0 / 24 wrong |
| single-stream / 8-conc decode | 90.8 / 326.0 tok/s | 89.8 / 321.2 tok/s (noise) |
| TTFT, ~1k prompt | 0.81 s | 0.89 s |

So the corruption did not reproduce here with this probe, and the copy-on-donate clone plus the barriers cost nothing measurable on decode or TTFT. Two differences from your setup that may matter for reproducing: my hit requests are short (one-shot answers, 16 tokens) so the donated slot is rarely still in flight, and the probe alternates only one cached prompt. If you can share the hammer script I can run it on this box, since a second machine confirming the fix would help it land.

---

Merged with conflicts against my image-aware cache key in scheduler/cache.py (three donate sites, mechanical). Two things needed on top:

- the two `torch.cuda.synchronize(self.device)` barriers also fire on a CPU device, which the scheduler unit tests use, so `tests/scheduler` fails without a GPU (`RuntimeError: No CUDA GPUs are available`, 7 tests); I guarded them on `device.type == "cuda"`.
- `test_hybrid_cache_manager_donate_then_hit` and `test_hybrid_finish_donates_live_slot` assert the old ownership transfer (`mamba_value == pp[0]`, ping-pong slot replaced) and need to follow the clone contract.

On this box the corruption does not reproduce with or without the PR: 0 mismatches in 50 hit requests (control, sequential aborts, 4-way concurrent aborts, 5k-token conversations) on a GDN + vision deployment. Server-side decode is unchanged (444-450 tok/s at 8 running, same as without the PR); one GDN slot here is 55 MiB, so the clone per donation is a 55 MiB device copy. Kept in production.

PR #292, opened 2026-08-30
feat: support GLM-5.3-Flash NVFP4

earlvanze:
Adds single-GPU hybrid-offload support for `LibertAIDAI/GLM-5.3-Flash-NVFP4`.

Highlights:
- GLM-5.3-Flash model/config, weight mapping, tool/reasoning parser compatibility, and benchmark support
- NVFP4 expert-bank loader avoids pathological PyTorch intra-op fan-out while restoring the caller thread count on success or failure
- benchmark flags for MoE CPU thread count and KV reserve sizing

Live validation:
- 1x RTX 5090 32 GB, 2 TiB host RAM, dual AMD EPYC 7V12 on Vast (Italy)
- GPU-local NUMA affinity, 48 MoE CPU threads, hybrid backend, CUDA graph, memory ratio 0.95
- 16.84 steady decode tok/s over 126 measured steps; 59.38 ms/token; 29.21 GiB server VRAM; 51 C peak
- longer 254-step validation: 14.73 tok/s; 67.89 ms/token

Checks:
- `uv run pytest tests/models/test_glm5_next_config.py tests/models/test_nvfp4_banks.py -q` (5 passed)
- focused Ruff F/I checks passed
- Python compile checks passed


earlvanze:
Live checkpoint validation found and fixed a KDA prefill layout bug in `c61ef42`.

`causal_conv1d_varlen` returns channel-major contiguous storage. The previous transpose left token-major Q/K/V views with a non-unit feature stride, while the Triton KDA kernel indexes the key/head dimensions as contiguous. Materializing the transposed result restores parity.

Actual GLM-5.3-Flash-NVFP4 layer-0 KDA comparison against the Transformers recurrent reference:

- before: cosine `0.0076478`, max abs diff `0.0086498`
- after: cosine `0.9999971`, max abs diff `1.53e-05`

Regression checks: `7 passed`; Ruff and `git diff --check` clean. A live RTX PRO 6000 WS / FreeToken offload run now produces the exact requested final answer (`PARIS`) and completed c8/c16 demand tests with zero failures.

earlvanze:
Live Vast Serverless qualification completed on commit 3d5354c891a36a3a6e19e1062de194c29a5b3fcb.

- CA/EU fixed-port template with port 3000 published and Vast model-load estimate set to 200 GiB
- Fresh worker cold load: 262.69 seconds
- Cached activation selected FREETOKEN_PROVISION_STAGE=fast_resume and reported ready in about 109 seconds, inside Vasts fixed 300-second starting window
- Semantic request: HTTP 200 in 13.761 seconds, finish_reason=stop, exact visible response GLM53_SERVERLESS_FAST_RESUME_OK
- Vast four-request worker benchmark: 20.572 workload/s
- Targeted Vast Serverless tests: 5 passed; shell syntax and diff checks passed

The worker and endpoint were retained; the inactivity controller is being allowed to park the GPU automatically.

Issue #274, opened 2026-08-29
FrontendAPI] ERROR    Backend supervisor: KeyError: 'model.layers.0.mlp.shared_expert.gate_up_proj.weight_scale'


XeonG:
<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/f278a1cb-46dd-4d34-bd02-9a6137f81849" />

```
[13:20:02] stdout/INFO  checkpoint started (id=Ornith-1.5-35B-A3B-uncensored-NVFP4 pid=14684)
[13:20:05] stdout/INFO  Converting dense weights: 0it [00:00, ?it/s]
[13:20:06] stdout/INFO  Loading weights:   0%|          | 0/3 [00:00<?, ?it/s]
[13:20:07] stdout/INFO  Converting dense weights: 161it [00:01, 170.23it/s]
[13:20:07] stdout/INFO  
[13:20:07] stdout/INFO  Loading weights:  33%|███▎      | 1/3 [00:01<00:03,  1.90s/it]
[13:20:08] stdout/INFO  Converting dense weights: 397it [00:03, 217.22it/s]
[13:20:08] stdout/INFO  
[13:20:09] stdout/INFO  Loading weights:  67%|██████▋   | 2/3 [00:03<00:01,  1.50s/it]
[13:20:09] stdout/INFO  Converting dense weights: 447it [00:04, 99.85it/s]
[13:20:09] stdout/INFO  
[13:20:09] stdout/INFO  Loading weights: 100%|██████████| 3/3 [00:04<00:00,  1.35s/it]
[13:20:09] stdout/INFO  [2026-08-29|13:20:09|core|rank=0] INFO     expert banks: slow path (serial build)
[13:20:10] stdout/INFO  Loading Qwen3.5 NVFP4 experts:   0%|          | 0/3 [00:00<?, ?it/s]
[13:20:10] stdout/INFO  Converting expert banks: 0.00B [00:00, ?B/s]
[13:20:10] stdout/INFO  Converting expert banks: 434MB [00:00, 1.47GB/s]
[13:20:11] stdout/INFO  
[13:20:11] stdout/INFO  Converting expert banks: 867MB [00:00, 876MB/s]
[13:20:12] stdout/INFO  
[13:20:12] stdout/INFO  Converting expert banks: 1.27GB [00:01, 761MB/s]
[13:20:12] stdout/INFO  
[...]

PR #294, opened 2026-08-30
fix(qwen3_5_moe): detect shared-expert quant for modelopt mixed checkpoints

### LINKED ISSUE #164: Qwen3.5 MoE modelopt_mixed loader assumes shared experts use routed-expert NVFP4 quantization
# Qwen3.5 MoE `modelopt_mixed` loader assumes shared experts use routed-expert NVFP4 quantization

## Summary

FreeToken 0.1.2 fails to load `apodex/Apodex-1.1-mini-NVFP4`, a Qwen3.5-MoE checkpoint whose routed experts are NVFP4 while its shared-expert projections are per-tensor FP8.

The Qwen3.5-MoE config currently derives `dense_quant="nvfp4"` from `expert_quant=="nvfp4"`. That allocates NVFP4 shared-expert modules even though the checkpoint's `quantized_layers` map declares FP8 for `*.mlp.shared_expert.{gate,up,down}_proj`. Loading then looks for a fused NVFP4 key that cannot exist.

## Environment

- FreeToken: 0.1.2
- Python: 3.12.3
- PyTorch: 2.11.0+cu130
- Transformers: 5.15.1
- CUDA runtime/toolkit: 13.0
- GPU: NVIDIA GeForce RTX 3080 Ti 12 GB (Ampere)
- Host: Linux x86_64, dual Xeon E5-2682 v4, 110 GiB RAM
- Model: `apodex/Apodex-1.1-mini-NVFP4`
- Architecture: `Qwen3_5MoeForConditionalGeneration`
- Quantization: `modelopt_mixed`; routed experts NVFP4, shared expert/attention FP8 per-tensor, lm_head BF16

## Minimal reproduction

```bash
ft serve \
  --model /path/to/Apodex-1.1-mini-NVFP4 \
  --served-model-name Apodex-1.1-mini-NVFP4 \
  --host 127.0.0.1 --port 1919 \
  --max-running-requests 1 \
  --graph 1 \
  --max-seq-len-override 32768 \
  --num-tokens 32768 \
  --kv-reserve-tokens 32768 \
  --max-prefill-length 2048 \
  --memory-ratio 0.80 \
  --cache-type radix \
  --expert-load auto \
  --moe-backend hybrid \
  --moe-cache-auto \
  --moe-prefill-hit-d2d \
  --moe-cpu-threads 32 \
  --nvfp4-backend auto
```

## Actual behavior

All three mixed-FP8 shards are read, then startup fails in the NVFP4 dense loader:

```text
KeyError: 'model.layers.0.mlp.shared_expert.gate_up_proj.weight'

freetoken/kernel/triton/nvfp4_linear.py:866
```

The checkpoint contains separate FP8 tensors such as:

```text
model.layers.0.mlp.shared_expert.gate_proj.weight
model.layers.0.mlp.shared_expert.up_proj.weight
model.layers.0.mlp.shared_expert.down_proj.weight
```

with their FP8 scale/input-scale tensors; it does not contain an NVFP4 `shared_expert.gate_up_proj.weight`.

## Expected behavior

For `modelopt_mixed`, shared-expert quantization should be detected independently from routed-expert quantization. This checkpoint should instantiate FP8 per-tensor shared-expert modules and fuse the FP8 gate/up pair using the existing per-tensor FP8 fusion path.

## Tested workaround

A local three-part compatibility patch made the model load and serve correctly:

1. Inspect `quantized_layers` for `*.mlp.shared_expert.*` and derive a separate shared-MLP quant mode.
2. Instantiate `Fp8PerTensorColMerged` for `shared_expert.gate_up_proj` and `Fp8PerTensorLinear` for `shared_expert.down_proj` when that mode is FP8 per-tensor.
3. Add this mapping to the existing FP8 gate/up fusion table:

```python
".mlp.shared_expert.gate_up_proj": (
    ".mlp.shared_expert.gate_proj",
    ".mlp.shared_expert.up_proj",
),
```

After the patch:

- 32K startup succeeded and completed 20/20 benchmark requests without service errors.
- A 196,608-token KV configuration started successfully.
- Nine retrieval probes from 8K through 188,423 actual prompt tokens returned exact answers.
- The model ran for more than an hour without loader/runtime errors.

## Suggested fix/tests

- Do not derive shared-expert quantization solely from routed `expert_quant` for `modelopt_mixed` checkpoints.
- Add a loader fixture with NVFP4 routed experts + FP8 shared expert + BF16 lm_head.
- Assert that fused FP8 gate/up weights, per-row scales, and input scales all map correctly.
- Preserve the current pure-NVFP4 fallback for checkpoints without a per-layer quantization map.

### LINKED ISSUE #183: Apodex-1.1-mini-NVFP4 fails: dense_quant hardcoded to NVFP4 when experts are NVFP4, but shared_expert is FP8 (modelopt MIXED_PRECISION)
### Environment
- FreeToken 0.1.1+cu130 (`freetoken-0.1.1+g30aa89115`)
- PyTorch 2.11.0+cu130
- GPU: RTX 4070 Ti 12 GB
- Model: `apodex/Apodex-1.1-mini-NVFP4` (huggingface)
- Conversion: `ft checkpoint --model ... --out ... --dtype bfloat16 --moe-backend offload` succeeds; `ft serve` crashes at weight load

### Model profile
- Architecture: `Qwen3_5MoeForConditionalGeneration` (Qwen3.5 MoE A3B, 256 experts, 40 layers)
- Experts: NVFP4 (the "NVFP4" in the name)
- **Shared expert: FP8 (float8_e4m3fn)**, per `hf_quant_config.json` — every `mlp.shared_expert.{gate,up,down}_proj` is tagged `quant_algo: "FP8"`
- `quantized_layers` covers `linear_attn.*` and `self_attn.{q,k,v,o}_proj` as `FP8` too
- So this is a modelopt `MIXED_PRECISION` checkpoint, not pure NVFP4

### Root cause
In `freetoken/models/qwen3_5_moe/config.py`, line ~194:

```python
dense_quant = "nvfp4" if expert_quant == "nvfp4" else _dense_mlp_quant(hf_config)
```

This unconditionally sets `dense_quant="nvfp4"` whenever experts are NVFP4. For Qwen3.5 MoE NVFP4 checkpoints where the shared_expert is actually FP8 (or bf16), this is wrong. FreeToken then routes the shared_expert through `Nvfp4DenseColMerged` (in `qwen3_5_moe/moe.py`), which calls `nvfp4_linear.py:866` looking for `weight`/`weight_scale`/`weight_global` — keys that don't exist on an FP8 shared_expert (which has `weight` + scalar `weight_scale` + `input_scale`, and no `weight_global`).

Result:
```
KeyError: 'model.layers.0.mlp.shared_expert.gate_up_proj.weight'
  at freetoken/kernel/triton/nvfp4_linear.py:866 in load_state_dict
  called from freetoken/models/qwen3_5_moe/moe.py:Nvfp4DenseColMerged.load_state_dict
```

`_dense_mlp_quant` is documented to *not* match `shared_expert.*` (only bare `.mlp.gate_proj`), so it never overrides the bad default.

### Suggested fix
Either:
1. **Probe `hf_quant_config.json` for the actual shared_expert quant algo** (similar to `_expert_quant`) and use it instead of forcing NVFP4. e.g. add a `_shared_expert_quant(hf_config)` helper that inspects `quantized_layers["...mlp.shared_expert.gate_proj"]`.
2. **Default to `dense_quant="none"`** when `_dense_mlp_quant` returns `"none"` — i.e. remove the unconditional `dense_quant = "nvfp4" if expert_quant == "nvfp4"` line and just use the function's return value. The current comment claims NVFP4 dense weights are "independent of attention quant" but that's only true for *pure NVFP4* checkpoints, not the mixed-precision ones modelopt produces.

Repro:
```bash
# Download (23 GB on D:)
huggingface-cli download apodex/Apodex-1.1-mini-NVFP4 --local-dir D:/freetoken-models/Apodex-1.1-mini-NVFP4

# Convert (works fine)
ft checkpoint --model "D:/freetoken-models/Apodex-1.1-mini-NVFP4" \
              --out "D:/freetoken-models/Apodex-1.1-mini-NVFP4-FTW" \
              --dtype bfloat16 --moe-backend offload --shard-gib 8
# OK: 20.19 GiB FTW written

# Serve (crashes at weight load)
ft serve --model "D:/freetoken-models/Apodex-1.1-mini-NVFP4-FTW" --port 1922 --memory-ratio 0.85
# KeyError: 'model.layers.0.mlp.shared_expert.gate_up_proj.weight'
```

The model itself is fine — `apodex.ai` demo and vLLM/SGLang work; the Apodex README even provides an SGLang launch command using the same `modelopt_mixed` quantization. So this is purely FreeToken's mixed-precision inference path.

JIAQI13:
Fixes #164
Fixes #183
Related: #274 (same shared-expert crash family on another mixed-quant uncensored checkpoint; the dense side routes through #296, but that exact checkpoint is untested)

## Problem

`apodex/Apodex-1.1-mini-NVFP4` (Qwen3.5 MoE A3B, 256 experts) is an nvidia **modelopt `MIXED_PRECISION`** checkpoint: the routed experts are NVFP4, but the **shared expert and attn/GDN projections are per-tensor FP8** (`quant_algo: "FP8"` in `quantized_layers`, fp8-e4m3 weight + scalar `weight_scale` + `input_scale`, no `weight_scale_2`).

`parse_config` assumed the shared expert always matched the routed experts:

```python
dense_quant = "nvfp4" if expert_quant == "nvfp4" else _dense_mlp_quant(hf_config)
```

So `dense_quant` was forced to `"nvfp4"`, the shared expert was built as `Nvfp4DenseColMerged` / `Nvfp4DenseLinear`, and weight loading crashed with

```
KeyError: 'model.layers.0.mlp.shared_expert.gate_up_proj.weight'
  at freetoken/kernel/triton/nvfp4_linear.py:866
```

because an FP8 shared expert has no `weight_scale_2` / `weight_global` buffers. vLLM/SGLang serve the same checkpoint fine, so this is purely the FreeToken mixed-precision path.

## Changes

**`models/qwen3_5_moe/config.py`**
- Add `_shared_expert_quant()`: probe the modelopt `quantized_layers` map for the shared-expert `quant_algo`. Returns `"nvfp4"` when packed FP4 (kept native W4A16) or `"none"` when FP8/bf16 (dequantized to bf16 at load). Pure-NVFP4 checkpoints have no per-layer map and keep the native FP4 default; dense (`num_experts == 0`) checkpoints are unaffected.

**`models/qwen3_5_moe/weight.py`**
- In `_iter_weights_attn_fp8`, the native fp8 (W8A16) path now only applies to attn/GDN projections (`.self_attn.` / `.linear_attn.`), which actually have fp8 linears. Any other per-tensor FP8 weight — the MoE shared expert on mixed checkpoints — falls through to the existing bf16 dequant path, where the already-present shared-expert gate/up fusion (`shared_expert.{gate,up}_proj -> gate_up_proj`) builds the bf16 state dict.

**`tests/models/test_qwen3_5_moe_shared_expert_quant.py`**
- Unit tests for `_shared_expert_quant`: no quant config, NVFP4 without layer map, mixed with FP8 shared expert, mixed with NVFP4 shared expert, and a map without a shared-expert entry. The FP8 case fails on `main` (the detection is absent and the checkpoint crashes at load) and passes with this change.

Pure NVFP4, block-FP8, dense NVFP4, and compressed-tensors NVFP4 checkpoints take unchanged code paths.

## Tested on (real hardware)

- GPU: NVIDIA GeForce RTX 3080 Laptop, 16 GB VRAM (sm_86), driver 610.57.04
- CPU: Intel Core i7-10870H (8 cores), 32 GB RAM + 31 GB swap
- OS: CachyOS (Linux, x86_64)
- FreeToken 0.1.2, torch 2.11.0+cu130 (CUDA 13.0)
- Tested commit: `e6e2b26` on top of `main` (`4b94bdc`)
- Checkpoint: `apodex/Apodex-1.1-mini-NVFP4`
- Command: `ft serve --model-path <Apodex-1.1-mini-NVFP4> --moe-backend offload --moe-cache-auto --port 8899 --host 127.0.0.1`
- `main`: crashes at weight load with the `KeyError` above.
- This branch: model loads, `API server is ready`, `/v1/chat/completions` returns correct generations; decode throughput **44–60 tok/s** in the engine logs (`gen throughput (token/s)` lines); GPU utilization shows the expected MoE-offload bursts (~87–100% during active decode, power up to ~114 W vs ~34 W idle); ~14.6/16 GB VRAM used.
- The new unit tests were run as standalone pytest-style functions (no weights / GPU required).

## Notes

- The FP8 shared expert is dequantized to bf16 at load (W8A16 precision). It is small relative to the 256 routed experts, but a native fp8 shared-expert module would preserve the weight-bandwidth benefit — that is a separate enhancement, not needed to fix the crash.
- This covers the **modelopt** `MIXED_PRECISION` format. A compressed-tensors `mixed-precision` checkpoint (128x128 block-FP8 shared expert + NVFP4 experts, e.g. some Qwen3.6-35B-A3B releases) uses `config_groups` rather than `quantized_layers` and is not handled here.

### DIFF
--- a/python/freetoken/models/qwen3_5_moe/config.py
+++ b/python/freetoken/models/qwen3_5_moe/config.py
@@ -104,6 +104,27 @@ def _dense_mlp_quant(hf_config: Any) -> str:
     return "none"
 
 
+def _shared_expert_quant(hf_config: Any) -> str:
+    """Quant format of the MoE shared expert (``.mlp.shared_expert.gate_proj``).
+
+    Returns ``"nvfp4"`` when it is packed FP4 (kept native W4A16) or ``"none"`` when it
+    is per-tensor FP8 / bf16 (dequantized to bf16 at load). Pure-NVFP4 checkpoints have
+    no per-layer ``quantized_layers`` map, so the shared expert is native FP4; modelopt
+    MIXED_PRECISION checkpoints tag it explicitly (FP8 on Apodex-1.1-mini even though the
+    routed experts are NVFP4)."""
+    get = _quant_accessor(hf_config)
+    if get is None:
+        return "nvfp4"
+    layers = get("quantized_layers") or {}
+    if not isinstance(layers, dict):
+        return "nvfp4"
+    for name, spec in layers.items():
+        if name.endswith(".mlp.shared_expert.gate_proj"):
+            algo = str((spec or {}).get("quant_algo", "")).lower()
+            return "nvfp4" if "fp4" in algo else "none"
+    return "nvfp4"
+
+
 def _attn_quant(hf_config: Any) -> str:
     """Per-tensor FP8 on the *dense* attention/GDN projections. The modelopt
     ``MIXED_PRECISION`` checkpoints tag ``self_attn.{q,k,v,o}_proj`` and
@@ -176,7 +197,22 @@ def parse_config(hf_config: Any) -> ModelConfig:
     # NVFP4. The lm_head is detected separately (only the mixed checkpoint quantizes it).
     # MoE-NVFP4 keeps the shared_expert dense MLP native FP4 (expert_quant=="nvfp4"); a dense
     # (non-MoE) modelopt checkpoint instead tags the bare .mlp.{gate,up,down}_proj as NVFP4.
-    dense_quant = "nvfp4" if expert_quant == "nvfp4" else _dense_mlp_quant(hf_config)
+    if expert_quant == "nvfp4":
+        # In a MoE checkpoint the routed experts and the shared expert are quantized
+        # independently. Most NVFP4 MoE checkpoints keep the shared expert packed FP4
+        # (native W4A16), but modelopt MIXED_PRECISION variants (e.g. Apodex-1.1-mini:
+        # NVFP4 experts + per-tensor FP8 shared expert) do not -- forcing W4A16 there
+        # builds Nvfp4DenseColMerged and then fails looking for weight_scale_2 /
+        # weight_global buffers that do not exist on an FP8 shared expert. Dense
+        # (num_experts == 0) NVFP4 checkpoints keep the .mlp projections packed FP4
+        # like the experts and stay on the native path.
+        dense_quant = (
+            _shared_expert_quant(hf_config)
+            if int(getattr(text, "num_experts", 0) or 0) > 0
+            else "nvfp4"
+        )
+    else:
+        dense_quant = _dense_mlp_quant(hf_config)
     lm_head_quant = _lm_head_quant(hf_config)
 
     # compressed-tensors NVFP4 (dense Qwen3.6-27B): the attention (q/k/v/o, GDN out_proj) AND
--- a/python/freetoken/models/qwen3_5_moe/weight.py
+++ b/python/freetoken/models/qwen3_5_moe/weight.py
@@ -474,7 +474,14 @@ def _iter_weights_attn_fp8(
                     raw_base = raw_name[: -len(".weight")]
                     has_s2 = raw_base + ".weight_scale_2" in keyset
                     has_s = raw_base + ".weight_scale" in keyset
-                    if has_s and not has_s2:  # per-tensor FP8 dense projection
+                    # Native FP8 (W8A16) only for the attn/GDN projections: those have
+                    # fp8 linears in the model. Any other per-tensor FP8 .weight (the
+                    # MoE shared_expert on a mixed checkpoint whose experts are NVFP4,
+                    # e.g. Apodex-1.1-mini) has no fp8 module -- let it fall through to
+                    # bf16 dequant + the shared-expert gate/up fusion below.
+                    if has_s and not has_s2 and (
+                        ".self_attn." in base or ".linear_attn." in base
+                    ):
                         w = f.get_tensor(raw_name)  # fp8-e4m3, kept verbatim
                         sc = f.get_tensor(raw_base + ".weight_scale")
                         # modelopt's calibrated activation scale: kept (not dropped with the
--- a/tests/models/test_qwen3_5_moe_shared_expert_quant.py
+++ b/tests/models/test_qwen3_5_moe_shared_expert_quant.py
@@ -1,0 +1,88 @@
+"""Shared-expert quant detection for MoE checkpoints (issue #183).
+
+modelopt MIXED_PRECISION checkpoints can quantize the routed experts and the
+shared expert independently: Apodex-1.1-mini-NVFP4 ships NVFP4 experts with a
+per-tensor FP8 shared expert. The old code assumed the shared expert matched
+the experts (``dense_quant = "nvfp4"`` whenever experts are NVFP4), built an
+``Nvfp4DenseColMerged`` module, and crashed at load with
+``KeyError: 'model.layers.0.mlp.shared_expert.gate_up_proj.weight'`` because an
+FP8 shared expert has no ``weight_scale_2`` / ``weight_global`` buffers.
+"""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+
+from freetoken.models.qwen3_5_moe.config import _shared_expert_quant
+
+
+def _hf(quantization_config):
+    return SimpleNamespace(quantization_config=quantization_config)
+
+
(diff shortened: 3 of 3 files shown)


XeonG:
is this getting merged?

PR #293, opened 2026-08-30
fix(qwen4_exp): fix nvfp4 ftw weight loading, qsa indexer, and ple tables

3Simplex:
I wanted to post this as a reference of the hacks needed to run oakmindai/Qwen3.8-Flash-Next-NVFP4-FTW which were generated by Gemini 3.7 flash.

### Summary

Fixes weight loading, tensor layout mismatches, and host table initialization when running `Qwen3.8-Flash-Next-NVFP4-FTW` (and `Qwen4Exp` hybrid models) from FTW checkpoints.

### Key Changes

1. **`python/freetoken/models/qwen4_exp/gdn.py`**:
   - Initialized `out_proj` as `LinearReplicated` to match unquantized BF16 weights in NVFP4 checkpoints where linear attention is excluded from quantization (`quantization_config['ignore']`).

2. **`python/freetoken/models/qwen4_exp/attention.py`**:
   - Added flexible `load_state_dict()` in `Qwen4ExpIndexer` to resolve checkpoints storing indexer weights flat under `self_attn.index_*` (e.g. `index_qk_proj.weight`, `index_q_norm.weight`) in addition to nested `self_attn.indexer.*`.

3. **`python/freetoken/engine/engine.py`**:
   - Added dynamic fusion in `_load_weight_state_dict` to concatenate and 16-row pad separate `input_mix_weight_down.weight` and `block_inject_weight.weight` tensors into `input_mix_weight_down_block_inject.weight` when loading FTW checkpoints.

4. **`python/freetoken/models/qwen4_exp/ple.py` & `model.py`**:
   - Added `load_state_dict()` override to `NGramEmbedding` so derived hashing metadata is not strictly expected in dense checkpoint state dicts.
   - Replaced `.copy_()` onto meta tensors with direct CUDA int64 tensor initialization using `torch.cuda.current_device()`.
   - Added safe runtime device migration check in `NGramEmbedding.row_ids()`.

5. **`python/freetoken/models/qwen4_exp/weight.py`**:
   - Added direct binary loader support in `load_ple_table()` for unified `qwen4_ngram.bin` + `qwen4_ngram.json` files.
   - Added `FREETOKEN_ZERO_PLE=1` environment variable support allowing systems without 100+ GB of host RAM to attach `ZeroTable`.

---

### Test Environment

- **GPU**: NVIDIA GeForce RTX 5060 Ti (16 GB VRAM)
[...]

Issue #152, opened 2026-08-24
API access security verification


jackrao168:
In the current version, it seems that there is no parameter configuration option for API access security verification (such as: --api-key). We hope to provide it in the future. Thank you.

wuyongan:
+1 急需，WorkBuddy要求必须要有api-key才能添加

PR #295, opened 2026-08-30
Add Docker container capability

recrudesce:
### Summary of Changes

Adds complete Docker and Docker Compose support for deploying FreeToken with GPU acceleration and native Hugging Face model integration.

---

### What's Changed

1. **`Dockerfile`**:
   - Uses `nvidia/cuda:13.3.1-devel-ubuntu26.04` as the base image to provide `nvcc` and CUDA headers needed for FreeToken's C++ extensions and JIT kernel compilation.
   - Sets up build environment paths (`CUDA_HOME`, `CPATH`, `LIBRARY_PATH`, `LD_LIBRARY_PATH`) to ensure compilation headers (e.g. `cuda_runtime_api.h`) are resolved without errors.
   - Manages dependencies using `uv` with a pinned Python 3.11 virtual environment to adhere to PEP 668 and avoid ABI mismatches with PyTorch extension builds.
   - Installs `freetoken[accel]` with pre-installed build toolchain (`torch`, `setuptools`, `wheel`, `ninja`).
   - Sets `ft serve` as the container entrypoint exposing port `1919`.

2. **`docker-compose.yml`**:
   - Configures NVIDIA GPU resource reservations via the NVIDIA Container Toolkit.
   - Sets `ipc: host` to allocate sufficient shared memory for tensor parallelism and KV cache operations.
   - Binds `~/.cache/huggingface` to persist downloaded model weights across container restarts.
   - Parameterizes the model identifier via the `MODEL_NAME` environment variable.

3. **Configuration & Documentation**:
   - Added `DOCKER_README.md` containing end-to-end instructions for building, configuring, running via Compose or CLI, and verifying inference via OpenAI-compatible endpoints.

---

### Verification

- [x] Docker image builds successfully without header or ABI compilation errors.
- [x] NVIDIA GPU passthrough verified inside the container runtime.
- [x] Container boots `ft serve` and pulls model weights directly from Hugging Face into the mounted cache volume.
- [x] Verified OpenAI-compatible completions endpoint on `http://127.0.0.1:1919/v1/chat/completions`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+# Use CUDA devel base image so nvcc and headers are available for kernel compilation
+FROM nvidia/cuda:13.3.1-devel-ubuntu26.04
+
+ENV DEBIAN_FRONTEND=noninteractive \
+    PYTHONUNBUFFERED=1 \
+    CUDA_HOME=/usr/local/cuda \
+    PATH="/opt/venv/bin:/usr/local/cuda/bin:$PATH" \
+    LD_LIBRARY_PATH="/usr/local/cuda/lib64:$LD_LIBRARY_PATH" \
+    CPATH="/usr/local/cuda/include:$CPATH" \
+    LIBRARY_PATH="/usr/local/cuda/lib64:$LIBRARY_PATH"
+
+# Install build prerequisites and git
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    build-essential \
+    ninja-build \
+    git \
+    curl \
+    ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install uv binary
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+
+WORKDIR /app
+
+# Provision Python 3.11/3.12 virtualenv via uv (avoids Python 3.14 ABI incompatibilities)
+RUN uv venv /opt/venv --python 3.11 && \
+    uv pip install --no-cache "setuptools<70.0.0" wheel torch ninja && \
+    uv pip install --no-cache "freetoken[accel]"
+
+EXPOSE 1919
+
+ENTRYPOINT ["ft", "serve"]
+CMD ["--host", "0.0.0.0", "--port", "1919"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,27 @@
+services:
+  freetoken:
+    image: freetoken:latest
+    container_name: freetoken-server
+    ipc: host
+    ports:
+      - "1919:1919"
+    environment:
+      - HF_TOKEN=${HF_TOKEN:-}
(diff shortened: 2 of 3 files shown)


benwilson:
Two install traps worth guarding against in this Dockerfile, both hit on a stock `pytorch/pytorch:2.11.0-cuda13.0-cudnn9-devel` base before we got a working container (your `nvidia/cuda` + `uv` base may already sidestep them, but they cost us two launches):

1. **`python3 -m venv` fails on images whose Debian python has no `ensurepip`** (`The virtual environment was not created successfully because ensurepip is not available`). `python3 -m venv --system-site-packages --without-pip` and then `venv/bin/python -m pip install …` works, because the image's pip is importable through `--system-site-packages` and installs into the venv prefix.
2. **`pip install freetoken[accel]` from git needs `--no-build-isolation`** with `setuptools>=77 wheel ninja` installed first — an isolated build resolves a second torch and the C++ extension links against the wrong one.

With those two, the pytorch devel image works as a second base: driver ≥ 580 / `nvcc` 13.0 / torch 2.11.0+cu130 preflight, then `ft serve` — RTX 3090, Qwen3.8-Flash-Next NVFP4, coherent to 192k tokens (numbers on #293). One more thing the README could say: `ft serve` has no `--api-key` (#152), so a container that publishes 1919 is open — we run a small bearer-check proxy in front of it in the meantime.

recrudesce:
None of these issues exist on this Dockerfile as I'm not using the base image you mention.  I assume if you wanted to deviate from the Dockerfile, then you'll likely hit on problems that are outside of the scope of this PR.

1: This is not a problem in this base image.
2: This is also not a problem in this Dockerfile, as it's specifically installing `setuptools<=70.0.0` https://github.com/FlashML-org/FreeToken/pull/295/changes#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R28

Regarding the `--api-key` comment, that's outside of the scope of this PR also, as FreeToken doesn't support it as a capability.

RealFascinated:
can this please be merged in

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+# Install uv binary
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+
+WORKDIR /app
+
+# Provision Python 3.11/3.12 virtualenv via uv (avoids Python 3.14 ABI incompatibilities)
+RUN uv venv /opt/venv --python 3.11 && \
+    uv pip install --no-cache "setuptools<70.0.0" wheel torch ninja && \
+    uv pip install --no-cache "freetoken[accel]"
```
xashr, line 29: Why not install from [source](https://github.com/FlashML-org/FreeToken/blob/main/docs/install.md)? I would expect the image to have the latest available changes (checked out locally) when building the image.

Issue #280, opened 2026-08-29
Feature Request: opt-in FP8 KV-cache support for long-context MoE inference


mbetrifork:
## Summary

Please consider adding an opt-in FP8 key/value (KV) cache for long-context inference.

Quantizing the KV cache can substantially increase the usable context capacity on
consumer GPUs, especially when serving large MoE models whose weights already use
NVFP4/FP8 formats. This would complement the broader quantized-cache request in
[#262](https://github.com/FlashML-org/FreeToken/issues/262) with a specific FP8
design point that should preserve more quality than lower-bit cache formats.

## Proof of concept

I built and tested an experimental implementation in a public fork:

- Repository: https://github.com/mbetrifork/FreeToken
- Branch: https://github.com/mbetrifork/FreeToken/tree/experimental/fp8-kv-cache
- Commit: https://github.com/mbetrifork/FreeToken/commit/09eb4bf740cb8d3c2d56a73b7aa7bdd26f974832

The proof of concept:

- Adds an opt-in `--kv-cache-dtype fp8_e4m3` option.
- Stores K/V values as E4M3 FP8 with independent FP32 dynamic scales.
- Dequantizes inside the Triton attention kernels.
- Includes KV-pool sizing/accounting and configuration validation.
- Includes CPU/configuration tests and optional CUDA kernel tests.

On an 8 GB RTX 2000 Ada laptop GPU, it was used to serve
`Qwen3.6-35B-A3B-NVFP4` with a 131K-token KV pool and a coding-agent client.
The regular attention KV data is approximately half the BF16 storage, with a
small scale-table overhead.

## Current proof-of-concept limitations

The fork intentionally keeps the first implementation narrow:

- Qwen3.5/Qwen3.6 MoE checkpoints only.
- BF16 compute only.
- Triton attention only.
- Ordinary paged MHA KV storage only.
- Tensor parallel size 1 only.
- NVIDIA compute capability 8.9 (Ada) or newer.
- Qwen's recurrent Gated DeltaNet state remains unquantized.
- Dynamic per-token/per-KV-head scales are used; checkpoint calibration scales
  are not yet supported.

Unsupported combinations fail during configuration validation, and the feature
is disabled by default. The fork is intended as a reproducible demonstration of
the design space; maintainers may prefer a different abstraction or kernel
implementation.

## Possible next steps

- Benchmark BF16 versus FP8 KV memory, prefill/decode throughput, and quality.
[...]

PR #309, opened 2026-08-31
feat(kvcache): add reliable quantized KV cache

### LINKED ISSUE #262: Feature Request: Add quantized cache management (q8_0, q4_0 or even better turbo4, turbo3)
If I understood well the main scope of the project is to allow standard PC with low VRAM to run big models with MoE architecture (the current standard). The idea behind this is to perform all the main computation using only GPU and VRAM. But not using quantized weights and cache the possibility to run effectively big models is still limited. To improve significantly the inference speed it is necessary to use both quantized weights and cache. Suppose for example to run Qwen3.8-Flash-Next, this is a 125B-parameter main model, supplemented by an additional 51B N-gram embeddings (that can be stored in the RAM), with 6B parameters activated per token (only these must be located in VRAM). With q4_0 quantization the VRAM usage of the selected experts will require about 3GB of VRAM. Using a PC with 12GB VRAM, this imply that will be enough space to store about 8GB of cache (good but not enough for coding agent), while using q4_0 quantized cache it will be possible to store in VRAM the equivalent of about 32GB of uncompressed cache. This is the difference between having a toy model and a working model.

plsgivemeachane:
Closes #262. Related to #280. Supersedes #268 while preserving and crediting @fangyuan-3149 for the original work.

## Summary

- Add Q8, FP8, Q6 and Q4 KV-cache storage.
- Wire `--kv-cache-dtype` through serving and pool allocation.
- Correct KV memory accounting.
- Fix Triton scale/layout handoff, K/V strides and packed offsets.
- Add storage and attention regression tests.

The Q8 corruption occurred because attention read raw int8 values without their dequantization scales.

## Validation

- RTX 5060 8 GB, Ubuntu 26.04 LTS: Q8 serving produces coherent output after the fix.
- Focused CPU/CUDA suite: `92 passed, 1 skipped`.
- Q8/Q4/Q6 Triton store/load parity against the PyTorch reference.
- Decode, extend and paged-attention handoff coverage.
- `git diff --check` passes.
- Original Q4/Q6 hardware results and authorship retained from #268.

The exact launch command and checkpoint used for the RTX 5060 smoke test were not recorded.

### DIFF
--- a/docs/kv_cache_branch_delta.md
+++ b/docs/kv_cache_branch_delta.md
@@ -1,0 +1,97 @@
+# Quantized KV-cache branch delta
+
+This is a compact recovery map for the KV-cache work on
+`feat/reliable-quantized-kv-cache`, compared with `main`. It is intended to make
+the implementation reconstructable without copying engine source into a second,
+divergent tree.
+
+Reference points when this document was created:
+
+- `main`: `4b94bdc` (`docs: add SECURITY.md`)
+- last known-good quantized-KV implementation: `716c1a2`
+- implementation commits: `c086b28..716c1a2`
+
+Recreate the exact source diff with:
+
+```bash
+git diff main...feat/reliable-quantized-kv-cache -- \
+  python/freetoken/attention/triton.py \
+  python/freetoken/engine/config.py \
+  python/freetoken/engine/engine.py \
+  python/freetoken/kernel/triton/attention.py \
+  python/freetoken/kernel/triton/kv_quant.py \
+  python/freetoken/kvcache \
+  python/freetoken/server/args.py
+```
+
+## What the branch adds
+
+The branch adds one shared `--kv-cache-dtype` setting for both K and V. Its
+values are `auto`, `q8_0`, `fp8_e4m3`, `q4_0`, and `q6_0`. It does **not** add
+independent K/V formats or TurboQuant layouts.
+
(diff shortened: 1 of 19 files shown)


fangyuan-3149:
Thanks for building on this work with the attribution intact — a 262K-context user becoming a contributor is exactly the outcome we hoped for when opening #268.

Status sync for maintainers: the core of both PRs is now the same lineage (this PR's first three commits carry fangyuan-3149's authorship), and the two things this PR added on top were genuinely good catches — the missing scale/layout forwarding in the attention wrapper (the silent Q8 raw-int8 path) and the fp32 sink accumulator. Both are now in #268 with credit (fc0771a, 3096737), alongside bd37c60 for the sizing issue you reported there.

What #268 additionally carries at this point: a six-scheme accuracy/speed ladder (q2_lm → q8_0, same protocol, greedy), needle-in-haystack from 8K through 220K, 220K–300K context measurements on an 8 GB consumer GPU, Windows cold-start fixes, and the mag=8 rounding tweak (~5% rel_err over the GGUF clamp). Given the overlap, running two PRs in parallel probably serves nobody — we're happy either way: stack your remaining deltas (the test names are nicer than ours in places) on #268 as follow-up commits, or let the maintainers pick a direction. Thanks again for the energy here.

YigitOker:
Second this

Issue #303, opened 2026-08-30
fix(engine): account for attention workspace in cache-pool sizing


ralfkuh-lab:
## Summary

On a 6 GB card, `ft serve` with the auto-selected FlashInfer attention
backend can fail at startup with a CUDA OOM for FlashInfer's 256 MiB
`float_workspace_buffer` (`attention/fi.py`): the cache-budget planner
(`engine/cache_budget.py`) sizes the MoE/KV pools from `--memory-ratio`
without explicitly charging the attention backend's known fixed workspace,
so at higher ratios the workspace no longer fits into what is left. The
failure is clean (worker exits, API server shuts down), but the knob that
caused it is not obvious from the error.

Related: #110 discusses request-time OOM handling; this report is only
about the startup-time budget. At lower ratio settings that start
successfully, the same headroom must still cover warmup/graph-capture
allocations.

## Environment

- FreeToken 0.1.2, source install; upstream base `4b94bdc`, run at local HEAD `bd7482d` (local patches touch the server layer only: an opt-in distributed-port override, unset here, and a thinking override, not passed here — neither is involved in engine memory planning). torch 2.11.0+cu130, CUDA toolkit 13.0.2, driver 595.84
- RTX 3060 Laptop 6 GB (sm_86), Ryzen 9 5900HX, 30 GB RAM, Linux Mint 22.3
- Checkpoint: `nvidia/Qwen3.6-35B-A3B-NVFP4`

## Reproduce

```bash
ft serve --model /home/user/models/Qwen3.6-35B-A3B-NVFP4 --memory-ratio 0.95 --kv-reserve-tokens 4096
```

```
ERROR Backend supervisor: OutOfMemoryError: CUDA out of memory. Tried to allocate
256.00 MiB. GPU 0 has a total capacity of 5.67 GiB of which 238.12 MiB is free. …
  File ".../freetoken/attention/fi.py", line 114, in __init__
    self.float_workspace_buffer = torch.empty(
```

With `--memory-ratio 0.92` and explicit pool sizing (the reference
configuration below) the same checkpoint starts and serves fine on this
card.

<details><summary>Full log, --memory-ratio 0.95 run</summary>

```
[2026-08-31|00:41:26] INFO     Parsed arguments:
[...]

PR #311, opened 2026-08-31
feat(qwen4_exp): stream the PLE n-gram table from disk

jason-fxz:
## What

Qwen3.8-Flash-Next has a huge PLE n-gram table (47.7 GiB as fp8 in the -FP8/-NVFP4 checkpoints). Instead of preloading it into pinned host RAM, this PR reads the needed rows from disk on each forward. `--ple-backend disk` is the new default; `pinned` keeps the old behavior.

## Data path

`checkpoint shards → TableFile → BatchReader → dedup → pinned staging → GPU`

1. Rows are read in place from the checkpoint's fp8 safetensors shards; an extent table maps row id → (file, offset). No copy, no conversion.
2. TableFile: one O_DIRECT fd per file.
3. BatchReader: each fill becomes one batched read round — io_uring at constant QD64.
4. Duplicate rows in a fill are read once and copied to every destination. No RAM cache: an on/off A/B showed zero decode difference.
5. Rows land in fixed pinned staging; in the CUDA graph, lookup is just an H2D copy + fp8→bf16 dequant.

Row ids are hashed on the host from the request's token history, so prefix hits, restores and forks need no bookkeeping. The C++ store is stateless: `stage(token run)` + `flush(signal)`.

## Sync

Fast path: the decode graph launches first and waits on a `cuStreamWaitValue64` flag right before consuming the rows; the host fills staging while the GPU runs embedding + layer 0, then sets the flag. Fallback: fill before launch. The engine hook is one context manager on the model (`forward_host_ctx`), a no-op for other models.

## Decode / prefill costs

- Decode: one round of ≤16 reads per lane, ~0.1–0.4 ms, hidden behind embedding + layer 0. Disk metric: 4K random-read latency.
- Prefill: all chunk rows in one pipelined round; ≈0.05–0.16 ms per token of TTFT. Disk metric: random-read IOPS.

## Performance

H100 (80 GB), -NVFP4 checkpoint. Decode: AIME problems. Prefill: real mixed text (tech report, blog, distinct GSM8K problems).

| | decode tok/s | prefill TTFT 1K tok | 4K | 16K | 32K |
|---|---|---|---|---|---|
| pinned | 108.23 | 1.33 s | 1.90 s | 8.21 s | 15.14 s |
| **disk** | **105.37 (−2.6%)** | 1.46 s | 2.38 s | 9.39 s | 16.87 s |

RTX PRO 6000 Blackwell (sm_120), VRAM held to 32 GiB to model a consumer card, bare NVMe. Same workloads, except the prefill filler is unique tokens (rows dedup 1.00:1, the worst case); the MoE cache holds ~11% of the experts here, so every step already waits on expert traffic and the round trip hides completely.

| | decode tok/s | prefill TTFT 1K tok | 4K | 16K | 32K |
|---|---|---|---|---|---|
| pinned | 66.08 | 1.25 s | 1.28 s | 4.37 s | 8.77 s |
| **disk** | **66.18 (+0.2%)** | 1.31 s | 1.49 s | 5.01 s | 9.95 s |

## Fallbacks

O_DIRECT → buffered; io_uring (Linux ≥ 5.6) → 16-thread pread pool; stream memops → launch gating; `disk` → `pinned`. Windows is planned — the OS-specific parts sit behind these seams.

### DIFF
--- a/python/freetoken/engine/config.py
+++ b/python/freetoken/engine/config.py
@@ -23,6 +23,8 @@ class EngineConfig:
     moe_backend: str = "auto"
     # NVFP4 routed-expert GEMM backend (--nvfp4-backend): auto|marlin|flashinfer|triton.
     nvfp4_backend: str = "triton"
+    # PLE table backend: "disk" (default) reads rows from the checkpoint files per fill, "pinned" preloads the table into page-locked host RAM.
+    ple_backend: str = "disk"
     # Expert-bank host load (--expert-load): auto|serial|parallel. "auto" reads scattered
     # experts in parallel but falls back to serial when free RAM can't cover the banks + the
     # parallel reader's extra (non-reclaimable) whole-shard buffer; "serial" forces the
--- a/python/freetoken/engine/engine.py
+++ b/python/freetoken/engine/engine.py
@@ -917,11 +917,9 @@ class Engine:
 
     def forward_batch(self, batch: Batch, args: BatchSamplingArgs) -> ForwardOutput:
         assert torch.cuda.current_stream() == self.stream
-        with self.ctx.forward_batch(batch):
-            if self.graph_runner.can_use_cuda_graph(batch):
-                logits = self.graph_runner.replay(batch)
-            else:
-                logits = self.model.forward()
+        use_graph = self.graph_runner.can_use_cuda_graph(batch)
+        with self.ctx.forward_batch(batch), self.model.forward_host_ctx(batch, use_graph):
+            logits = self.graph_runner.replay(batch) if use_graph else self.model.forward()
         if self.cpu_moe_executor is not None:
             # One pinned read: surfaces a fired flag-handshake watchdog (dead coordinator
             # -> stale expert outputs) as a loud error instead of silent corruption.
--- a/python/freetoken/kernel/csrc/ple_store/ple_store_ext.cpp
+++ b/python/freetoken/kernel/csrc/ple_store/ple_store_ext.cpp
@@ -1,0 +1,589 @@
+// Disk-backed PLE row store: rows read straight from the checkpoint's fp8 shard tensors
+// through an extent table (PleRowSource in ple_ssd.py). Engine-thread only, no locks.
+// Duplicate rows in one fill dedup into ONE batched read round; no RAM cache and no
+// per-sequence state. Hash reference: tests/models/qwen4_exp/test_ple_disk.py.
+// Platform seams: TableFile (O_DIRECT+pread; Win: NO_BUFFERING), BatchReader (io_uring,
+// pread-pool fallback = the portable shape), cumemop_* (dlopen libcuda; Win: nvcuda).
+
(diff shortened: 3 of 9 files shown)


andy-yang-1:
@jason-fxz Great work! On single RTX 5090:

| Metric | disk | pinned | disk vs. pinned |
|---|---:|---:|---:|
| Decode (AIME, 12K tokens) | 68.3 tok/s | 68.7 tok/s | -0.5% |
| TTFT (4K prompt) | 1.30 s | 1.27 s | +2% |
| TTFT (16K prompt) | 4.20 s | 3.65 s | +15% |
| Resident Memory | 74 GB | 120 GB | 46 GB less |
| Load Time | 28 s | 68 s | 40 s faster |

chbornman:
Was excited to see this PR so I tested on my box:

**Hardware:** RTX 5080 16 GB, Ryzen 9 9900X, 96 GB DDR5 (dual channel), checkpoint on a gen5 NVMe (Samsung 9100 PRO, ext4). Arch, driver 610.57, CUDA 13.3, this branch @ ccddade, `RadixArk/Qwen3.8-Flash-Next-NVFP4`.

**Config that's serving right now, with an 80K window:**

```
PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True \
ft serve --model RadixArk/Qwen3.8-Flash-Next-NVFP4 \
  --ple-backend disk --expert-load serial \
  --attention-backend qsa_sparse --nvfp4-backend triton \
  --moe-backend offload --page-size 64 --max-running-requests 1 \
  --memory-ratio 0.94 --max-seq-len-override 81920 --kv-reserve-tokens 81920 \
  --moe-cache-size 512 --disable-moe-prefill-overlap --max-prefill-length 2048
```

**Numbers:**

- KV pool allocates **81,472 tokens (1.92 GiB)**, 0.73 GiB free after CUDA graphs
- Decode: **~30 tok/s**, and it stays there deep into context (29.6 tok/s at 72K)
- Fed it a 72,628-token prompt: prefill in 91 s (797 tok/s), and it correctly pulled a fact buried ~52K tokens deep
- Small-context configs decode faster if you don't need the window: 36 tok/s at 4K with the cache on auto
- Steady state host RAM: ~7 GiB available of 96 GB. 

Running Pi agent at 32-35 tok/s with basic computer use tests. 

Two things I hit on the way that might save others time:

1. `--disable-moe-prefill-overlap` is the unlock for big KV on 16 GB: the 1024-slot cache floor is really 2x512 double-buffering, so this frees ~1.3 GiB (+50K tokens of KV) and cost me 2-4 tok/s of decode, since decode at this cache size is PCIe-bound anyway.
2. At `--memory-ratio 0.95` the first real multi-thousand-token prompt OOMs in `gdn.py _conv_prefill` (160 MiB alloc, 140 MiB free) and kills the worker. But 0.94 + `--max-prefill-length 2048` has been rock solid (probably issue #303).

Happy to run more tests on this config if useful.

PR #337, opened 2026-09-02
feat(moe): NVMe disk tier for MoE expert banks

CraigStone-Dev:
## What

Adds an optional **NVMe disk tier for MoE expert banks** (NVFP4 checkpoints). When the expert banks don't fit in host RAM, the tail of the bank stays on NVMe and is fetched on demand, so models whose expert weights exceed RAM become runnable.

```
freetoken serve <ckpt> --moe-disk-tier on --expert-ram-experts 224
```

- `--expert-ram-experts N` (per layer): first N experts resident in RAM (pinned), the rest on disk in the original checkpoint.
- Default (flag off) is byte-identical behavior to before — the disk tier is a no-op unless enabled.
- All NVFP4 families (qwen3_5_moe, qwen4_exp, glm4_moe, glm5_next, gemma4, minimax_m2, minimax_m3); the tier speaks the native NVFP4 (triton) bank layout, and a checkpoint whose provider returns no disk index fails loudly instead of serving zeroed experts.
- Rebased onto `main` (2026-09-10) on top of the quantization refactor (#418/#426/#427): the index now resolves each family's `nvfp4_expert_spec` hook — the same spec the reader uses — so index and loader read the same rows.

## How

- `moe/disk_tier.py` (new): `Nvfp4DiskIndex` — a row-indexed view over the checkpoint's safetensors expert tensors (no copy of the weights), plus `DiskTier`, a pool of O_DIRECT fetch workers that copy disk-resident slot-cache misses into a small pinned staging ring and H2D them into the slot the LRU kernel already assigned, shrinking the miss list so the existing PCIe `copy_missing` path only moves the RAM-resident misses.
- `moe/host_banks.py`: banks are `MAP_PRIVATE|MAP_ANONYMOUS` (a shared anonymous mmap would never return released pages); `pin_prefix(K)` pins only the RAM-resident rows; `release_range` is one `madvise(MADV_DONTNEED)` that really frees; a `mincore` startup check (`check_tail_unbacked`) warns if anything backed the released tail.
- `moe/expert_banks.py` / `expert_pieces.py` / `models/nvfp4_banks.py`: load-time RAM peak is the K/E prefix, not the full expert set — the NVFP4 reader never reads rows `[K, E)` (serial: no `get_tensor`; parallel: filtered at the reader) and yields an empty piece per skipped expert so `build_expert_banks` completes the layer, pins only the prefix, and releases the tail next to where the index is built.
- `moe/offload_cache.py` / `layers/moe.py`: prefill materializes the RAM prefix into identity slots and fetches the routed disk experts; decode fetches disk misses before the PCIe copy. `FT_DISK_TIER_VERIFY=1` gives slot-by-slot byte verification against the checkpoint (all verify probes are gated on the tier and skip CUDA-graph capture).
- `engine/` + `server/args.py`: config plumbing, all v0 preconditions collected into one boot error (`--moe-strategy offload`, `--disable-moe-prefill-overlap`, `--cuda-graph-max-bs 0`, `0 < N < num_experts`).
- `tests/moe/test_disk_tier.py`, `test_disk_tier_families.py` (new) + `test_offload.py`: CPU tests for the index, fetch path, release/tail invariants, the per-family spec hook, and the verify-probe gate.

## Validation (2026-09-10, rebased head)

Rudi: 2× RTX PRO 4000 Blackwell, 62 GB RAM, `shmem_enabled=always`, `FT_DISK_TIER_VERIFY=1`:

| Model (family) | K | Result |
|---|---|---|
| Qwen3.8-Flash-Next-NVFP4 (qwen4_exp, E=512) | 224 | slot verifies 0 mismatches, tail `0 MiB of 36551 MiB` |
| Gemma-4-26B-A4B-NVFP4 (gemma4, E=128) | 64 | 450/450 slot verifies, cgroup 9.37 GiB anon / 2.29 GiB shmem, correct answers |
| GLM-5.3-Flash-NVFP4 (glm5_next, E=288) | — | index byte-for-byte vs checkpoint (18/18 segments); load-verified only (24 GB GPU can't serve it) |

CPU suite: `test_disk_tier.py` 8/8, `test_disk_tier_families.py` 15/15, `test_offload.py` green.

## Notes

- v0 scope: native NVFP4 (triton) layout, `decode_target == "gpu"`, no prefill overlap, no CUDA graphs, original (non-FTW) checkpoints.
- Keep `N × (smallest bank row bytes)` a multiple of 4096 (noted in `--help`) or the small scale banks' tails stay resident (warns, does not abort).

### DIFF
--- a/python/freetoken/engine/config.py
+++ b/python/freetoken/engine/config.py
@@ -49,6 +49,13 @@ class EngineConfig:
     # CPU MoE backend (--moe-strategy cpu): number of CPU worker threads computing
     # the decode experts. 0 = auto (physical cores). Ignored by other backends.
     moe_cpu_threads: int = 0
+    # Disk tier (--moe-disk-tier, see moe/disk_tier.py): "off" = classic behavior.
+    # When "on", experts [0, expert_ram_experts) per layer stay pinned in RAM and the
+    # rest are fetched from the original checkpoint on slot-cache miss. Requires the
+    # native NVFP4 layout, gpu decode target, no prefill overlap, no cuda graphs.
+    moe_disk_tier: str = "off"
+    expert_ram_experts: int = 0
+    disk_fetch_workers: int = 8
     # Hybrid CPU/GPU decode (--moe-strategy offload only): which MoE layers decode on
     # the CPU executor instead of the GPU offload/PCIe path. Spec is an explicit id
     # list ("3,7,11"), a count ("8" -> 8 layers evenly strided across depth), or a
--- a/python/freetoken/engine/engine.py
+++ b/python/freetoken/engine/engine.py
@@ -544,6 +544,29 @@ class Engine:
                 "(locked layers prefill via synchronous pageable copies)"
             )
             object.__setattr__(config, "moe_prefill_overlap", False)
+        disk_tier = None
+        if config.moe_disk_tier == "on":
+            from freetoken.moe.disk_tier import DiskTierSpec
+
+            E = config.model_config.num_experts
+            # Collect ALL unmet preconditions and raise once: each used to surface as a
+            # separate boot-time ValueError, costing a full boot per missing flag.
+            problems = []
+            if not 0 < config.expert_ram_experts < E:
+                problems.append(
+                    f"--expert-ram-experts must be in (0, {E}) with --moe-disk-tier on")
+            if decode_target != "gpu":
+                problems.append(
+                    "--moe-disk-tier v0 requires the gpu decode path (--moe-strategy offload)")
+            if config.moe_prefill_overlap:
+                problems.append("--moe-disk-tier v0 requires --disable-moe-prefill-overlap")
+            if config.cuda_graph_max_bs is None or config.cuda_graph_max_bs >= 1:
+                problems.append(
+                    "--moe-disk-tier v0 requires --cuda-graph-max-bs 0 (cuda graphs disabled)")
+            if problems:
+                raise ValueError(
+                    "--moe-disk-tier on: unmet preconditions:\n  - " + "\n  - ".join(problems))
+            disk_tier = DiskTierSpec(ram_experts=config.expert_ram_experts)
         # Fast path: an FTW checkpoint loads its repacked banks directly.
         # Slow path: load_expert_banks auto-picks parallel vs serial baseline by
         # expert-tensor granularity. Both pin-after-fill.
@@ -570,6 +593,7 @@ class Engine:
                 parallel=expert_parallel,
                 decode_target=("cpu" if decode_target in ("cpu", "hybrid") else "gpu"),
                 layer_residency=requested_residency,
+                disk_tier=disk_tier,
             )
         except PinFailed as exc:
             raise RuntimeError(f"{exc}; {_pin_hint(self._host_tables_bytes)}") from exc
@@ -619,6 +643,21 @@ class Engine:
         # before set_bank_sources: the residency validation and the copy plan's skip of non-pinned layers key on the CPU-layer set
         cache.cpu_layer_ids = cpu_layer_ids
         cache.set_bank_sources(banks.sources, layer_residency=banks.layer_residency)
+        if banks.disk_index is not None:
+            cache.attach_disk_tier(
+                banks.disk_index, banks.disk_ram_experts,
+                workers=config.disk_fetch_workers)
+            logger.info_rank0(
+                f"disk tier: {banks.disk_ram_experts}/{config.model_config.num_experts} "
+                f"experts/layer pinned in RAM; the rest fetched from "
+                f"{config.model_path} on slot-cache miss")
+        elif disk_tier is not None:
+            # The loader released experts [K, E) but no fetcher came back: serving would
+            # multiply by zeroed rows and log nothing. Fail where the flag was set.
+            raise NotImplementedError(
+                "--moe-disk-tier on: this checkpoint's expert provider returned no disk "
+                "index, so experts released at load would never be refetched "
+                f"(quant_format={banks.quant_format!r})")
         cache.set_alphas(banks.gate_up_alpha, banks.down_alpha)
         if decode_target == "hybrid":
             self._resolve_hybrid_fetch(config, cache)
--- a/python/freetoken/layers/moe.py
+++ b/python/freetoken/layers/moe.py
@@ -278,6 +278,9 @@ class OffloadMoELayer(MoELayer):
             return self._decode_hybrid(cache, hidden_states, topk_weights, topk_ids)
         cache.ensure_experts(self.layer_id, topk_ids)
         cache.copy_missing()
+        if (cache.disk_tier_enabled and self.layer_id == 0
+                and os.environ.get("FT_DISK_TIER_VERIFY")):
+            cache._disk_tier.verify_decode_mapping(cache, self.layer_id, topk_ids)
         return self._expert_gemm(
             cache,
             hidden_states,
@@ -364,7 +367,12 @@ class OffloadMoELayer(MoELayer):
             )
             cache.release_prefill_layer(self.layer_id)
             return out
-        cache.materialize_layer(self.layer_id)
+        if cache.disk_tier_enabled:
+            # Disk tier: stream only the RAM-resident prefix + fetch the routed
+            # disk-resident experts (identity slots, so topk_ids pass through).
+            cache.materialize_layer(self.layer_id, topk_ids)
+        else:
+            cache.materialize_layer(self.layer_id)
         cache.copy_missing()
         return self._expert_gemm(
             cache,
--- a/python/freetoken/models/nvfp4_banks.py
+++ b/python/freetoken/models/nvfp4_banks.py
@@ -115,7 +115,7 @@ def iter_nvfp4_expert_pieces(
         if kind not in ("weight", "weight_scale", "weight_scale_2"):
             raise ValueError(f"{spec.desc}: unknown NVFP4 expert tensor kind {kind!r}")
         wanted[name] = (bank_layer, int(match.group("expert")), spec.proj_to_role[proj] + _kind_suffix(kind))
-    experts = config.num_experts - (skip_experts_from or 0)
+    experts = min(skip_experts_from, config.num_experts) if skip_experts_from is not None else config.num_experts
     expected = _num_moe_layers(config) * experts * 9
     if len(wanted) != expected:
         raise ValueError(f"{spec.desc}: found {len(wanted)} expert tensors, expected {expected}")
--- a/python/freetoken/models/nvfp4_banks.py
+++ b/python/freetoken/models/nvfp4_banks.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import collections
+import itertools
 import json
 import os
 import re
@@ -74,6 +75,7 @@ def iter_nvfp4_expert_pieces(
     chunk: int = 8 << 20,
     drop_page_cache: DropPageCache | None = None,
     primary: bool = True,
+    skip_experts_from: int | None = None,
 ):
     """One piece per routed expert: ``gate`` / ``up`` / ``down`` codes plus their ``_scale``
     (fp8 block scales) and ``_global`` (the per-tensor scale, reciprocal for quant-side dialects,
@@ -81,6 +83,12 @@ def iter_nvfp4_expert_pieces(
 
     Serial reads walk the shards in order; ``parallel`` uses the chunked O_DIRECT reader. Either
     way tensors of one expert may span shards, so they are grouped by (layer, expert) as they land.
+
+    ``skip_experts_from`` (the disk tier): experts ``[skip_experts_from, E)`` are disk-resident.
+    The serial reader never calls ``get_tensor`` for them (no I/O); the parallel reader filters
+    them at the reader (the whole-shard read is unchanged, but no per-expert work happens). Each
+    skipped expert still yields an EMPTY piece so the bank fill completes the layer without
+    touching the (released) tail rows.
     """
     from freetoken.models.loader import drop_page_cache as _drop
     from freetoken.moe.expert_pieces import per_expert_pieces
@@ -95,6 +103,8 @@ def iter_nvfp4_expert_pieces(
         match = spec.key_pattern.match(name)
         if match is None:
             continue
+        if skip_experts_from is not None and int(match.group("expert")) >= skip_experts_from:
+            continue  # disk-resident: never read
         bank_layer = _bank_layer(spec, int(match.group("layer")), config)
         if bank_layer is None:
             continue
@@ -105,7 +115,8 @@ def iter_nvfp4_expert_pieces(
         if kind not in ("weight", "weight_scale", "weight_scale_2"):
             raise ValueError(f"{spec.desc}: unknown NVFP4 expert tensor kind {kind!r}")
         wanted[name] = (bank_layer, int(match.group("expert")), spec.proj_to_role[proj] + _kind_suffix(kind))
-    expected = _num_moe_layers(config) * config.num_experts * 9
+    experts = config.num_experts - (skip_experts_from or 0)
+    expected = _num_moe_layers(config) * experts * 9
     if len(wanted) != expected:
         raise ValueError(f"{spec.desc}: found {len(wanted)} expert tensors, expected {expected}")
 
@@ -133,7 +144,19 @@ def iter_nvfp4_expert_pieces(
                 tensor = _ingest_global(spec, tensor)
             yield name, tensor
 
-    return per_expert_pieces(_parallel() if parallel else _serial(), wanted.get, tensors_per_expert=9)
+    pieces = per_expert_pieces(_parallel() if parallel else _serial(), wanted.get, tensors_per_expert=9)
+    if skip_experts_from is not None:
+        pieces = itertools.chain(pieces, _disk_skip_pieces(config, skip_experts_from))
+    return pieces
+
+
+def _disk_skip_pieces(config, skip_from: int):
+    """Empty pieces for the disk-resident experts ``[skip_from, E)``: the rows are
+    complete without a checkpoint read (``build_expert_banks`` marks them written,
+    pins only the prefix, and releases the tail)."""
+    for bank_layer in range(_num_moe_layers(config)):
+        for e in range(skip_from, config.num_experts):
+            yield bank_layer, e, e + 1, {}
 
 
 __all__ = ["Nvfp4ExpertSourceSpec", "iter_nvfp4_expert_pieces"]
--- a/python/freetoken/moe/disk_tier.py
+++ b/python/freetoken/moe/disk_tier.py
@@ -1,0 +1,685 @@
+"""Disk tier: NVMe-backed MoE experts (VRAM <- RAM <- NVMe).
+
+Lets the offload backend serve experts that do NOT fit in pinned RAM: the RAM
+bank holds only the first ``ram_experts`` experts per layer (pinned), the rest
+stay on disk in the original checkpoint. When the GPU slot cache misses a
+disk-resident expert, :class:`DiskTier` fetches its rows with O_DIRECT preadv
+into a small pinned staging buffer and H2D-copies them into the slot the LRU
+kernel already assigned, then shrinks the miss list so the existing PCIe
+``copy_missing`` path only moves the RAM-resident misses.
+
+v0 scope (prototype):
+* native NVFP4 layout only (the "triton" backend banks -- what sm_120 picks);
+* ``decode_target == "gpu"`` (offload) only -- the CPU executor reads banks
+  directly and would read released pages;
+* synchronous fetch (the layer waits for its disk misses); no CUDA-graph
+  capture (the miss-list D2H/H2D round trip is host-side and variable);
+* prefill_overlap off (the double-buffer prefill path bypasses the slot cache).
+
+The bank rows are read from the ORIGINAL safetensors shards: every expert
+tensor is a contiguous per-expert tensor, so a bank row is one (or two, for
+the gate|up-fused banks) aligned super-block preads. No FTW conversion needed.
+"""
+
+from __future__ import annotations
+
+import ctypes
+import json
+import os
+import struct
+import threading
+from concurrent.futures import ThreadPoolExecutor
+from dataclasses import dataclass
+
+import torch
+
+from freetoken.moe.host_banks import HostBank
+
+_ALIGN = 4096
+
+
+@dataclass(frozen=True)
+class DiskTierSpec:
+    """Engine -> loader: how many experts per layer stay pinned in RAM.
+
+    Experts ``[0, ram_experts)`` are pinned as usual; ``[ram_experts, E)`` keep
+    their bank rows allocated but their pages are released after load and are
+    served from disk by :class:`DiskTier`."""
+
+    ram_experts: int
+
+
+def release_bank_tails(banks_by_name: dict[str, list[HostBank]], num_experts: int,
+                       ram_experts: int) -> None:
+    """MADV_DONTNEED the unpinned tail rows of every bank layer (post-load).
+
+    The release is an optimization, not an invariant: the tail rows were never
+    written at load, so when a row boundary is not page-aligned (the small scale
+    banks) we warn and skip that bank instead of failing the boot."""
+    _PAGE = 4096
+    for layer_banks in banks_by_name.values():
+        for bank in layer_banks:
+            row_bytes = bank.nbytes // num_experts
+            offset = ram_experts * row_bytes
+            size = bank.nbytes - offset
+            if offset % _PAGE or size % _PAGE:
+                print(f"[disk-tier] WARNING: bank row boundary not page-aligned "
+                      f"(ram_experts={ram_experts}, row_bytes={row_bytes}); skipping "
+                      f"the release for this bank -- the tail rows were never written, "
+                      f"so nothing is lost", flush=True)
+                continue
+            bank.release_range(offset, size)
+
+
+def tail_resident_bytes(bank: HostBank, num_experts: int, ram_experts: int) -> int:
+    """Bytes the kernel currently backs in the released tail rows ``[ram_experts, E)``.
+
+    mincore(2) over the tail's byte range: one syscall, per-page residency.
+    Conservative -- mincore also reports private-anon pages mapped from the
+    shared zero page (a plain READ of a tail row), so this overcounts what
+    actually costs RAM (cgroup memory.stat shmem is the real number).
+    Returns -1 if mincore itself fails."""
+    row_bytes = bank.nbytes // num_experts
+    off = ram_experts * row_bytes
+    size = bank.nbytes - off
+    if size <= 0:
+        return 0
+    _PAGE = 4096
+    vec = (ctypes.c_ubyte * (size // _PAGE))()
+    libc = ctypes.CDLL("libc.so.6", use_errno=True)
+    libc.mincore.argtypes = [ctypes.c_void_p, ctypes.c_size_t, ctypes.POINTER(ctypes.c_ubyte)]
+    if libc.mincore(ctypes.c_void_p(bank.addr + off), size, vec) != 0:
+        return -1
+    return sum(vec) * _PAGE
+
+
+def check_tail_unbacked(banks_by_name: dict[str, list[HostBank]], num_experts: int,
+                        ram_experts: int) -> None:
+    """Startup sanity check for the lazy-tail invariant, right after
+    ``release_bank_tails``: nothing reads or writes the tail rows, so the kernel
+    should be backing ~none of them. The expected worst case is one 2 MiB THP
+    huge page per bank layer (shmem_enabled=always|force can back the
+    prefix/tail boundary as a huge page); more than that means something touched
+    the tail and the disk-tier RAM math no longer holds. Always logs, warns
+    above the bound."""
+    _HUGE = 2 << 20
+    resident = tail = n_banks = 0
+    for layer_banks in banks_by_name.values():
+        for bank in layer_banks:
+            row_bytes = bank.nbytes // num_experts
+            t = bank.nbytes - ram_experts * row_bytes
+            if t <= 0:
+                continue
+            r = tail_resident_bytes(bank, num_experts, ram_experts)
+            if r < 0:
+                continue  # mincore failed: skip rather than warn on our own probe
+            resident += r
+            tail += t
+            n_banks += 1
+    if n_banks == 0:
+        return
+    from freetoken.distributed import try_get_tp_info
+    tp = try_get_tp_info()
+    rank = getattr(tp, "rank", "?")
+    size_ = getattr(tp, "size", "?")
+    bound = n_banks * _HUGE
+    print(f"[disk-tier] tail check rank={rank}/{size_}: resident {resident >> 20} MiB "
+          f"of {tail >> 20} MiB (warn bound {bound >> 20} MiB = 1x2MiB per bank layer)",
+          flush=True)
+    if resident > bound:
+        print(f"[disk-tier] WARNING: tail rows more resident than the THP bound -- "
+              f"something is reading the released rows; the disk-tier RAM math no longer holds",
+              flush=True)
+
+# Native NVFP4 bank order (== _BANK_SCHEMAS["nvfp4"]) and, per bank, the
+# checkpoint segments that make up one expert row: (proj, kind, dst_row_start,
+# dst_row_end). The gate|up-fused banks splice gate rows then up rows on the
+# output-row axis; down banks are a single segment. Row ends are None = rest.
+_NVP4_BANK_SEGS = (
+    (("gate_proj", "weight", 0, None), ("up_proj", "weight", None, None)),
+    (("gate_proj", "weight_scale", 0, None), ("up_proj", "weight_scale", None, None)),
+    (("gate_proj", "weight_scale_2", 0, None), ("up_proj", "weight_scale_2", None, None)),
+    (("down_proj", "weight", 0, None),),
+    (("down_proj", "weight_scale", 0, None),),
+    (("down_proj", "weight_scale_2", 0, None),),
+)
+
+
+
+def _preadv_error(tier, staging, shard_idx: int, off: int, a0: int, slen: int,
+                  direct: bool) -> OSError:
+    vma = "?"
+    try:
+        for line in open("/proc/self/maps"):
+            lo, hi = line.split()[0].split("-")
+            if int(lo, 16) <= staging.addr < int(hi, 16):
+                vma = line.strip()[:120]
+                break
+    except OSError:
+        pass
+    return OSError(
+        f"disk-tier preadv failed: shard={shard_idx} off={off} a0={a0} "
+        f"slen={slen} direct={direct} buf={hex(staging.addr)} "
+        f"staging_size={tier._staging_size} thread={threading.current_thread().name} "
+        f"vma={vma}"
+    )
+
+
+def _read_safetensors_offsets(path: str) -> dict[str, tuple[int, int]]:
+    """{tensor_name: (start, end)} from a shard's safetensors header, as ABSOLUTE
+    file offsets (data_offsets are relative to the data section, i.e. after the
+    8-byte length + header JSON)."""
+    with open(path, "rb") as f:
+        (hlen,) = struct.unpack("<Q", f.read(8))
+        meta = json.loads(f.read(hlen))
+    base = 8 + hlen
+    return {
+        k: (v["data_offsets"][0] + base, v["data_offsets"][1] + base)
+        for k, v in meta.items() if k != "__metadata__"
+    }
+
+
+class Nvfp4DiskIndex:
+    """(bank, layer, expert) -> per-segment (shard_idx, offset, nbytes) locations.
+
+    Built from the original checkpoint: the HF index json (name -> shard) plus
+    each referenced shard's safetensors header (name -> byte range). Expert
+    tensors are per-expert and contiguous, so a row is exactly one byte range
+    per segment.
+    """
+
+    def __init__(self, model_dir: str, config, spec) -> None:
+        from freetoken.models.nvfp4_banks import _num_moe_layers
+        from freetoken.utils.hf import download_hf_weight
+
+        model_dir = download_hf_weight(model_dir)  # hub id -> local cache dir; no-op if local
+        index_path = os.path.join(model_dir, "model.safetensors.index.json")
+        with open(index_path, encoding="utf-8") as f:
+            weight_map = json.load(f)["weight_map"]
+
+        num_layers = _num_moe_layers(config)
+        # (bank_layer, expert, proj, kind) -> (tensor_name, shard)
+        loc: dict[tuple[int, int, str, str], tuple[str, str]] = {}
+        for name, shard in weight_map.items():
+            m = spec.key_pattern.match(name)
+            if m is None:
+                continue
+            bank_layer = spec.layer_to_bank(int(m.group("layer")), config)
+            if bank_layer is None:
+                continue
+            loc[(bank_layer, int(m.group("expert")), m.group("proj"), m.group("kind"))] = (
+                name, shard)
+
+        shards = sorted(set(shard for _, shard in loc.values()))
+        self.shard_paths = [os.path.join(model_dir, s) for s in shards]
+        offsets = {s: _read_safetensors_offsets(os.path.join(model_dir, s)) for s in shards}
+        shard_idx = {s: i for i, s in enumerate(shards)}
+
+        E = config.num_experts
+        seg_size = struct.calcsize("<iqq")  # (shard_idx, offset, nbytes)
+        self.entries: list[list[bytes]] = []  # [bank][layer] -> packed segments per expert
+        for bank_idx in range(len(_NVP4_BANK_SEGS)):
+            per_layer = []
+            for layer in range(num_layers):
+                rows = bytearray()
+                for e in range(E):
+                    for proj, kind, _, _ in _NVP4_BANK_SEGS[bank_idx]:
+                        key = (layer, e, proj, kind)
+                        entry = loc.get(key)
+                        if entry is None:
+                            raise KeyError(
+                                f"disk tier: no {proj}.{kind} tensor for layer {layer} expert {e} "
+                                f"(bank {bank_idx}) in {index_path}"
+                            )
+                        name, shard = entry
+                        start, end = offsets[shard][name]
+                        rows += struct.pack("<iqq", shard_idx[shard], start, end - start)
+                per_layer.append(bytes(rows))
+            self.entries.append(per_layer)
+        self._seg_size = seg_size
+
+    def row_segments(self, bank_idx: int, layer: int, expert: int) -> list[tuple[int, int, int]]:
+        """[(shard_idx, offset, nbytes)] for one expert row, in segment order."""
+        base = expert * self._seg_size * len(_NVP4_BANK_SEGS[bank_idx])
+        raw = self.entries[bank_idx][layer][base:base + self._seg_size * len(_NVP4_BANK_SEGS[bank_idx])]
+        return [
+            struct.unpack_from("<iqq", raw, i * self._seg_size)
+            for i in range(len(_NVP4_BANK_SEGS[bank_idx]))
+        ]
+
+
+class DiskTier:
+    """Runtime fetcher: disk-resident slot-cache misses -> staging -> GPU slot."""
+
+    def __init__(self, index: Nvfp4DiskIndex, cache, ram_experts: int, workers: int = 8) -> None:
+        self._index = index
+        self._ram = ram_experts
+        self._banks = list(cache.banks)  # [(per_layer_host, gpu_cache)] in schema order
+        self._row_bytes = [
+            b[0][0][0].numel() * b[0][0][0].element_size() for b in self._banks
+        ]  # full expert-row bytes per bank (staging must hold the biggest one)
+        # Per-bank destination row slices (gate|up split at the row midpoint).
+        self._dst_slices: list[list[tuple[int, int]]] = []
+        for bank_idx, (host_layer, _gpu) in enumerate(self._banks):
+            row = host_layer[0][0]
+            if len(_NVP4_BANK_SEGS[bank_idx]) == 2:
+                mid = row.shape[0] // 2
+                self._dst_slices.append([(0, mid), (mid, row.shape[0])])
+            else:
+                self._dst_slices.append([(0, row.shape[0])])
+        if os.environ.get("FT_DISK_TIER_VERIFY"):
+            # TP=2 debug: prove per-rank whether the host bank rows are full or
+            # TP-sharded, and that the disk index's full-row segments match them.
+            from freetoken.distributed import try_get_tp_info
+            tp = try_get_tp_info()
+            host_shapes = [tuple(b[0][0][0].shape) for b in self._banks]
+            disk_bytes = [
+                sum(nb for _, _, nb in self._index.row_segments(bi, 0, 0))
+                for bi in range(len(self._banks))
+            ]
+            print(f"[disk-tier-init] tp_rank={getattr(tp, 'rank', '?')}/{getattr(tp, 'size', '?')} "
+                  f"ram={ram_experts} host_row_shapes={host_shapes} "
+                  f"host_row_bytes={self._row_bytes} disk_row_bytes={disk_bytes}", flush=True)
+        max_row = max(self._row_bytes)
+        self._staging_size = ((max_row + _ALIGN - 1) // _ALIGN + 2) * _ALIGN
+        self._staging = threading.local()
+        self._pool = ThreadPoolExecutor(max_workers=workers, thread_name_prefix="disk-tier")
+        self._fd_lock = threading.Lock()
+        self._fds: dict[int, tuple[int, bool]] = {}
+        self._fetches = 0
+        self._fetch_bytes = 0
+        self._decode_verify_steps = 0
+        self._map_verify_steps = 0
+        self._cache = cache
+
+    # ------------------------------------------------------------------ fds
+    def _fd(self, shard_idx: int) -> tuple[int, bool]:
+        """(fd, o_direct) for a shard; O_DIRECT falls back to plain preadv where the
+        filesystem refuses it (tmpfs/overlayfs -- tests)."""
+        ent = self._fds.get(shard_idx)
+        if ent is None:
+            with self._fd_lock:
+                ent = self._fds.get(shard_idx)
+                if ent is None:
+                    path = self._index.shard_paths[shard_idx]
+                    try:
+                        fd = os.open(path, os.O_RDONLY | os.O_DIRECT)
+                        direct = True
+                    except OSError:
+                        fd = os.open(path, os.O_RDONLY)
+                        direct = False
+                    ent = (fd, direct)
+                    self._fds[shard_idx] = ent
+        return ent
+
+    # --------------------------------------------------------------- staging
+    # Staging ring depth per worker thread. A buffer must not be overwritten by the
+    # next preadv until the async H2D copy that read it has finished (pinned-memory
+    # reuse race -- the copy is DMA, still reading host bytes after copy_ returns).
+    # The depth only needs to cover one copy's DMA time in host-side preadv time;
+    # the per-slot CUDA event below makes any shallower lap correct, just slower.
+    _STAGING_RING = 8
+
+    def _staging_ring(self) -> list:
+        ring = getattr(self._staging, "ring", None)
+        if ring is None:
+            # Fresh worker threads default to CUDA device 0, but the rank may
+            # live on another device (TP>1). The H2D copies land on the
+            # destination tensor's device stream, while ev.record() below uses
+            # the thread's CURRENT stream -- without this, the ring's reuse
+            # guard waits on an idle stream and a preadv can overwrite the
+            # buffer mid-DMA (corrupted slot rows on TP=2 rank 1).
+            dev = self._banks[0][1].device
+            if dev.type == "cuda":
+                torch.cuda.set_device(dev)
+            ring = []
+            for _ in range(self._STAGING_RING):
+                buf = HostBank((self._staging_size,), torch.uint8)
+                buf.pin()  # pin once per worker thread
+                ev = torch.cuda.Event() if torch.cuda.is_available() else None
+                if ev is not None:
+                    ev.record()  # start "complete"; re-recorded after each copy
+                ring.append([buf, ev])
+            self._staging.ring = ring
+        return ring
+
+    # ---------------------------------------------------------------- fetch
+    def _fetch_expert(self, layer: int, expert: int, slot: int) -> None:
+        # The server runs under inference_mode; the fetch pool threads do not,
+        # so the H2D writes into the (inference) slot cache need their own scope.
+        with torch.inference_mode():
+            self._fetch_expert_inner(layer, expert, slot)
+
+    def _fetch_expert_inner(self, layer: int, expert: int, slot: int) -> None:
+        ring = self._staging_ring()
+        ri = getattr(self._staging, "ri", 0)
+        for bank_idx, (_host_layer, gpu_cache) in enumerate(self._banks):
+            row = gpu_cache[slot]
+            segs = self._index.row_segments(bank_idx, layer, expert)
+            for (d0, d1), (shard_idx, off, nbytes) in zip(self._dst_slices[bank_idx], segs):
+                staging, ev = ring[ri]
+                if ev is not None:
+                    # This buffer's last async H2D copy must be done before the
+                    # preadv below overwrites it (pinned-memory reuse race).
+                    ev.synchronize()
+                ri = (ri + 1) % len(ring)
+                fd, direct = self._fd(shard_idx)
+                if direct:
+                    a0 = off & ~(_ALIGN - 1)
+                    slen = (off + nbytes - a0 + _ALIGN - 1) & ~(_ALIGN - 1)
+                else:
+                    a0, slen = off, nbytes
+                mv = (ctypes.c_char * slen).from_address(staging.addr)
+                try:
+                    os.preadv(fd, [mv], a0)
+                except OSError:
+                    raise _preadv_error(self, staging, shard_idx, off, a0, slen, direct)
+                row_off = off - a0
+                if bank_idx in (2, 5):
+                    # Global-scale banks: the checkpoint stores a per-expert fp32
+                    # SCALAR (weight_scale_2); the bank row is that value as fp16
+                    # broadcast across the row -- convert + fill, no byte copy.
+                    val = staging.tensor[row_off:row_off + 4].view(torch.float32)[0].to(
+                        torch.float16)
+                    row[d0:d1].fill_(val)
+                    continue
+                src = staging.tensor[row_off:row_off + nbytes]
+                dst = row[d0:d1]
+                dst.copy_(src.view(dst.dtype).view(dst.shape), non_blocking=True)
+                if ev is not None:
+                    # Arm: the next reuse of this buffer waits for this copy.
+                    ev.record()
+        self._staging.ri = ri
+        self._fetches += 1
+        self._fetch_bytes += sum(self._row_bytes)
+
+    def _sync_fetches(self) -> None:
+        """Wait for the pool threads' async H2D copies to land.
+
+        The copies are enqueued on the pool threads' default stream; f.result() only
+        waits for them to be ENQUEUED. The GEMM's stream is not ordered with that
+        stream, so sync the default stream before the GEMM reads the slots."""
+        # Key this on where the BANKS live, not on whether the machine has a GPU: the
+        # disk-tier unit tests build CPU banks on a CUDA box, and default_stream() rejects
+        # a CPU device outright. There is nothing to order for a CPU->CPU copy anyway.
+        device = self._banks[0][1].device
+        if device.type != "cuda":
+            return  # CPU banks: the copies are synchronous CPU->CPU
+        torch.cuda.default_stream(device).synchronize()
+
+    def _verify_slot(self, cache, layer: int, expert: int, slot: int | None = None,
+                     phase: str = "prefill") -> None:
+        """One-shot debug: read back an expert's slot rows and compare against the
+        checkpoint bytes (ground truth). Gated on FT_DISK_TIER_VERIFY."""
+        import torch
+        if slot is None:
+            slot = expert  # identity mapping (prefill)
+        for bank_idx, (_host_layer, gpu_cache) in enumerate(self._banks):
+            slot_row = gpu_cache[slot].contiguous()
+            flat = slot_row.view(torch.uint8).reshape(-1)
+            ref = self._ref_row(bank_idx, layer, expert, flat.numel(),
+                                slot_row.element_size(),
+                                slot_row.numel() // slot_row.shape[0] if slot_row.dim() > 1 else 1)
+            try:
+                ref = ref.to(flat.device)
+                match = bool(torch.equal(flat, ref))
+                if match:
+                    print(f"[verify] {phase} L{layer} bank={bank_idx} expert={expert} "
+                          f"slot={slot} match=True", flush=True)
+                    continue
+                diff = (flat != ref)
+                nz = torch.nonzero(diff).flatten()
+                print(f"[verify] {phase} L{layer} bank={bank_idx} expert={expert} "
+                      f"slot={slot} match=False n_diff={int(diff.sum())}/{flat.numel()} "
+                      f"first_off={int(nz[0])} last_off={int(nz[-1])} "
+                      f"slot_norm={slot_row.float().norm().item():.4f} "
+                      f"ref_norm={ref.view(slot_row.dtype).view(slot_row.shape).float().norm().item():.4f} "
+                      f"slot_head={flat[:8].tolist()} ref_head={ref[:8].tolist()}", flush=True)
+                self._identify_overwriter(layer, bank_idx, expert, flat)
+            except Exception as exc:  # never crash the server in debug
+                print(f"[verify] bank={bank_idx} expert={expert} ERROR {exc!r}", flush=True)
+
+    def _ref_row(self, bank_idx: int, layer: int, expert: int, row_bytes: int,
+                 row_el: int, row_leading: int) -> torch.Tensor:
+        """Reference row bytes for (bank, layer, expert) straight from the checkpoint."""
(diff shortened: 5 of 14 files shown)


MT-z:
Tested this PR head (623ca1d) rebased onto `main` (clean at a80b4d3 and 6eca2d7) on an RTX 4090 24 GB + 61 GB RAM box (Ubuntu, kernel 7.0), with three NVFP4 checkpoints: `RadixArk/Qwen3.8-Flash-Next-NVFP4` (qwen4_exp), `LibertAIDAI/GLM-5.3-Flash-NVFP4` (glm5_next) and `ornith-ai/Ornith-1.5-35B-A3B-NVFP4` (qwen3_5_moe). The tier works end to end once the three fixes below are applied -- Qwen3.8 (63 GiB of experts on a 61 GB box) boots in 23 s at `--expert-ram-experts 224` and `FT_DISK_TIER_VERIFY=1` reports 0 mismatches over ~16k slot checks. Findings in severity order; diffs for 1-3 are folded at the end.

### 1. Load-time RAM peak is the *full* expert set, regardless of `--expert-ram-experts`

`load_nvfp4_expert_source_banks` (and the `_parallel` twin) allocate `[E, ...]` banks, fill **every** expert row from every shard, and only then call `release_bank_tails` (`nvfp4_banks.py:153-154, 214-216` / `285-286, 341-343`). So a model whose experts do not fit in RAM cannot get through the load -- which is the case the PR exists for.

Repro: GLM-5.3-Flash-NVFP4 (288 experts, 166 GiB of expert rows), `--expert-ram-experts 48`, 61 GB RAM. The serial loader ran at ~1.5 it/s to shard 73/118, then fell into swap (19 s/it, then 56 s/it) and had to be killed.

Fix: keep the `[E, ...]` allocation (rows `>= K` are the tier's fetch destination and must exist as address space) but never read or write them at load: the serial loader skips the row before `get_tensor`, the parallel loader filters at the reader so the cold 83% costs no I/O, and `LayerCompletionTracker` / the final `placed == expected` assert count `rows_per_layer` instead of `E`. Same model afterwards: 118/118 shards in 13 s, cgroup `shmem` 28.9 GiB (= 48/288 of 166 GiB), no swap.

### 2. `disk_tier=` only reaches `qwen3_5_moe`; every other NVFP4 family dies at load

`models/weight.py` and `moe/expert_banks.py` now call `load_nvfp4_expert_sources(..., disk_tier=disk_tier)` unconditionally, but only `qwen3_5_moe/weight.py:1081` accepts the keyword. With `--moe-disk-tier on`:

```
TypeError: load_nvfp4_expert_sources() got an unexpected keyword argument 'disk_tier'
```

on gemma4 (`weight.py:278`), glm4_moe (194), glm5_next (98), minimax_m2 (91), minimax_m3 (251), qwen4_exp (292) and their `_parallel` variants -- 13 wrappers. The base `load_nvfp4_expert_source_banks{,_parallel}` already take `disk_tier`; the wrappers just have to forward it (mechanical; verified with `inspect.signature` on all seven families, exercised end to end on glm5_next and qwen4_exp).

### 3. 623ca1d breaks two of the PR's own CPU tests

`tests/moe/test_disk_tier.py::test_fetch_pending_filters_and_rewrites` and `::test_fetch_pending_all_disk_clears_list` fail on the PR head and pass on its parent. `_sync_fetches` (`disk_tier.py:328`) gates the stream sync on `torch.cuda.is_available()`, so on a CUDA machine the CPU-bank tests fall through to `torch.cuda.default_stream(cpu_device)`, which raises. Keying the guard on the banks' device type (`self._banks[0][1].device.type != "cuda"`) fixes both; a CPU->CPU copy has nothing to order anyway.

### 4. Minor

- `disk_tier.py:540`: an unconditional `print("[disk-tier dbg] ...")` for `layer_id < 3` on every prefill.
- `disk_tier.py:544`: `int(cache.step.item())` is a device-to-host sync per layer on the decode path.

### 5. The v0 preconditions surface one boot at a time

`--disable-moe-prefill-overlap`, `--cuda-graph-max-bs 0` and the gpu decode target are each discovered by a separate `ValueError` (`engine.py:578/580/583`), which cost three boots of a 182 GB model here. Validating them together, or listing them in `--help` for `--moe-disk-tier`, would help.

The three fixes are small commits on top of your head; I can push them as a branch or open a PR against yours -- say which you prefer.

<details><summary>Fix 3 -- key the fetch sync on the banks' device (0001)</summary>

```diff
diff --git a/python/freetoken/moe/disk_tier.py b/python/freetoken/moe/disk_tier.py
index 9f16063..b5c7f14 100644
--- a/python/freetoken/moe/disk_tier.py
+++ b/python/freetoken/moe/disk_tier.py
@@ -325,9 +325,13 @@ class DiskTier:
         The copies are enqueued on the pool threads' default stream; f.result() only
         waits for them to be ENQUEUED. The GEMM's stream is not ordered with that
         stream, so sync the default stream before the GEMM reads the slots."""
-        if not torch.cuda.is_available():
-            return  # CPU-only tests: the copies are synchronous CPU->CPU
-        torch.cuda.default_stream(self._banks[0][1].device).synchronize()
+        # Key this on where the BANKS live, not on whether the machine has a GPU: the
+        # disk-tier unit tests build CPU banks on a CUDA box, and default_stream() rejects
+        # a CPU device outright. There is nothing to order for a CPU->CPU copy anyway.
+        device = self._banks[0][1].device
+        if device.type != "cuda":
+            return  # CPU banks: the copies are synchronous CPU->CPU
+        torch.cuda.default_stream(device).synchronize()
 
     def _verify_slot(self, cache, layer: int, expert: int, slot: int | None = None,
                      phase: str = "prefill") -> None:
```
</details>

<details><summary>Fix 1 -- never materialize disk-resident rows at load (0002)</summary>

```diff
diff --git a/python/freetoken/models/nvfp4_banks.py b/python/freetoken/models/nvfp4_banks.py
index 68a5cfb..4386da8 100644
--- a/python/freetoken/models/nvfp4_banks.py
+++ b/python/freetoken/models/nvfp4_banks.py
@@ -152,6 +152,13 @@ def load_nvfp4_expert_source_banks(
 
     _hb = _alloc_nvfp4_host_banks(num_layers, E, H, I)  # unpinned; pinned after fill
     K = disk_tier.ram_experts if disk_tier is not None else None
+    # Rows this loader actually materializes per layer. The banks are still allocated
+    # at the full [E, ...] shape -- rows K..E-1 are the disk tier's fetch destination and
+    # must exist as address space -- but they are never read or written here, so they
+    # stay unbacked. Filling them and releasing afterwards (the previous order) made the
+    # load-time peak the FULL expert set, which is exactly the case the tier exists for:
+    # GLM-5.3-Flash (166 GiB of experts) swapped a 61 GB box to a halt at shard 73/118.
+    rows_per_layer = E if K is None else min(K, E)
     if disk_tier is not None and layer_sink is not None:
         raise NotImplementedError("disk tier: the converter (layer_sink) path is not supported yet")
     gate_up_packed = [b.tensor for b in _hb["gate_up_packed"]]
@@ -164,7 +171,7 @@ def load_nvfp4_expert_source_banks(
     from freetoken.moe.host_banks import LayerCompletionTracker, PinPipeline
 
     def _load(sink) -> int:
-        tracker = LayerCompletionTracker(E * 6, _hb, sink)
+        tracker = LayerCompletionTracker(rows_per_layer * 6, _hb, sink)
         placed = 0
         for shard in tqdm(sorted(weight_shards), desc=f"Loading {spec.desc}", disable=not primary):
             path = os.path.join(folder, shard)
@@ -172,6 +179,8 @@ def load_nvfp4_expert_source_banks(
                 for name, match, bank_layer_id in weight_shards[shard]:
                     layer = int(match.group("layer"))
                     expert = int(match.group("expert"))
+                    if expert >= rows_per_layer:
+                        continue  # disk-resident row: fetched on demand, never loaded here
                     proj = match.group("proj")
                     role = spec.proj_to_role[proj]
                     kind = _canon_kind(spec, match.group("kind"))
@@ -213,7 +222,7 @@ def load_nvfp4_expert_source_banks(
 
             release_bank_tails(_hb, E, K)
 
-    expected = num_layers * E * 6
+    expected = num_layers * rows_per_layer * 6
     assert placed == expected, f"{spec.desc}: loaded {placed} expert tensors, expected {expected}"
     return {
         "gate_up_packed": gate_up_packed,
@@ -284,6 +293,13 @@ def load_nvfp4_expert_source_banks_parallel(
 
     _hb = _alloc_nvfp4_host_banks(num_layers, E, H, I)  # unpinned; pinned after fill
     K = disk_tier.ram_experts if disk_tier is not None else None
+    # Rows this loader actually materializes per layer. The banks are still allocated
+    # at the full [E, ...] shape -- rows K..E-1 are the disk tier's fetch destination and
+    # must exist as address space -- but they are never read or written here, so they
+    # stay unbacked. Filling them and releasing afterwards (the previous order) made the
+    # load-time peak the FULL expert set, which is exactly the case the tier exists for:
+    # GLM-5.3-Flash (166 GiB of experts) swapped a 61 GB box to a halt at shard 73/118.
+    rows_per_layer = E if K is None else min(K, E)
     if disk_tier is not None and layer_sink is not None:
         raise NotImplementedError("disk tier: the converter (layer_sink) path is not supported yet")
     gate_up_packed = [b.tensor for b in _hb["gate_up_packed"]]
@@ -297,10 +313,15 @@ def load_nvfp4_expert_source_banks_parallel(
 
     # Pass 2: bulk weight/weight_scale via the common parallel reader; place by name.
     def _load(sink) -> int:
-        tracker = LayerCompletionTracker(E * 6, _hb, sink)
+        tracker = LayerCompletionTracker(rows_per_layer * 6, _hb, sink)
         placed = 0
+        def _wanted(n: str) -> bool:
+            info = weight_info.get(n)
+            # Filter at the reader so disk-resident rows cost no I/O at all, not just no write.
+            return info is not None and int(info[0].group("expert")) < rows_per_layer
+
         for name, tensor in iter_expert_tensors_parallel(
-            folder, lambda n: n in weight_info, workers=workers, chunk=chunk
+            folder, _wanted, workers=workers, chunk=chunk
         ):
             match, bank_layer_id = weight_info[name]
             layer = int(match.group("layer"))
@@ -340,7 +361,7 @@ def load_nvfp4_expert_source_banks_parallel(
 
             release_bank_tails(_hb, E, K)
 
-    expected = num_layers * E * 6
+    expected = num_layers * rows_per_layer * 6
     assert placed == expected, f"{spec.desc}: loaded {placed} expert tensors, expected {expected}"
     return {
         "gate_up_packed": gate_up_packed,
```
</details>

<details><summary>Fix 2 -- thread disk_tier through every NVFP4 family's loader (0003)</summary>

```diff
diff --git a/python/freetoken/models/gemma4/weight.py b/python/freetoken/models/gemma4/weight.py
index 43c2355..ad9c43e 100644
--- a/python/freetoken/models/gemma4/weight.py
+++ b/python/freetoken/models/gemma4/weight.py
@@ -276,7 +276,7 @@ def iter_weights_parallel(
 
 
 def load_nvfp4_expert_sources(
-    model_path: str, config, *, layer_sink=None
+    model_path: str, config, *, layer_sink=None, disk_tier=None
 ) -> dict[str, list[torch.Tensor]]:
     """CPU NVFP4 expert source banks for the offload cache; see load_nvfp4_expert_source_banks."""
     return load_nvfp4_expert_source_banks(
@@ -286,11 +286,12 @@ def load_nvfp4_expert_sources(
         drop_page_cache=drop_page_cache,
         primary=get_tp_info().is_primary(),
         layer_sink=layer_sink,
+        disk_tier=disk_tier,
     )
 
 
 def load_nvfp4_expert_sources_parallel(
-    model_path: str, config, *, workers: int = 8, chunk: int = 8 << 20, layer_sink=None
+    model_path: str, config, *, workers: int = 8, chunk: int = 8 << 20, layer_sink=None, disk_tier=None
 ):
     """parallel: same NVFP4 source banks via the common chunked multi-threaded reader."""
     from freetoken.models.nvfp4_banks import load_nvfp4_expert_source_banks_parallel
@@ -304,6 +305,7 @@ def load_nvfp4_expert_sources_parallel(
         workers=workers,
         chunk=chunk,
         layer_sink=layer_sink,
+        disk_tier=disk_tier,
     )
 
 
diff --git a/python/freetoken/models/glm4_moe/weight.py b/python/freetoken/models/glm4_moe/weight.py
index 7efc6e6..a00005e 100644
--- a/python/freetoken/models/glm4_moe/weight.py
+++ b/python/freetoken/models/glm4_moe/weight.py
@@ -191,7 +191,7 @@ def _iter_resident_weights(reader, config, primary) -> Iterator[tuple[str, torch
 # --------------------------------------------------------------------------------------
 # Routed expert host banks (NVFP4) for the offload cache.
 # --------------------------------------------------------------------------------------
-def load_nvfp4_expert_sources(model_path: str, config, *, layer_sink=None) -> dict[str, torch.Tensor]:
+def load_nvfp4_expert_sources(model_path: str, config, *, layer_sink=None, disk_tier=None) -> dict[str, torch.Tensor]:
     """Build the pinned CPU NVFP4 banks for GLM-4's routed experts.
 
     experts exist only for layers [first_k_dense_replace, num_layers) and pack by MoE layer
@@ -205,11 +205,12 @@ def load_nvfp4_expert_sources(model_path: str, config, *, layer_sink=None) -> di
         drop_page_cache=drop_page_cache,
         primary=get_tp_info().is_primary(),
         layer_sink=layer_sink,
+        disk_tier=disk_tier,
     )
 
 
 def load_nvfp4_expert_sources_parallel(
-    model_path: str, config, *, workers: int = 8, chunk: int = 8 << 20, layer_sink=None
+    model_path: str, config, *, workers: int = 8, chunk: int = 8 << 20, layer_sink=None, disk_tier=None
 ):
     """parallel: same NVFP4 source banks via the common chunked multi-threaded O_DIRECT reader."""
     from freetoken.models.nvfp4_banks import load_nvfp4_expert_source_banks_parallel
@@ -223,6 +224,7 @@ def load_nvfp4_expert_sources_parallel(
         workers=workers,
         chunk=chunk,
         layer_sink=layer_sink,
+        disk_tier=disk_tier,
     )
 
 
diff --git a/python/freetoken/models/glm5_next/weight.py b/python/freetoken/models/glm5_next/weight.py
index fda8522..2e3b7f6 100644
--- a/python/freetoken/models/glm5_next/weight.py
+++ b/python/freetoken/models/glm5_next/weight.py
@@ -95,7 +95,10 @@ def _select_expert_source_spec(model_path: str) -> Nvfp4ExpertSourceSpec:
 _KDA_IN_PROJ = ("q_proj", "k_proj", "v_proj", "b_proj", "f_a_proj", "g_a_proj")
 
 
-def load_nvfp4_expert_sources(model_path: str, config, layer_sink=None):
+def load_nvfp4_expert_sources(model_path: str, config, *, layer_sink=None, disk_tier=None):
+    # disk_tier is threaded through the same way qwen3_5_moe does it: the caller builds the
+    # NVMe tier and every NVFP4 family's loader has to pass it down, or the tail of the bank
+    # is never registered and load fails with an unexpected-kwarg TypeError.
     return load_nvfp4_expert_source_banks(
         model_path,
         config,
@@ -103,6 +106,7 @@ def load_nvfp4_expert_sources(model_path: str, config, layer_sink=None):
         drop_page_cache=drop_page_cache,
         primary=get_tp_info().is_primary(),
         layer_sink=layer_sink,
+        disk_tier=disk_tier,
     )
 
 
diff --git a/python/freetoken/models/minimax_m2/weight.py b/python/freetoken/models/minimax_m2/weight.py
index 103f8ec..a010225 100644
--- a/python/freetoken/models/minimax_m2/weight.py
+++ b/python/freetoken/models/minimax_m2/weight.py
@@ -92,7 +92,7 @@ def load_nvfp4_expert_sources(
     model_path: str,
     config,
     *,
-    layer_sink=None,
+    layer_sink=None, disk_tier=None,
 ) -> dict[str, torch.Tensor]:
     """CPU NVFP4 expert source banks for the offload cache; see load_nvfp4_expert_source_banks."""
     return load_nvfp4_expert_source_banks(
@@ -102,11 +102,12 @@ def load_nvfp4_expert_sources(
         drop_page_cache=drop_page_cache,
         primary=get_tp_info().is_primary(),
         layer_sink=layer_sink,
+        disk_tier=disk_tier,
     )
 
 
 def load_nvfp4_expert_sources_parallel(
-    model_path: str, config, *, workers: int = 8, chunk: int = 8 << 20, layer_sink=None
+    model_path: str, config, *, workers: int = 8, chunk: int = 8 << 20, layer_sink=None, disk_tier=None
 ):
     """parallel: same NVFP4 source banks via the common chunked multi-threaded O_DIRECT reader."""
     from freetoken.models.nvfp4_banks import load_nvfp4_expert_source_banks_parallel
@@ -120,6 +121,7 @@ def load_nvfp4_expert_sources_parallel(
         workers=workers,
         chunk=chunk,
         layer_sink=layer_sink,
+        disk_tier=disk_tier,
     )
 
 
diff --git a/python/freetoken/models/minimax_m3/weight.py b/python/freetoken/models/minimax_m3/weight.py
index e10a723..bfcfbbd 100644
--- a/python/freetoken/models/minimax_m3/weight.py
+++ b/python/freetoken/models/minimax_m3/weight.py
@@ -249,7 +249,7 @@ def iter_weights(
 
 
 def load_nvfp4_expert_sources(
-    model_path: str, config, *, layer_sink=None
+    model_path: str, config, *, layer_sink=None, disk_tier=None
 ) -> dict[str, list[torch.Tensor]]:
     """CPU NVFP4 expert source banks for the offload cache; see load_nvfp4_expert_source_banks."""
     return load_nvfp4_expert_source_banks(
@@ -259,11 +259,12 @@ def load_nvfp4_expert_sources(
         drop_page_cache=drop_page_cache,
         primary=get_tp_info().is_primary(),
         layer_sink=layer_sink,
+        disk_tier=disk_tier,
     )
 
 
 def load_nvfp4_expert_sources_parallel(
-    model_path: str, config, *, workers: int = 8, chunk: int = 8 << 20, layer_sink=None
+    model_path: str, config, *, workers: int = 8, chunk: int = 8 << 20, layer_sink=None, disk_tier=None
 ):
     """parallel: same NVFP4 source banks via the common chunked multi-threaded O_DIRECT reader."""
     from freetoken.models.nvfp4_banks import load_nvfp4_expert_source_banks_parallel
@@ -277,6 +278,7 @@ def load_nvfp4_expert_sources_parallel(
         workers=workers,
         chunk=chunk,
         layer_sink=layer_sink,
+        disk_tier=disk_tier,
     )
 
 
diff --git a/python/freetoken/models/qwen4_exp/weight.py b/python/freetoken/models/qwen4_exp/weight.py
index f8d2a74..edd38f4 100644
--- a/python/freetoken/models/qwen4_exp/weight.py
+++ b/python/freetoken/models/qwen4_exp/weight.py
@@ -289,7 +289,7 @@ def load_ple_table(model_path: str, qwen4_args, *, pin: bool = True,
 # ======================================================================================
 
 
-def load_nvfp4_expert_sources(model_path: str, config, *, layer_sink=None) -> dict:
+def load_nvfp4_expert_sources(model_path: str, config, *, layer_sink=None, disk_tier=None) -> dict:
     """Build the CPU NVFP4 expert source banks for the offload cache (gate/up fused on the output-row axis, down separate; weight_scale_2 carried as the per-row global scale)."""
     return load_nvfp4_expert_source_banks(
         model_path,
@@ -298,11 +298,12 @@ def load_nvfp4_expert_sources(model_path: str, config, *, layer_sink=None) -> di
         drop_page_cache=drop_page_cache,
         primary=get_tp_info().is_primary(),
         layer_sink=layer_sink,
+        disk_tier=disk_tier,
     )
 
 
 def load_nvfp4_expert_sources_parallel(
-    model_path: str, config, *, workers: int = 8, chunk: int = 8 << 20, layer_sink=None
+    model_path: str, config, *, workers: int = 8, chunk: int = 8 << 20, layer_sink=None, disk_tier=None
 ) -> dict:
     """parallel: same NVFP4 source banks via the common chunked multi-threaded reader."""
     from freetoken.models.nvfp4_banks import load_nvfp4_expert_source_banks_parallel
@@ -316,6 +317,7 @@ def load_nvfp4_expert_sources_parallel(
         workers=workers,
         chunk=chunk,
         layer_sink=layer_sink,
+        disk_tier=disk_tier,
     )
```
</details>

<sub>Written with AI assistance; every number above was measured on my hardware and I can reproduce it.</sub>

CraigStone-Dev:
Thanks for the thorough review — all three fixes are applied verbatim as `de43469` (no PR needed from your side), and the two minor items plus the precondition UX are in as `795659d`:

- **Fix 1** (load-time RAM peak): applied to both serial and parallel loaders.
- **Fix 2** (`disk_tier=` threading): all six remaining families forward it now; verified with `inspect.signature` on all seven families (glm5_next has no `_parallel` variant).
- **Fix 3** (fetch sync keyed on banks' device): `tests/moe/test_disk_tier.py` is 6/6 green in a CUDA container, including the two `fetch_pending` tests you flagged.
- **Item 4**: the `layer<3` prefill print is now behind `FT_DISK_TIER_DEBUG`, and the per-layer `int(cache.step.item())` sync is gone — `cache.usage[disk]` takes the 0-d `cache.step` tensor directly (same dtype/device).
- **Item 5**: all four v0 preconditions are collected and raised as a single `ValueError` listing every unmet flag, and the exact flags are listed in `--moe-disk-tier --help`.

Re-validated end to end on a 2× RTX PRO 4000 box (TP=1, `nvidia/Qwen3.6-35B-A3B-NVFP4`, `FT_DISK_TIER_VERIFY=1`): RAM=64 → 4296/4296 slot checks match; RAM=32 (87.5% of experts disk-resident, exercising the new skip path harder) → 4902/4902 match, 0 mismatches, 11.69 tok/s median decode (11.85 at RAM=64).

One question on Fix 1: the unbacked rows `[K, E)` rely on the allocation being lazy (no physical pages until the tier fetches into them). That held on your box (28.9 GiB shmem) and ours — worth keeping in mind if anyone runs this on a kernel/mount with eager page allocation.

Happy to un-draft once you've had a look at the two new commits.

MT-z:
Looked at both commits, and re-ran the head on my box.

**Review.** `de43469` is byte-identical to the three fixes I tested here (diffed the two trees: no differences). In `795659d`: the `layer_id < 3` print behind `FT_DISK_TIER_DEBUG` and the combined precondition error both check out, and `cache.usage[disk] = cache.step` is a plain device-side broadcast of the 0-d step tensor (same dtype and device), so the per-layer D2H sync really is gone.

**Re-validation on this box** (RTX 4090 24 GB, 61 GB RAM), `795659d` as-is on its own base:
- Precondition check: `--moe-disk-tier on` without `--disable-moe-prefill-overlap` and without `--cuda-graph-max-bs 0` now fails once, listing both: `ValueError: --moe-disk-tier on: unmet preconditions: - requires --disable-moe-prefill-overlap - requires --cuda-graph-max-bs 0`.
- Ornith-1.5-35B-A3B-NVFP4, `--expert-ram-experts 128` (of 256 per layer), `FT_DISK_TIER_VERIFY=1`: boot 31 s, 128/256 experts per layer pinned, 906 slot verifies all `match=True`, RAM-prefix verify `mismatches=0/768`, no `[disk-tier dbg]` output without `FT_DISK_TIER_DEBUG`, no tracebacks. cgroup shmem 9.9 GiB against ~10.2 GiB expected for 128/256 of the 21.8 GB expert bytes -- the tail stayed unbacked.
- The GLM-5.3-Flash (K=48) and Qwen3.8-Flash-Next (K=224) numbers from the bug report were taken on a tree equal to `de43469`, so they stand for this head as well.

**On the lazy `[K, E)` rows** -- agreed it is an invariant worth stating, and it can be checked rather than assumed. What holds it up, phase by phase:

- *During load* the bank is `mmap.mmap(-1, size)`, i.e. MAP_SHARED anonymous (shmem). Untouched shmem pages cost nothing, but on shmem a **read** fault allocates a page too (no zero-page sharing for shared mappings), so the invariant is "nothing reads or writes rows `>= K` before `release_bank_tails`". Both loaders now honour it (the serial one skips before `get_tensor`, the parallel one filters at the reader), and `PinPipeline(prefix_rows=K)` pins only the prefix -- that matters because `cudaHostRegister` (and `mlock`) fault every page in.
- *After load* `release_range` replaces `[K, E)` with a MAP_PRIVATE|MAP_ANONYMOUS mapping at the same address, so from then on the rows have plain private-anon semantics: reads map the shared zero page and only writes allocate -- and nothing writes them: the fetch path lands O_DIRECT `preadv` in the pinned staging buffer and goes H2D straight into the slot, never through the bank tail. That part does not depend on the shmem mount at all.

Things that would break it: `MAP_POPULATE` (not used), pinning or mlocking a range that covers the tail, a debug/verify pass that reads the whole bank, or THP with `shmem_enabled=always|force`, which can back a 2 MiB huge page on first touch of a row -- bounded to one extra huge page per *touched* row; untouched rows stay unbacked. Our runs were with `shmem_enabled=never`, `enabled=madvise` (Ubuntu defaults). A quick experiment confirms the read-fault part: a 64 MiB MAP_SHARED anon mapping shows 0 resident, writing a 4 MiB prefix backs exactly 4 MiB, and reading one page of the tail backs exactly one page.

Evidence from real loads: cgroup `memory.stat` shmem = 28.9 GiB for GLM-5.3-Flash at K=48 (= 48/288 x 166 GiB) and 30.0 GiB for Qwen3.8-Flash-Next at K=224 (~224/512 x 63.3 GiB plus staging), plus the 9.9 GiB for Ornith at K=128 on this head (above), all under `MemoryMax=46G` with `MemorySwapMax=0` and `oom_kill 0` -- eager backing would have shown up as a kill, not as a silent regression.

Suggestion: assert it cheaply at startup with `mincore(2)` over the tail byte range right after `release_bank_tails` -- one syscall per bank layer, returns per-page residency -- and log "disk tier: tail rows resident X MiB of Y GiB", warning above a small threshold (`mincore` also counts zero-page reads, so it is conservative; cgroup `memory.stat` shmem is the number that actually costs RAM). The same check makes a CPU unit test: allocate a small bank, fill the prefix, release the tail, assert zero resident pages in `[K, E)`. Happy to send that as a follow-up if you want it in this PR.

No objection to un-drafting from my side.

<sub>Written with AI assistance; every number above was measured on my hardware and I can reproduce it.</sub>

CraigStone-Dev:
Great write-up — the phase-by-phase breakdown is exactly the right way to state the invariant, and I verified the load-side mechanics against the tree (in-place `MAP_PRIVATE|MAP_ANONYMOUS|MAP_FIXED` remap, `PinPipeline(prefix_rows=K)` → `pin_prefix` only, fetch path staging→H2D never through the bank tail). I also reproduced your 64 MiB read-fault experiment locally: 0 pages fresh, 1024 after a 4 MiB prefix write, 1025 after reading one tail page.

One data point from our side that makes your suggestion timely: our test box (2× RTX PRO 4000) runs `shmem_enabled=always` — one of the two configs you named. So rather than wait for a follow-up, I implemented the check in this PR as `b602e6d`:

- `tail_resident_bytes()` — `mincore(2)` over one bank's tail byte range.
- `check_tail_unbacked()` — runs in both NVFP4 loaders right after `release_bank_tails`; logs `[disk-tier] tail check rank=r/n: resident X MiB of Y MiB (warn bound Z MiB = 1x2MiB per bank layer)` and warns above the THP bound (one 2 MiB huge page per bank layer, per your boundary analysis).
- `test_tail_unbacked_after_release` — small bank, fill prefix, release tail, assert zero resident pages in `[K, E)`; then one tail write backs exactly one page.

Result on the `shmem_enabled=always` box: real boot at `--expert-ram-experts 32` (of 256) logs `resident 0 MiB of 15172 MiB` — the tail is fully unbacked even under the risky THP config, no warning. Full E2E still green: 4902/4902 slot verifies, 11.59 tok/s median. `test_disk_tier.py` is 7/7.

Un-drafting now — thanks again for the review, it caught the one bug that would have made the tier useless for its target case.

MT-z:
Writing that phase-by-phase answer sent me back into the release path, because I wanted to be sure of the second half of what I had claimed: that once `release_bank_tails` has run, the tail rows behave like ordinary private anonymous memory. Reading `release_range` again with that in mind, the step does not hold. **It gives back the address range, not the pages.** So I measured it, on a 2 GiB bank with the whole bank written and then the 1.5 GiB tail released, inside a systemd scope with memory accounting:

| tail release | `memory.current` | `memory.stat` shmem | RSS |
|---|---|---|---|
| current `release_range` (munmap + `MAP_FIXED`) | 2056 -> **2056** | 2048 -> 2048 | 2051 -> 515 |
| shared + `MADV_DONTNEED` | 2055 -> **2055** | 2048 -> 2048 | 2051 -> 515 |
| shared + `MADV_REMOVE` | 2056 -> **518** | 2048 -> 512 | 2051 -> 515 |
| `MAP_PRIVATE` + `MADV_DONTNEED` | 2053 -> **517** | 0 (anon instead) | 2051 -> 515 |

RSS falls in every row, which is what makes this easy to miss. The pages are still charged: `mmap(-1)` is `MAP_SHARED` anonymous, so the mapping is backed by an internal shmem object, and the prefix VMA keeps that whole object alive. Unmapping the tail range removes the VMA but leaves the object's pages in the page cache. Only punching a hole in the object (`MADV_REMOVE`) or not sharing it in the first place actually returns them. Measured on the charge, the remap and the `MADV_DONTNEED` it replaced land in the same place, and `HostBank.release()` -- a whole-buffer `MADV_DONTNEED` -- frees nothing either. What that means for this PR: if anything ever touches the tail -- the eager-allocation case you asked about, a future verify pass, a debug read -- `release_bank_tails` cannot get that RAM back for the life of the process. Your new boot check catches the touch; nothing undoes it.

**The one-line version of the fix is to stop sharing.** `mmap.mmap(-1, n)` is `MAP_SHARED` only because that is CPython's default; passing `flags=MAP_PRIVATE|MAP_ANONYMOUS` gives:

- reads of an untouched row map the shared zero page, so the invariant weakens from "nothing may read or write rows `>= K`" to "nothing may write them" -- a stray read pass over the tail costs 1536 MiB today and 0 with this change;
- `release_range` becomes one `madvise(MADV_DONTNEED)` that really frees, instead of `munmap` + `mmap(MAP_FIXED)`, which also removes the window where the range is unmapped and a concurrent touch would take SIGSEGV;
- `release()` starts doing what its docstring says.

`MADV_REMOVE` on the shared mapping is the alternative and it frees correctly too, but it keeps the read hazard and keeps the special case.

To be explicit about what this is not: it frees nothing today. With the loaders as they are, rows `>= K` are never touched, so there is nothing to release and the A/B totals below match. The change is about what happens the first time something does touch them -- today that is silent, costly and irreversible, and the same slip is easy to make again, since the tail is ordinary addressable memory that looks free to read.

Checks before proposing it: `cudaHostRegister` on a private mapping and a H2D copy out of the pinned prefix both work (`pin_prefix` unchanged); the loaders are thread pools, and TP ranks are `mp.Process` under `set_start_method("spawn")` (`server/launch.py:158`) with each rank allocating its own banks, so nothing needs the mapping shared and there is no fork/COW path. One addition in the patch: with `madvise` instead of a remap, dropping pages under a `cudaHostRegister`'d range would corrupt silently, so `release_range` now asserts the range does not overlap the pinned prefix -- the constraint your docstring already stated.

Note for your new tail check: with a private bank the tail shows as `anon` rather than `shmem`, and the THP warn bound is unchanged (`enabled=always` can still back the prefix/tail boundary with one 2 MiB page).

Validation on this box (RTX 4090 24 GB, 61 GB RAM), on `b602e6d` with the patch:

| A/B on this box, 2 runs each, alternating | stock `b602e6d` | + patch |
|---|---|---|
| cgroup `memory.current` | 12.32, 12.32 GiB | 12.30, 12.30 GiB |
| cgroup `memory.peak` | 25.84, 25.84 GiB | 25.83, 25.83 GiB |
| cgroup anon / shmem | 2.36 / 9.86 GiB | 10.89 / 1.33 GiB |
| slot verifies match=True / False | 414 / 0 | 414 / 0 |
| RAM-prefix verify | `mismatches=0/768` | `mismatches=0/768` |
| your tail check | `resident 0 MiB of 8670 MiB` | `resident 0 MiB of 8670 MiB` |
| time to first completion | 31, 31 s | 26, 26 s |
| `oom_kill` | 0 | 0 |

Ornith-1.5-35B-A3B-NVFP4, `--expert-ram-experts 128` of 256, `FT_DISK_TIER_VERIFY=1`, same flags throughout, greedy output identical. An earlier pair of runs gave the same verify counts and tail check. The bank moves from `shmem` to `anon` and the totals agree to 0.02 GiB, so this costs nothing and saves nothing. The load did come out faster on the patched runs (an earlier stock run also hit 26 s), but with this many runs I would not read a speed-up into it. `tests/moe/test_disk_tier.py` is 7/7 with the patch (your new tail-check test included), and `tests/moe/test_offload.py` + `tests/kernels/test_pinned_tensor.py` are 32 passed.

Diff below. Happy to open it as a PR against your branch instead if you prefer -- or to leave it until after this one merges, since the mapping flag is main's code and only the release path is yours.

<details><summary>Patch</summary>

```diff
diff --git a/python/freetoken/moe/host_banks.py b/python/freetoken/moe/host_banks.py
index 388893c..c863f95 100644
--- a/python/freetoken/moe/host_banks.py
+++ b/python/freetoken/moe/host_banks.py
@@ -79,7 +79,7 @@ class HostBank:
 
     The buffer is rounded up to the O_DIRECT block; ``tensor`` views exactly ``nbytes``. ``backing=None`` follows ``FREETOKEN_BANK_CUDA_ALLOC``."""
 
-    __slots__ = ("tensor", "addr", "nbytes", "_buf", "_pinned", "_locked")
+    __slots__ = ("tensor", "addr", "nbytes", "_buf", "_pinned", "_pinned_bytes", "_locked")
 
     def __init__(self, shape: tuple[int, ...], dtype: torch.dtype,
                  *, backing: str | None = None):
@@ -104,11 +104,19 @@ class HostBank:
             self.addr = raw.data_ptr() + off
             assert self.addr % _BLK == 0
             self._pinned = True  # born pinned+mapped; pin() is a no-op
+            self._pinned_bytes = asize
         else:
-            self._buf = mmap.mmap(-1, asize)  # lazy: address space only, no resident pages yet
+            # MAP_PRIVATE, not CPython's default MAP_SHARED: on a shared anonymous mapping a
+            # *read* fault allocates a page (no zero-page sharing) and MADV_DONTNEED is ignored,
+            # so an untouched region is only free by convention and a freed one never comes back.
+            # Private anonymous gives both for real: reads map the shared zero page, and
+            # release_range() actually returns memory. Nothing needs the mapping to be shared --
+            # the loaders are thread pools and ranks are mp-spawned, each with its own banks.
+            self._buf = mmap.mmap(-1, asize, flags=mmap.MAP_PRIVATE | mmap.MAP_ANONYMOUS)
             _LIVE_BUFFERS.append(self._buf)
             self.addr = ctypes.addressof(ctypes.c_char.from_buffer(self._buf))
             self._pinned = False
+            self._pinned_bytes = 0
         self.tensor = torch.frombuffer(self._buf, dtype=dtype, count=self.nbytes // elsize).view(*shape)
         self._locked = False
 
@@ -140,6 +148,7 @@ class HostBank:
                 f"cudaHostRegister failed for {len(self._buf) / 2**30:.1f} GiB"
             ) from exc
         self._pinned = True
+        self._pinned_bytes = len(self._buf)
 
     def pin_prefix(self, nrows: int) -> None:
         """Pin only the first ``nrows`` rows (disk tier: the rest stays disk-resident).
@@ -160,44 +169,32 @@ class HostBank:
                 f"cudaHostRegister failed for {nbytes / 2**30:.1f} GiB prefix"
             ) from exc
         self._pinned = True
+        self._pinned_bytes = nbytes
 
     def release_range(self, offset: int, nbytes: int) -> None:
-        """Free a byte range of the backing mapping by replacing it IN PLACE with a
-        fresh MAP_PRIVATE anonymous mapping at the same virtual address.
-
-        HostBank's buffer is a MAP_SHARED /dev/zero mapping (CPython's
-        ``mmap(-1)``), and the kernel silently ignores MADV_DONTNEED on shared
-        mappings -- the pages would stay resident. Replacing the range with a
-        private zero mapping frees them while keeping every existing pointer
-        and torch view valid (same address). The range must be page-aligned
-        and must not overlap a pinned prefix (the disk tier's unpinned tails).
-        """
-        import ctypes
+        """Free a byte range of the backing mapping with MADV_DONTNEED.
+
+        The bank is a MAP_PRIVATE anonymous mapping, so dropping a range frees the
+        pages outright and a later read faults the shared zero page again; every
+        existing pointer and torch view stays valid (the mapping is never replaced).
 
-        _BLK = 4096
+        The range must be page-aligned and must not overlap the pinned prefix:
+        dropping pages under a cudaHostRegister'd range corrupts silently, so it is
+        asserted here rather than left to the caller (the disk tier's unpinned tails).
+        """
         assert offset % _BLK == 0 and nbytes % _BLK == 0, (
             "release_range: page-aligned range required")
-        libc = ctypes.CDLL("libc.so.6", use_errno=True)
-        libc.munmap.argtypes = [ctypes.c_void_p, ctypes.c_size_t]
-        libc.munmap.restype = ctypes.c_int
-        libc.mmap.restype = ctypes.c_void_p
-        libc.mmap.argtypes = [
-            ctypes.c_void_p, ctypes.c_size_t, ctypes.c_int, ctypes.c_int, ctypes.c_int, ctypes.c_long]
-        addr = self.addr + offset
-        if libc.munmap(addr, nbytes) != 0:
-            raise OSError(ctypes.get_errno(), "munmap failed")
-        PROT_READ_WRITE = 3
-        MAP_PRIVATE_ANON = 0x22  # MAP_PRIVATE | MAP_ANONYMOUS
-        MAP_FIXED = 0x10
-        MAP_FAILED = (1 << 64) - 1
-        new_addr = libc.mmap(addr, nbytes, PROT_READ_WRITE, MAP_PRIVATE_ANON | MAP_FIXED, -1, 0)
-        if new_addr in (None, MAP_FAILED):
-            raise OSError(ctypes.get_errno(), "mmap(MAP_FIXED) failed")
-        assert new_addr == addr, "MAP_FIXED returned a different address"
+        assert offset >= self._pinned_bytes, (
+            f"release_range: [{offset}, {offset + nbytes}) overlaps the pinned prefix "
+            f"[0, {self._pinned_bytes})")
+        if nbytes:
+            self._buf.madvise(mmap.MADV_DONTNEED, offset, nbytes)
+
     def release(self) -> None:
         """Drop the resident pages; the address space stays valid, the contents become undefined.
 
-        For buffers that are done being read (the converter). No-op for born-pinned banks: registered pages cannot be dropped."""
+        For buffers that are done being read (the converter). No-op for born-pinned banks: registered pages cannot be dropped.
+        (This frees memory only because the mapping is MAP_PRIVATE; the kernel ignores MADV_DONTNEED on shared ones.)"""
         if self._pinned:
             return
         self._buf.madvise(mmap.MADV_DONTNEED)
```

</details>

<sub>Written with AI assistance; every number above was measured on my hardware and I can reproduce it.</sub>

MT-z:
Found one more while stress-testing the tier on a deliberately small card: **`FT_DISK_TIER_VERIFY=1` crashes any boot that captures CUDA graphs, whether or not the tier is on.**

Repro on this box with no tier flags at all (RTX 4090, 22.67 GiB free):

```bash
FT_DISK_TIER_VERIFY=1 ft serve --model-path ornith-ai/Ornith-1.5-35B-A3B-NVFP4 \
  --moe-cache-auto --memory-ratio 0.9 --kv-reserve-tokens 8192
```

```
Capturing graphs: bs = 4 | avail_mem = 0.78 GiB
[copy-miss] layer=0 fused=True n=8 evict=[0, 1, 2, 3] src=[74, 112, 160, 196]
torch.AcceleratorError: CUDA error: operation not permitted when stream is capturing   (cudaErrorStreamCaptureUnsupported)
torch.AcceleratorError: CUDA error: operation failed due to a previous error during capture   (cudaErrorStreamCaptureInvalidated)
```

The frame is `offload_cache.py:1037`, in `copy_missing`:

```python
if os.environ.get("FT_DISK_TIER_VERIFY") and layer_id == 0:
    print(f"[copy-miss] layer={layer_id} fused={self._copy_fused_ok} "
          f"n={int(self.num_indices.item())} "
          f"evict={self.evict_slots[:4].cpu().tolist()} "
          f"src={self.src_indices[:4].cpu().tolist()}", flush=True)
```

`copy_missing` runs inside the captured decode graph, and `.item()` / `.cpu()` are device-to-host syncs, which capture forbids.

Two things make it easy to hit:

- The gate is the environment variable alone. The other two verify sites reachable with the tier off check the tier first (`offload_cache.py:1068` tests `self._disk_tier is not None`, `layers/moe.py:315` tests `cache.disk_tier_enabled`); this one does not, so it fires with `--moe-disk-tier` off entirely.
- With the tier **on** it can never fire, because v0 requires `--cuda-graph-max-bs 0`. So it only appears when the variable is still exported from an earlier tier session -- which is exactly how I hit it: every graph-capturing run in my harness failed, and I first misread it as a small-card problem.

Not memory related: it fails identically at 22.67 GiB free and at 10.29 GiB free, and lowering `--memory-ratio` to 0.85 or 0.8 does not help. With the variable unset, the same 10.29 GiB configuration boots in 21 s (2950 slots) and serves.

Suggested fix -- gate it like its neighbours and skip it while capturing:

```python
if (self._disk_tier is not None and layer_id == 0
        and os.environ.get("FT_DISK_TIER_VERIFY")
        and not torch.cuda.is_current_stream_capturing()):
```

or simply drop the print: `verify_decode_mapping` and `verify_ram` already cover the same ground, and both are properly gated.

While I was there, the rest of the low-VRAM picture came out clean: with VRAM ballasted down to 10.29 GiB free, Ornith-1.5-35B-A3B-NVFP4 boots through the tier at `--expert-ram-experts 128` in 25 s (2910 slots, 252 slot verifies all matching, tail check `resident 0 MiB of 8670 MiB`), and `--expert-ram-experts 256` is rejected with a legible `--expert-ram-experts must be in (0, 256)`.

<sub>Written with AI assistance; every number above was measured on my hardware and I can reproduce it.</sub>

MT-z:
Went back over the provider -> engine path on the merged head, and two of the four things I found need fixing before this lands. Neither showed up in my runs, because this box only exercises the two families that happen to work.

### 1. Only `qwen3_5_moe` (and whoever re-exports its setup) ever attaches a fetcher -- the other NVFP4 families release the tail and then read zeros

The chain:

- `--moe-disk-tier on` builds a `DiskTierSpec` and hands it to `load_expert_banks` (`engine.py:568-590`, `:616`).
- `Nvfp4DiskIndex` is constructed in exactly one place in the tree -- `qwen3_5_moe/weight.py:898` (`grep -rn "Nvfp4DiskIndex(" python/` returns that one hit), and that is also the only `dataclasses.replace(banks, disk_index=...)`.
- A family whose module does not resolve `setup_offload_expert_banks` goes to `_nvfp4_banks` (`expert_banks.py:168`), which forwards `disk_tier` into the base loader -- so rows `[K, E)` are skipped and `release_bank_tails` runs (`nvfp4_banks.py:221-223`, `:361-363`) -- and then returns `ExpertBanks` with `disk_index=None`.
- The attach is guarded by `if banks.disk_index is not None` (`engine.py:653`), and the "disk tier: K/E experts/layer pinned" log line sits inside that same guard. So no fetcher, no message, and the precondition block (`engine.py:576-586`) has no "is this family supported" branch to say otherwise.
- `copy_missing` then copies the released rows into the slot cache and the grouped GEMM multiplies by zeroed experts. Wrong logits, no error.

Resolving `_load_attr(spec.module, "setup_offload_expert_banks")` over all of `_MODEL_REGISTRY`:

| | families |
|---|---|
| resolves a setup (fetcher attaches) | `qwen3_5_moe`, `qwen4_exp` (re-exported at `qwen4_exp/__init__.py:25`), `gpt_oss` |
| falls through to `_nvfp4_banks` (no index) | `glm5_next`, `glm4_moe`, `gemma4`, `minimax_m2`, `minimax_m3`, `deepseek_v4`, `glm_moe_dsa`, `muse_glimmer` |

That table also explains why my earlier runs looked clean: Ornith-1.5 is `qwen3_5_moe` and Qwen3.8-Flash-Next is `qwen4_exp` through the re-export, so both log the attach line and both verify (16302 slot checks on Qwen3.8 at K=224, 414 on Ornith at K=128, zero mismatches). The GLM-5.3-Flash figures in my first report were load-phase only -- shard count, load time, cgroup `shmem` -- which is exactly the part that looks right whether or not a fetcher exists.

The cheap fix is to fail where the code already fails loudly (`_nvfp4_banks`'s own `NotImplementedError`, the `eq != "nvfp4"` guard):

```python
if banks.disk_index is not None:
    cache.attach_disk_tier(...)
    logger.info_rank0(...)
elif disk_tier is not None:
    raise NotImplementedError(
        "--moe-disk-tier on: this model family builds no disk index; experts [K, E) "
        "are released at load and would never be refetched")
```

Building the index inside `_nvfp4_banks` instead would fix the families rather than reject them, since what `qwen3_5_moe` adds is a per-family source spec -- but the guard is what stops silent corruption today.

### 2. `release_bank_tails` asserts on any `--expert-ram-experts` whose row offset is not page-aligned

`release_bank_tails` (`disk_tier.py:57-58`) passes `ram_experts * row_bytes` and `bank.nbytes - offset` to `release_range`, which hard-asserts 4 KiB alignment on both (`host_banks.py:178-179`). `row_bytes = bank.nbytes // num_experts` is not page-aligned in general -- the scale banks are small. Real values from the `[disk-tier-init]` line here: Ornith `host_row_bytes=[1048576, 131072, 2048, 524288, 65536, 4096]`, Qwen3.8-Flash-Next `[1638400, 204800, 2560, 819200, 102400, 5120]`. So whether a boot survives depends on `K`:

| E, row bytes, K | `release_bank_tails` |
|---|---|
| 256, 2048, 128 | ok |
| 256, 2048, **127** | `AssertionError: release_range: page-aligned range required` |
| 512, 2560, 224 | ok |
| 512, 2560, **225** | `AssertionError` |

Repro, no GPU and no checkpoint needed:

```python
import torch
from freetoken.moe.host_banks import HostBank
from freetoken.moe.disk_tier import release_bank_tails

banks = {"gate_up_scale": [HostBank((256, 2048), torch.uint8)]}  # a real Ornith bank row size
release_bank_tails(banks, 256, 127)   # AssertionError; 128 is fine
```

In other words any odd `--expert-ram-experts` on Ornith, or anything not a multiple of 8 on Qwen3.8, fails at load -- after the wait for the weights, and with a message that points at page alignment rather than at the flag. Since rows `[K, E)` were never written, the release is an optimization and not an invariant: warning and skipping (or rounding the offset up to the next page and releasing the remainder) turns this back into a no-op instead of a boot failure.

### 3. Minor: `qwen3_5_moe`'s default `fp8_block` path drops the flag silently

`setup_offload_expert_banks` only enters the disk-tier branch inside `if eq != "fp8_block":`, and the `eq != "nvfp4"` guard (`qwen3_5_moe/weight.py:891-893`) lives inside that branch, so it never fires for the default path. `_build_fp8_expert_banks(...)` is called without `disk_tier`. Net effect on a default `fp8_block` checkpoint: the flag is accepted, nothing is released, nothing is fetched, no diagnostic. Raising the same `NotImplementedError` there would be consistent.

### 4. Trivial: the `FT_DISK_TIER_VERIFY` init probe indexes expert 100

`disk_tier.py:264` sums `self._index.row_segments(bi, 0, 100)` to print `disk_row_bytes`. The intent is one expert's row, so it wants `row_segments(bi, 0, 0)`; as written it slices past the per-layer buffer and `struct.unpack_from` raises for any model with `E < 100`. Gated on the verify flag, and my three checkpoints have 256/288/512 experts, so it never fired here.

Happy to send 1 and 2 as a patch if you want them in this PR rather than in your own words.

<sub>Written with AI assistance; every number above was measured on my hardware and I can reproduce it.</sub>

CraigStone-Dev:
All three findings verified against the tree and applied — new head `a6bd5c0`, tested on Rudi (2× RTX PRO 4000 Blackwell, 62 GB RAM, `shmem_enabled=always`), Qwen3.8-Flash-Next-NVFP4 (512 experts, the unaligned-row-geometry case).

**5518521155 (MAP_SHARED bank, released tails never free):** applied your patch to `host_banks.py` verbatim — `MAP_PRIVATE|MAP_ANONYMOUS` bank, `release_range` = one `madvise(MADV_DONTNEED)` + the pinned-prefix overlap assert via `_pinned_bytes`, `release()` docstring. Confirmed in the E2E cgroup accounting: the expert banks now show as `anon` (30.6 GiB anon / 2.2 GiB shmem at K=224, where shmem used to hold the whole bank set), tail check `resident 0 MiB of 36551 MiB`.

**5519070434 (`FT_DISK_TIER_VERIFY` probe kills graph capture):** applied your suggested gate verbatim (`self._disk_tier is not None` + `not torch.cuda.is_current_stream_capturing()`). Added `test_copy_miss_verify_probe_gated_on_disk_tier` (tier off + env set → no probe).

**5520770513:**
1. Engine attach site now has the `elif disk_tier is not None: raise NotImplementedError(...)` guard, so a family that releases tails but builds no index fails loudly instead of serving zeros. Re-ran your `_load_attr` table: ATTACHES = gpt_oss, qwen3_5 (both arch names), qwen4_exp (re-export); everything else falls through and is now rejected.
2. `release_bank_tails` warns and skips banks whose row boundary is not page-aligned (tail rows were never written, so nothing is lost) + regression test `test_release_bank_tails_unaligned_row_boundary` (your K=127/256, row 2048 repro). E2E at **K=225** (was an `AssertionError` at load): boots, 96 warn+skip lines (48 layers × the 2560/5120-byte banks), 1278/1278 verifies, coherent output.
3. The `fp8_block` default path raises the same `NotImplementedError` (verified: `disk tier: only nvfp4 experts are supported (got expert_quant='fp8_block')`).
4. Init probe now reads `row_segments(bi, 0, 0)` — the `[disk-tier-init]` line's `disk_row_bytes` came out `[1638400, 204800, 8, 819200, 102400, 4]` (segment counts, not expert-100's row).

**Test results (Rudi):** `tests/moe/test_disk_tier.py` 8/8 (incl. both new tests), `tests/moe/test_offload.py` 25/25, `tests/kernels/test_pinned_tensor.py` 7/7. E2E K=224 with `FT_DISK_TIER_VERIFY=1`: 1284/1284 slot verifies, `mismatches=0/1344` RAM-prefix, correct answers, tail unbacked.

MT-z:
`a6bd5c0` reproduces here, on the other side of the hardware split -- one Ada card (RTX 4090 24 GB, sm_89), single rank, `shmem_enabled=never` where Rudi has `always`. Qwen3.8-Flash-Next-NVFP4 at `--expert-ram-experts 224`, `FT_DISK_TIER_VERIFY=1`, nothing local on top of your head:

- **cgroup accounting matches your box almost exactly**: `anon` 30.44 GiB, `shmem` 2.16 GiB, against your 30.6 / 2.2. `tail check rank=0/1: resident 0 MiB of 36551 MiB`.
- 456 slot verifies, all `match=True`; RAM-prefix `mismatches=0/1344`; no tracebacks; coherent output.
- No `not page-aligned` lines at this K -- see below for why that is arithmetic rather than luck.

The cgroup line is the one number the old logs could not show, and it now says the same thing on two kernels with opposite THP settings.

**One thing worth putting in `--help`, now that the unaligned case warns instead of aborting.** Which `K` values release cleanly is decided by the model's *smallest* bank row -- the fp32 global scales -- so it is a per-model arithmetic rule rather than a per-model accident:

| model | smallest row | rule for `--expert-ram-experts` | usable values |
|---|---|---|---|
| Qwen3.8-Flash-Next-NVFP4 (E=512) | 2560 B = 2^9 x 5 | multiple of **8** | 63 of 511 |
| Ornith-1.5-35B-A3B-NVFP4 (E=256) | 2048 B = 2^11 | **even** | 127 of 255 |

`K * min(row_bytes) % 4096 == 0` is the whole of it: 2560 needs `K*5 = 0 mod 8`, 2048 needs `K` even. That is why your K=225 run produced 96 warn+skip lines (48 layers x the 2560/5120-byte bands) while K=224 produces none -- 224*2560 is exactly 140 pages. I picked K=255 for a measurement earlier today and hit the old assert with it, which is how I know the failure mode is easy to walk into; a line in `--help` saying "keep `K * smallest bank row` page-aligned, or the small scale bands stay resident" would have saved me a boot.

**And an offer on the family guard.** The `elif disk_tier is not None: raise NotImplementedError(...)` is the right call -- loud beats zeroed experts. If you would rather lift it than keep it, the work is small, because every family already owns the `Nvfp4ExpertSourceSpec` its loader reads with: expose it (`nvfp4_expert_source_spec`, re-exported from each package the way the loaders are) and build the index in `_nvfp4_banks`, which is also where the release happens, so the two cannot drift. I have that as a commit, with the index verified byte-for-byte against a GLM-5.3-Flash checkpoint -- 288 experts, 42 MoE layers, 118 shards, 9 of 9 spot checks across packed weights, per-row scales and the fp32 globals -- plus a test that every NVFP4 family exposes a spec and that a family without one is refused at load.

What it does NOT have is an end-to-end serving run on any newly supported family: GLM's 18.0 GiB of non-expert weights plus one layer's 3.89 GiB expert floor do not fit this 24 GB card, so I can verify where the index points but not that the model then answers correctly. If that gap is why you would rather keep the guard as the contract, that is a good reason and I would not argue with it. Say the word either way.

<sub>Written with AI assistance; every number above was measured on my hardware and I can reproduce it.</sub>

CraigStone-Dev:
Thanks for the second-hardware confirmation — the cgroup line matching across opposite THP settings (`never` vs `always`) is exactly the evidence the shmem→anon change needed.

**The `--help` line is in** (`0f4c26d`, now the PR head). I re-derived your table independently and it's exact: Qwen3.8-Flash-Next (E=512, rows 1638400/204800/2560/819200/102400/5120) → usable K are precisely the 63 multiples of 8; Ornith (E=256, rows …/2048/…) → precisely the 127 even K. And K=225 → exactly the 2560/5120 banks unaligned (the 96 warn lines), K=224 → none. The help text now says: keep `N * (smallest bank row bytes)` a multiple of 4096 or the small scale banks' tails stay resident (warns, does not abort), with both examples.

**On the family guard: we'd like to lift it — send the commit.** I don't see a public fork/branch for it; push the branch anywhere I can fetch (or paste the diff) and I'll take it in. I checked the premise against the tree: the NVFP4 families that currently hit the guard are exactly gemma4, glm4_moe, glm5_next, minimax_m2, minimax_m3, and each already owns its `_NVFP4_SOURCE_SPEC` (glm5_next's `_select_expert_source_spec` for the compressed-tensors variant is the one to expose), so building the index in `_nvfp4_banks` next to the release is the right seam.

**The E2E gap is closable here:** Rudi has 62 GB RAM + 2× RTX PRO 4000 (24 GB each), so GLM-5.3-Flash-NVFP4 (18 GiB non-expert + the per-layer expert floor) should fit with the tier on — that's the plan for validating the newly enabled families once the commit lands.

MT-z:
Pushed — it sits directly on your current head, so it should be a clean cherry-pick:

```bash
git fetch https://github.com/MT-z/FreeToken.git feat/moe-disk-tier-all-nvfp4-families
git cherry-pick FETCH_HEAD
```

`cb7d07e`, parent `0f4c26d`. I verified it fetches from an empty clone over https, so nothing
depends on my local state. 18 files, +186/-26.

## It lands where you said it should

Your reading was right, and it is the reading the commit already had: the index is built in
`_nvfp4_banks` **next to the release**, not in a per-family setup. The comment in the diff says why:

> Build the index HERE, not in a per-family setup: the loader below releases expert rows `[K, E)`
> for every family, so a family that reaches the release without an index would serve zeroed
> experts. Resolving the spec through the family hook keeps the index and the loader reading the
> same rows.

The seam is one new hook, `nvfp4_expert_source_spec(model_path, model_config)`, resolved through
the existing `_model_override` machinery in `models/weight.py`. Each family exposes the spec it
already owns; `glm5_next` exposes the selector rather than a constant, exactly as you flagged,
because it picks between compressed-tensors and modelopt naming per checkpoint.

## The guard stays, with a corrected message

This is the one conflict against your head, and it is worth a look. The current wording is the one
I suggested and you took:

> only qwen3_5_moe and re-exporters attach one

**That parenthetical becomes false with this commit**, since every NVFP4 family then attaches one.
I resolved in favour of the version in my branch, which drops the stale list and adds the format:

> `--moe-disk-tier on: this checkpoint's expert provider returned no disk index, so experts
> released at load would never be refetched (quant_format={...!r})`

The guard itself is unchanged in spirit — it is now a last-resort net for a checkpoint whose family
genuinely has no NVFP4 spec, rather than a list of families to keep in sync by hand.

## Tested here

```
on 0f4c26d + this commit, RTX 4090 (sm_89)
  tests/moe/test_disk_tier_families.py   15 passed   (new in this commit)
  tests/moe/test_disk_tier.py             8 passed
  tests/moe/test_offload.py              24 passed
```

And against real checkpoints, resolving the spec end to end (no GPU needed for this part):

| family | checkpoint | spec |
|---|---|---|
| **glm5_next** | `LibertAIDAI/GLM-5.3-Flash-NVFP4` | resolves |
| **gemma4** | `nvidia/Gemma-4-26B-A4B-NVFP4` | resolves |
| qwen3_5_moe | `ornith-ai/Ornith-1.5-35B-A3B-NVFP4` | resolves |
| qwen4_exp | `RadixArk/Qwen3.8-Flash-Next-NVFP4` | resolves |

The first two are on your list of newly enabled families, so that is two of the five with a real
checkpoint behind them before it reaches Rudi. I cannot take either of them further here —
GLM-5.3-Flash does not fit on a single 24 GB card even with the tier, which is why your 62 GB /
2×24 GB plan is the right place for the E2E. `minimax_m2`, `minimax_m3` and `glm4_moe` are
untested by me at any level beyond the unit tests; I have no checkpoint for those.

Shout if you would rather have it shaped differently — it is your PR and I am happy to redo it.

<sub>Written with AI assistance; the test results and checkpoint resolutions above are from my own
hardware and I can reproduce them.</sub>

CraigStone-Dev:
Cherry-picked cleanly onto `0f4c26d` (now the PR head, `7915ac1`) and validated on Rudi (2× RTX PRO 4000 Blackwell, 62 GB RAM, `shmem_enabled=always`).

**CPU suite (his tree):** `test_disk_tier_families.py` 15/15, `test_disk_tier.py` 8/8, `test_offload.py` 25/25, `test_pinned_tensor.py` 7/7.

**Byte-for-byte, re-run here against the real checkpoint** (GLM-5.3-Flash-NVFP4, glm5_next, on Rudi): the index resolves through `nvfp4_expert_source_spec` (modelopt variant, no kind_map), builds 6 banks × 42 MoE layers, and 5 sampled rows (packed weights, per-row scales, fp32 globals) are byte-identical to the safetensors tensors at the indexed offsets — 8/8 segments, 0 mismatches.

**The E2E gap is closed — on gemma4.** `nvidia/Gemma-4-26B-A4B-NVFP4` (the checkpoint you flagged as resolving) served end-to-end on Rudi at `--expert-ram-experts 64` (a multiple of 16, per the smallest-row rule: 2816 B rows), `FT_DISK_TIER_VERIFY=1`:
- boots clean, `disk tier: 64/128 experts/layer pinned in RAM`
- **450/450 slot verifies, 0 mismatches**; RAM-prefix `mismatches=0/384`
- tail check `resident 0 MiB of 6141 MiB`
- cgroup **9.36 GiB anon / 2.29 GiB shmem** (banks anon, as expected)
- coherent + correct: "396" for 17·23+5, "2, 3, 5, 7, 11" for the primes

**Regression on the refactored path (qwen4_exp):** Qwen3.8-Flash-Next at K=224 re-run on the new head — 1284/1284 slot verifies, `mismatches=0/1344`, correct output, tail `0 MiB of 36551 MiB`. So the family that previously built the index in its own setup is unchanged end-to-end, now going through the shared provider.

**GLM-5.3-Flash E2E:** confirmed your fit analysis on Rudi — TP=1 leaves only 646 MB of cache budget after the ~18 GiB non-expert weights (the 288-slot floor needs 4.18 GB), and TP=2 is rejected by `glm5_next` (TP=1-only weight loading). So GLM-5.3-Flash stays load-verified-only here too; gemma4 is the family that carries the serving E2E.

The guard message change is the right call — the old "only qwen3_5_moe and re-exporters" parenthetical was already stale.

CraigStone-Dev:
@MT-z — I rebased the branch onto the latest `main` (the quantization refactor #418/#426/#427 landed in between and the PR had conflicts). The disk tier now rides the new architecture:

- The index is built from each family's `nvfp4_expert_spec` hook (your round-4 design, now upstream's), so index and reader read the same rows.
- The NVFP4 reader takes a `skip_experts_from`: experts `[K, E)` are never read (serial: no `get_tensor`; parallel: filtered at the reader), and each skipped expert yields an empty piece, so `build_expert_banks` completes the layer, pins only the K prefix, and releases the tail next to where the index is built.
- Everything else (O_DIRECT fetch workers, staging ring, `FT_DISK_TIER_VERIFY`, the boot preconditions) carried over unchanged.

Re-validated on 2× RTX PRO 4000 Blackwell, 62 GB RAM:

| Model | K | Result |
|---|---|---|
| Qwen3.8-Flash-Next-NVFP4 (E=512) | 224 | 1086/1086 slot verifies, tail 0 MiB of 36551 MiB, correct output |
| Gemma-4-26B-A4B-NVFP4 (E=128) | 64 | 450/450 slot verifies, cgroup 9.4/2.3 GiB, correct output |
| GLM-5.3-Flash-NVFP4 (E=288) | — | index byte-for-byte vs checkpoint (18/18 segments) |

The rebase also caught a real bug in the reader's completeness check (it expected `L*(E-K)*9` tensors instead of `L*K*9` — masked whenever K == E-K). CPU suite green, no regressions vs `main`.

iSOcH:
Briefly tested on Qwen3.8-Flash-Next-NVFP4.

**System:** RTX 4090 24 GB, 96 GB DDR5 6000 host RAM, checkpoint on a ZFS-on-NVMe pool.

Baseline (`main` @ `9535656`, no tier flags):
```
ft serve --model Qwen3.8-Flash-Next-NVFP4 --host 0.0.0.0 --port 8080 \
  --max-running-requests 1 --kv-reserve-tokens 262144 --max-seq-len-override 262144
```
63.5 GiB of pinned expert banks, `prefill_overlap=True`, CUDA graphs captured. Decode 20–25 tok/s.

This branch, commit `badf5f4`:
```
<same as above> --moe-strategy offload --moe-disk-tier on \
  --expert-ram-experts 224 --disable-moe-prefill-overlap --cuda-graph-max-bs 0
```
Boots fine, output coherent, no errors. Decode ~3 tok/s at `expert-ram-experts` 224, 4–5 tok/s at 384 and 7-8 tok/s with 448.

CraigStone-Dev:
@MT-z — rebased onto current `main` (`71eb1f0`, v0.1.3 + 4) and re-validated. New head `91289ba`, PR is mergeable again.

Upstream moved 16 commits past the old merge base; the conflicts were small:

- **`engine/engine.py`** — the new `_weight_load_context()` wrapper (WeightLoadError, #518) now wraps `load_expert_banks`; the `disk_tier=` argument is passed inside it. The v0 precondition `ValueError` and the no-disk-index `NotImplementedError` stay *outside* the wrapper, so their error types and messages are unchanged.
- **`models/nvfp4_banks.py`** — upstream dropped the direct index-json read in favour of `loader.safetensors_weight_map` (index json, or shard headers when the checkpoint ships none); the `json` import is gone, `itertools` stays (used by the disk-skip pieces).
- **`moe/disk_tier.py`** — `Nvfp4DiskIndex` now resolves its weight map through the same `loader.safetensors_weight_map` helper the NVFP4 reader uses, keeping the index-reads-the-same-rows invariant from the last rebase.
- **`tests/moe/test_disk_tier_families.py`** — #438 moved `qwen3_5_moe`'s `nvfp4_expert_spec` hook onto the installed QuantConfig (dialect-aware stored names), so the `spec is mod._NVFP4_SOURCE_SPEC` assertion no longer applies to that family. The test now installs a `ModelOptConfig({"quant_algo": "NVFP4"})` and checks the hook's `kind_map` against the dialect's storage table — same invariant (hook and loader read the same rows), stated for the new architecture.

**CPU suite** (CUDA container, tree at the rebased head): `tests/moe` 48/48 (incl. both new tests above). Full `tests/` run: 1361 passed / 16 failed / 18 errors — and a clean `upstream/main` baseline in the same image gives the identical 16 failures + 18 errors (fixture/env issues, e.g. `torch.ops` import), so zero regressions vs main; the pass-count delta is exactly the 24 disk-tier tests.

**E2E** (Rudi: 2× RTX PRO 4000 Blackwell, 62 GB RAM, `shmem_enabled=always`, `FT_DISK_TIER_VERIFY=1`), `nvidia/Qwen3.6-35B-A3B-NVFP4` (qwen3_5_moe, E=256) at `--expert-ram-experts 128`:

- boots clean: `disk tier: 128/256 experts/layer pinned in RAM`
- **2316/2316 slot verifies `match=True`, 0 mismatches**
- tail check `rank=0/1: resident 0 MiB of 8670 MiB`
- coherent + correct output

The GLM-5.3-Flash byte-for-byte index check from the previous head is unchanged in logic (the index only changed its weight-map *source*, not the row math), and that checkpoint isn't on this box anymore, so I'm not re-claiming it here.

CraigStone-Dev:
@iSOcH — thanks for testing the head on your box. The 3 tok/s number is worth understanding, and I think the comparison is confounded by the v0 preconditions: your baseline ran with CUDA graphs captured and prefill overlap on, while the tier currently **requires** `--cuda-graph-max-bs 0` and `--disable-moe-prefill-overlap` (fetches can't run inside graph capture, and the prefill overlap path doesn't know about the fetch workers yet).

To separate the two effects I ran a controlled A/B on our box (RTX PRO 4000, 62 GB RAM, NVMe, `nvidia/Qwen3.6-35B-A3B-NVFP4`, E=256, TP=1, 300-token decodes, median of 3, rebased head `91289ba`):

| config | median tok/s |
|---|---|
| no tier, CUDA graphs + prefill overlap (your baseline shape) | 77.6 |
| no tier, `--disable-moe-prefill-overlap --cuda-graph-max-bs 0` | 24.2 |
| tier on, K=128 (50% of experts on disk), same two flags | 15.5 |
| tier on, K=64 (75% of experts on disk), same two flags | 13.2 |

So on this box the v0 flags cost **3.2×** and the disk tier itself costs **36%** at a 50/25 split and **46%** at 75/25 — the fetch path is on the critical path, but it is not the dominant term.

That would put your expected tier-on number around 6–9 tok/s (your 20–25 baseline divided by ~3), not 3 — so there may be a ZFS-specific component on top (O_DIRECT over ZFS can be much slower than over ext4/xfs, and the fetch workers are O_DIRECT `preadv` by design). The one run that would isolate it on your box: **same branch, same two flags, `--moe-disk-tier` off** (all 512 experts in RAM, which your 96 GB box holds). If that lands around 8–10 and tier-on stays at 3, the fetch path on ZFS is the thing to chase and I'd like to dig in; if it lands around 3 too, the whole gap is the graphs/overlap preconditions.

Either way the biggest v1 win is clear from these numbers: lifting the `--cuda-graph-max-bs 0` precondition (graph-safe fetches) would give back most of the 3.2× for every tier user. Happy to take that on as a follow-up once v0 lands.

PR #366, opened 2026-09-03
feat(gguf): implement Q8_0 dequantization

### LINKED ISSUE #358: gemma4 GGUF: Q8_0 is declared in BLOCK_SHAPE/GGML_NAME/__all__ but missing from _DEQUANT, so unsloth UD-*_XL quants fail with NotImplementedError
## Summary

`python/freetoken/models/gguf/dequant.py` declares Q8_0 everywhere except the one table that
matters. `GGML_Q8_0 = 8` has a `BLOCK_SHAPE` entry `(32, 34)`, a `GGML_NAME` entry `"Q8_0"`,
and is listed in `__all__` — but there is no `dequant_q8_0`, so Q8_0 is absent from `_DEQUANT`
and `dequantize()` falls through to:

```python
fn = _DEQUANT.get(ggml_type)
if fn is None:
    raise NotImplementedError(
        f"dequant for ggml type {GGML_NAME.get(ggml_type, ggml_type)} not implemented"
    )
```

The module docstring states the intended scope as "Q4_0, Q6_K, plus trivial F32/F16/BF16",
so this looks like a deliberate gap rather than an oversight — but Q8_0 being half-declared
makes it read as supported, and it is the format that blocks the most widely distributed
Gemma-4 GGUFs.

## Impact

unsloth's "UD-…-XL" dynamic quants — the default download for most people — place Q8_0 on the
attention projections, the dense FFN, and `token_embd`. In
`unsloth/gemma-4-26B-A4B-it-GGUF` → `gemma-4-26B-A4B-it-UD-Q6_K_XL.gguf`, **237 of 658
tensors are Q8_0**:

| tensor role | type | count |
|---|---|---|
| `token_embd.weight` | Q8_0 | 1 |
| `blk.N.attn_q/k/v/output.weight` | Q8_0 | 115 |
| `blk.N.ffn_gate/up/down.weight` | Q8_0 | 90 |
| `blk.N.ffn_down_exps.weight` | Q8_0 | 30 |
| `blk.N.ffn_gate_up_exps.weight` | Q8_0 | 1 |
| `blk.N.ffn_gate_up_exps.weight` | Q6_K | 29 |
| norms, scales, router, `rope_freqs` | F32 | 392 |

So of the three Gemma-4 GGUFs a user is likely to already have on disk, the two `UD-*_XL`
ones are unloadable on quant type and the dense one is unloadable on config (filed separately).

## Suggested fix

Q8_0 is the simplest ggml block format — 32 int8 weights and one fp16 scale per 34-byte block
— so this should mirror `dequant_q4_0` closely:

```python
def dequant_q8_0(raw: torch.Tensor, out_dtype: torch.dtype) -> torch.Tensor:
    d = raw[:, :2].contiguous().view(torch.float16).to(torch.float32)   # [n,1]
    qs = raw[:, 2:].view(torch.int8).to(torch.float32)                  # [n,32]
    return (d * qs).to(out_dtype)
```

plus the `_DEQUANT[GGML_Q8_0] = dequant_q8_0` entry. `BLOCK_SHAPE` and `row_bytes` already
handle it. If Q8_0 is intentionally out of scope for the packed on-GPU expert path, then the
clearer fix is the opposite one — drop it from `BLOCK_SHAPE`/`GGML_NAME`/`__all__` and fail
early at load with a message naming the tensor and type, rather than a bare
`NotImplementedError` deep in the loader.

## Secondary: `expert_quant` is hardcoded for gemma4 GGUF

`models/gemma4/gguf.py` sets `expert_quant="q4_0"` unconditionally, but this checkpoint's
expert banks are Q6_K on 29 layers and Q8_0 on one. Even with a Q8_0 dequant available, the
expert bank loader would be told the wrong block layout (Q4_0 is 18 bytes/32, Q6_K is
210 bytes/256). Happy to split this out if you'd prefer it tracked separately.

## Relationship to existing issues

- #188 — embedding quant type hardcoded to `GGML_Q6_K`. Same class of bug, and this
  checkpoint would hit it too (`token_embd.weight` is Q8_0, not Q6_K). #188 fixes *reading
  the type from the file*; this issue is about the type then having no implementation.
- #194 — heterogeneous GGUF expert banks; related, since the mixed Q6_K/Q8_0 expert banks
  above are exactly that shape.

## Method / caveat

Found by parsing the GGUF KV + tensor-info tables directly and reading `dequant.py`; I have
**not** driven this checkpoint to a runtime traceback, because the same file also trips #188
at the embedding before the expert banks are reached. Reported as a static finding — the
tensor census above is reproducible from the file's header alone.

## Environment

- FreeToken 0.1.2 (PyPI wheel + `freetoken_kernel_cache-0.1.2+cu130`)
- RTX 5090 (sm_120, 32 GB), driver 616.56, CUDA 13

Cyber-Marty:
## Summary

`dequant.py` declared Q8_0 in `BLOCK_SHAPE`, `GGML_NAME` and `__all__`, but there was no
`dequant_q8_0`, so `dequantize()` raised `NotImplementedError` on the format that the most
widely distributed Gemma-4 GGUFs use: unsloth's `UD-*_XL` dynamic quants place Q8_0 on the
attention projections, the dense FFN and `token_embd` — **237 of 658 tensors** in
`gemma-4-26B-A4B-it-UD-Q6_K_XL.gguf` (see #358).

## Fix

`dequant_q8_0` follows the ggml `block_q8_0` layout: per 32-elem block, fp16 scale `d`
+ 32 int8 quants `q`, **`w = d*q` with no offset** (unlike Q4_0's `(q-8)*d` — ggml's
`quantize_row_q8_0` stores `q = round(w/d)`, `d = max|w|/127`).

Registered in `_DEQUANT` and `__all__`; module docstring updated to list Q8_0. No
existing code path changes — Q4_0 / Q6_K / F32 / F16 / BF16 untouched.

## Tests

New `tests/models/test_dequant_q8_0.py` (4 cases, pure torch / CPU):

- **round-trip**: random fp32 → reference quantizer (mirrors `quantize_row_q8_0`) →
  `dequant_q8_0` → reconstruction error bounded by half a quantization step;
- **exact values**: hand-computed block (`d=0.5`, quants incl. -128/127/0) decodes exactly;
- **dispatch**: `dequantize(raw, GGML_Q8_0)` routes instead of raising;
- **metadata honesty**: `BLOCK_SHAPE[Q8_0] == (32, 34)` and `row_bytes` math unchanged.

`tests/models/test_gemma4_gguf_config.py` + `test_dequant_q8_0.py` + `test_cpu_moe_q4_0.py`:
7 passed, 5 skipped (CUDA-only cases, by design).

Fixes #358

### DIFF
--- a/python/freetoken/models/gemma4/gguf.py
+++ b/python/freetoken/models/gemma4/gguf.py
@@ -74,6 +74,14 @@ def parse_gguf_config(shim: "GgufConfigShim") -> ModelConfig:
     full_kv = int(kv_per_layer[full_layer_ids[0]]) if full_layer_ids else int(kv_per_layer[0])
 
     max_pos = int(g("context_length"))
+    # Dense Gemma-4 checkpoints (gemma-4-12B-it, gemma-4-31B-it) carry no expert fields
+    # in their GGUF metadata; only MoE members (e.g. Gemma-4-26B-A4B) do. Mirror the HF
+    # side (gemma4.config.parse_config): default the expert fields to 0 and derive
+    # moe_enabled from them instead of assuming every GGUF is MoE.
+    num_experts = int(m.get("gemma4.expert_count", 0) or 0)
+    num_experts_per_tok = int(m.get("gemma4.expert_used_count", 0) or 0)
+    moe_intermediate_size = int(m.get("gemma4.expert_feed_forward_length", 0) or 0)
+    moe_enabled = num_experts > 0
     full_rotary = RotaryConfig(
         head_dim=full_head_dim,
         rotary_dim=_full_rotary_dim(shim, full_head_dim),
@@ -105,15 +113,16 @@ def parse_gguf_config(shim: "GgufConfigShim") -> ModelConfig:
         rms_norm_eps=float(g("attention.layer_norm_rms_epsilon")),
         tie_word_embeddings=bool(shim.tie_word_embeddings),
         rotary_config=full_rotary,
-        num_experts=int(g("expert_count")),
-        num_experts_per_tok=int(g("expert_used_count")),
-        moe_intermediate_size=int(g("expert_feed_forward_length")),
+        num_experts=num_experts,
+        num_experts_per_tok=num_experts_per_tok,
+        moe_intermediate_size=moe_intermediate_size,
         norm_topk_prob=True,
         model_type="gemma4",
         architectures=list(shim.architectures),
-        moe_enabled=True,
-        expert_quant="q4_0",
-        moe_weight_format="q4_0",
+        moe_enabled=moe_enabled,
+        # Native-Q4_0 offload-cache path for the routed experts (MoE checkpoints only).
+        expert_quant="q4_0" if moe_enabled else "none",
+        moe_weight_format="q4_0" if moe_enabled else "none",
         use_qk_norm=True,
         attn_sm_scale=1.0,
         final_logit_softcapping=float(g("final_logit_softcapping")),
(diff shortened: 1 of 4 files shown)


Cyber-Marty:
Closing — this branch accidentally carried the unmerged #359 commits (branched from the wrong base). Reopening from a clean main-based branch with only the Q8_0 changes.

PR #367, opened 2026-09-03
fix(scheduler): reserve paged KV at allocation granularity

taking-lying-flat:
## Summary

- Account prefill KV reservations in whole pages per request.
- Use the same page-ceiling span as CacheManager.allocate_paged() for both admission checks and cumulative reservations.
- Keep requests queued when the remaining KV pool cannot provide a distinct page for each request.

## Problem

CacheManager.available_size reports token-equivalent capacity, but the allocator assigns whole pages independently to each request. PrefillAdder previously charged raw token counts, so multiple short requests could pass admission even when they required more physical pages than the pool contained. The subsequent allocate_paged() call could then fail during eviction or hit the "Eviction did not free enough space" assertion.

For example, with a 128-token page and one free page, two requests that each need two tokens were charged as four tokens total even though they require two separate pages.

## Validation

- 51 relevant scheduler tests passed across radix, chunked prefill, hybrid, SWA page-size, DSV4, and abort paths.
- Ruff check passed for python/freetoken/scheduler/prefill.py.
- git diff --check passed.

### DIFF
--- a/python/freetoken/scheduler/prefill.py
+++ b/python/freetoken/scheduler/prefill.py
@@ -48,6 +48,13 @@ class PrefillAdder:
     # across the admission loop -- without this, successive admits all see the full pool.
     reserved_swa: int = 0
 
+    def _kv_reservation_size(self, total_len: int, cached_len: int) -> int:
+        """Return the token-equivalent cost of the additional KV pages for a request."""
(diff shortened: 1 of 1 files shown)


jason-fxz:
LGTM, thanks for the fix.

Issue #349, opened 2026-09-02
Qwen3.8-Flash-Next: QSA prefill kernel exceeds the 64 KiB LDS budget on RDNA, capping prompts at 15 tokens


lukascechovic:
### Summary

On RDNA (`gfx1201`, Radeon AI PRO R9700, Linux), `qwen4_exp` serves **prompts of at most 15 tokens**.
A 16-token prompt raises `triton.runtime.errors.OutOfResources`, **kills the worker, and stops the
API server** — so no `/v1/chat/completions` request can succeed at all, because the chat template
alone exceeds 15 tokens.

The cause is a tile profile in `kernel/triton/qsa/attend.py` that is commented *"Tuned on GB300"* and
asks for more shared memory than any RDNA part has. A 10-line HIP-gated clamp fixes it. Decode is
unaffected either way; output on the previously-working path is byte-identical.

### The arithmetic

`_qsa_sparse_paged_gqa_splitk_kernel` (`kernel/triton/qsa/attend.py:302`) requests **65,792 B** of
LDS. gfx1201 has **65,536 B** per workgroup.

At this checkpoint `head_dim = 256`, bf16. The K tile `[head_dim, BLOCK_N]` and the V tile
`[BLOCK_N, head_dim]` alone cost `2 * head_dim * block_n * itemsize`:

```
2 * 256 * 64 * 2 = 65,536 B      <- the entire budget, before any index scratch
```

So at `block_n = 64` the pair exactly fills LDS and **any** addition overflows. The remaining 256 B
is index scratch we could name but not isolate.

### Why it presents as a 15-token limit

`block_n` is chosen from `base_programs = tokens * num_kv_heads`. With `num_kv_heads = 2`, a prompt of
≤ 15 tokens gives `base_programs < 32` and the kernel takes the `block_n = 16` path (16,384 B), which
fits. At 16 tokens it takes `block_n = 64` and dies.

⚠️ We predicted the 15/16 boundary from this arithmetic **before** testing it, then confirmed it.

✅ Decode never hits this — a decode step is one token, so `base_programs = 2` and decode already runs
at `block_n = 16` on every platform.

### Reproducer

Any `qwen4_exp` checkpoint on an RDNA card, ROCm, `ft serve`:

```bash
# 15 tokens: serves
curl -s localhost:1919/v1/completions -H 'content-type: application/json' \
[...]

PR #378, opened 2026-09-04
fix(rocm): make CPU/Hybrid MoE graph replay safe

### LINKED ISSUE #350: [ROCm] CpuMoeExecutor decode is silently wrong under CUDA-graph replay (rel err 1.36, no exception)
### Summary

On ROCm, `--moe-backend cpu` and `--moe-backend hybrid` produce **numerically unrelated output**
during CUDA-graph replay, with **no exception and no warning**. Relative error is **1.36** — that is
not drift, the answer is unrelated to the correct one. Eager is correct. `--moe-backend offload` is
unaffected.

We are not running these backends in production (we serve `offload`), which is exactly why we would
rather report it than sit on it: someone who does select `cpu` or `hybrid` on an AMD card today gets
wrong answers and nothing tells them.

### Reproducer

Four tests in the existing suite already catch it — they fail on ROCm and, as far as we can tell,
have never been run on an AMD card in CI:

```
tests/moe/test_cpu_moe.py            3 x *_cuda_graph_replay*
tests/moe/test_cpu_moe_q4_0.py       1 x *_cuda_graph_replay*
```

```
AssertionError: replay 0 rel err 1.3613485097885132
```

Reproduced on **all four expert formats**: bf16, mxfp4, ds_fp4, q4_0.

### Mechanism, as far as we traced it

`CpuMoeExecutor.__init__` runs a functional probe, `_cpu_moe.memops_probe(...)`, for CUDA stream
memory operations. On ROCm the probe returns **False** and the executor logs:

```
INFO  cpu-moe flag handshake unavailable: CUDA stream memory operations are not supported here
      (Windows WDDM / vGPU / old driver); using the cudaLaunchHostFunc sync
```

It then takes the `cudaLaunchHostFunc` fallback — the same path `FREETOKEN_CPU_MOE_FLAG_SYNC=0`
selects deliberately. **That fallback is not capture-safe.** The captured host nodes do not re-read
the freshly written pinned routing buffers, so every replay recomputes from the **capture-time**
values.

Two things make it silent rather than loud:

- ⛔ It does not raise. The output is simply wrong.
- ⛔ `raise_if_unhealthy()` guards the **flag-sync** path — precisely the path ROCm cannot take. The
  watchdog covers everything except the branch that is broken here.

The log line is also misleading on AMD: it attributes the probe failure to *"Windows WDDM / vGPU /
old driver"*, which sends a ROCm user looking in the wrong place.

### Suggested directions

We have not written a fix; the maintainers will know which is right. Three options as we see them:

1. Make the `cudaLaunchHostFunc` path refuse capture — raise at capture time rather than produce
   wrong answers at replay time. This is the smallest change and turns a silent failure into a loud
   one.
2. Extend `raise_if_unhealthy()` to cover the host-func path.
3. Make it capture-safe, presumably the way #316 does for the fused expert copy: use a path whose
   source is a direct tensor argument during capture.

⚠️ Even option 1 alone would be a real improvement — the current behaviour is the worst kind.

### Relationship to #316

#316 fixes a HIP-graph-capture fault in `moe/offload_cache.py`, where the fused multi-bank expert
copy reads pinned-host source addresses indirectly through a device pointer table. Same family of
problem — host memory and HIP graph capture — but a different file, a different code path, and a
different symptom: #316's presents as a fault, this one presents as a confident wrong answer. We do
not believe #316's fix covers this, but the maintainers should check that judgement.

---

## Two smaller findings, from the same run

⚠️ These are minor. Fold them in here or split them — whichever is easier to triage.

### `qsa_block_topk`'s wide-row split path is inexact under ties

`kernel/triton/qsa/topk.py`'s module docstring justifies the wide-row **split** path as exact:
*"a global winner beats at most k−1 columns, so it beats all but at most k−1 of its own slice — so the
answer stays exact."* That argument needs **strict** inequality and does not survive ties.

Consistent with that reading, only the split shape fails: `n_blocks=4096` (≥ `_MIN_CHUNK`) fails,
`n_blocks=512` (single tile, no split) passes.

✅ **Harmless in practice, and we want to be clear about that.** On continuous input the kernel is
exactly `torch.topk` — 0 of 2048 mismatched. Under the synthetic tie case the differing picks all
have **identical scores** to the picks they replaced, and the selected-score sums are exactly equal.
The failing test's own comment concedes *"Selection is a set"*.

⚠️ **This is probably not AMD-specific at all**, and that is why it is worth your time rather than
ours: the hole in the split-exactness argument is architecture-independent. We have no NVIDIA control
to check it against, and if upstream CI runs release wheels only, this test may never have executed
on any GPU.

⚠️ One thing we could not explain and are recording rather than asserting: the mismatch count varies
**across processes** on identical seeded input (30, 39, 46, 62, 82 of 2048 observed), while 20 repeats
**inside** one process gave a single result. We did not separate that mechanism.

### `FREETOKEN_CACHE_DIR` is read by nothing

Your own Dockerfile sets `FREETOKEN_CACHE_DIR`, but grep over the tree finds no reader. The live
variables are `XDG_CACHE_HOME` (the `ft bench bw` profile, `moe/bench_profile.py:35`) and
`FREETOKEN_KERNEL_CACHE_DIR` (the tvm-ffi kernel cache, `kernel/utils.py:15`).

Empirically: after a full suite run, our mounted `triton` cache dir held 120 MB and the mounted
`freetoken` dir was **empty**. Anyone mounting a cache by that name silently loses it. A one-line
docs/Dockerfile fix.

---

## Platform

- AMD Radeon AI PRO R9700 (`gfx1201`), Linux, one card of two used
- FreeToken `main@4b94bdc3` + PR **#132** + **#133** + **#134**
- PyTorch 2.11.0+rocm7.14.0, HIP 7.14.60850
- Full kernel suite: **1471 passed, 27 failed, 78 skipped** in 290 s. The 27 are classified in our
  notes; 16 are the known `dsv4_sparse_attn` Triton lowering failure the gfx1100 report also saw, 3
  are missing FlashInfer (CUDA-only), 4 are the bug above, 3 are `qwen4_exp` tolerance asserts we
  probed to mechanisms and believe are **not** wrong answers, and 1 is a top-k tie in a Gemma-4
  fallback path.

zihaomu:
## Summary

Fixes #350, where CPU/Hybrid MoE produces silently incorrect output during ROCm graph replay while eager execution remains correct.

- fail closed on ROCm: if the native stream-memory handshake does not pass a real capture/instantiate/replay probe, disable graph capture and continue on the correct eager path
- add the ROCm 7.14 native handshake with `hipMallocSignalMemory` and explicit graph batch-memory-op nodes
- keep HIP signal and graph-node parameter storage owned by `CpuMoeExecutor`, so their lifetime covers every graph replay without module-global growth
- size per-layer flag capacity from the actual configured graph batch sizes, including configurations with more than 16 sizes
- preserve the established CUDA module-level `memop_submit`/`memop_sync` path to limit NVIDIA regression risk
- document the CUDA/ROCm synchronization difference and the ROCm fallback behavior

## Dependency and scope

Depends on #132. This branch is based directly on the current #132 head (`c0713e4`). #132 itself is unchanged.

This is intentionally a Draft stacked PR while the native ROCm graph path receives RDNA3 and long-running model-serving stability validation. Until #132 merges, GitHub will also show the dependency commits in this PR; the diff will collapse to this single follow-up commit after the base lands.

CUDA and ROCm retain the same externally visible CPU-MoE ordering and results. Only their GPU/CPU synchronization implementations differ:

| Backend | Handshake implementation |
|---|---|
| CUDA | Existing mapped-pinned per-slot arrays through module-level `cuStreamWriteValue64`/`cuStreamWaitValue64` wrappers |
| ROCm 7.14 | Executor-owned HIP signal memory plus explicit graph batch-memory-op nodes |

## Validation

- PyTorch 2.11.0 + ROCm 7.14 on one Radeon AI PRO R9700: targeted safety and replay suite, **21 passed**
- capture/replay exercised across **19 graph batch sizes**, covering the former fixed-16-slot boundary
- eager plus repeated graph replay covered bf16, MXFP4, DS-FP4, and GGUF Q4_0 CPU-MoE formats
- ROCm 7.2.4 / PyTorch 2.10 compatibility build passed and remains fail-closed where graph batch-memory-op APIs are unavailable
- CUDA module API and dispatch routing are covered by unit tests; NVIDIA hardware validation remains required before this Draft is marked ready
- `git diff --check` passes

## Draft exit criteria

- reproduce the replay suite on RDNA3
- complete repeated graph rebuild/replay and long-running serving stability checks
- obtain NVIDIA CI coverage for the preserved CUDA path


lukascechovic:
Thanks for building this — we filed #350 and did not expect a fix, let alone one validated on the
same card we found it on. Your framing of the fix as **fail-closed** — disable capture and
continue on the correct eager path when the native handshake does not pass a real
capture/instantiate/replay probe — is the property we would have asked for and did not know how to
ask for. A silent wrong answer converted into a slower right one is exactly the right trade for
this class of bug.

Two things below: a ROCm datapoint your HIP memop work bears on directly, and an offer.

### 1. Your HIP memop port has a second consumer, and it is the half that is still private

Reading the diff, the thing that struck us is that the HIP `hipStreamWriteValue64` /
`hipStreamWaitValue64` / `hipStreamBatchMemOp` port is **already here** — we had it filed on our
own side as a piece of work someone would eventually have to do. It matters to us because the CPU
MoE executor is not the only thing in this engine that wants that handshake.

#311's disk-streamed PLE table resolves its stream memops by `dlopen`ing **`libcuda.so.1`**. On
ROCm that dlopen fails, the probe returns false, and the table falls back to **launch-gating**: a
blocking D2H of the sampled token that drains the stream, then hash, io_uring read, H2D and launch
on the host, with the GPU idle across the whole stretch. Measured on our box — **half-box** (one
R9700, TP=1), on tree `c4b96fc1` (upstream `4b94bdc3` plus our patches 0001..0009), with
`--ple-backend` the only flag changed between the arms:

| arm | decode at 10k | decode at 100k |
|---|---:|---:|
| `--ple-backend pinned` | **26.100 ± 0.351** tok/s (n=8/10) | 26.345 ± 0.110 (n=2/3) |
| `--ple-backend disk` | **18.984 ± 0.272** tok/s (n=9/10) | 19.432 ± 0.182 (n=2/3) |

`n` is post-filter: replies are dropped if their zlib compression ratio falls below 0.40, which is
how we catch a degenerate repeated-token generation before it flatters a decode number. Note the
100k pair rests on **two samples per arm** — we quote it because it agrees with the 10k pair, not
as an independent result.

**−27.3 % at 10k, −26.2 % at 100k**, roughly **14 ms per token**, against the **−2.6 % … +0.2 %**
band #311 reports on H100 and RTX PRO 6000 Blackwell — which was measured on the **wait-sync**
path. Prefill is unaffected: TTFT **+0.20 s** at 10k and **−0.22 s** at 100k, same half-box arms —
though every prefill number from this box carries *"gfx1201, `BLOCK_N` clamped to 32"* (see
Platform), and `block_n = 64` does not run here at all — it raises `OutOfResources` before
executing an instruction — so the clamp's cost is unmeasurable rather than zero. The memory win is
real and large — on the same arms the PLE table stops pinning **47.7 GiB** of host RAM — which is
why we run it anyway.

We can only **bound** the split, not measure it. Sampled from `/proc/diskstats` on the same
half-box `disk` arm, during decode the NVMe served 300 reads/s at 0.648 ms average = 16 reads per
token, one 4K `O_DIRECT` read per row. Fully serialised that is 10.4 ms; if the store's io_uring
batch puts all 16 in flight together it is ~0.65 ms. `/proc/diskstats` cannot tell those apart, so
the disk accounts for **0.65–10.4 ms** and the launch-gating stall for the remaining **4–13.7
ms**. What the bound does establish is that the disk alone cannot reach 14 ms even in its worst
case. Splitting it properly needs a wall-clock inside the fill, which we have not written — so
treat the split as bounded, not measured.

Where your PR lands on it: the **eager** half is now reachable — on ROCm your module-level
`memops_probe` / `memop_submit` / `memop_sync` resolve to the HIP implementations, so a second
consumer could stop dlopening `libcuda` and just call them. The half that is still
`CpuMoeExecutor`-owned is the **capture** half — and that is the half #311 needs, since by its own
description the decode graph launches first and waits on the flag right before consuming the rows.
If the ROCm graph-node splicing ever becomes shareable, there is a caller waiting for it.

And the ROCm-7.14 fact you documented in that file — that ordinary `hipStreamWrite`/`WaitValue`
and `hipStreamBatchMemOp` calls **execute during capture but are not recorded in the graph**, so
you add explicit batch-memory-op nodes and splice them into the dependency set — is worth flagging
loudly to anyone attempting a second port. Someone who ports the eager calls, sees them work, and
captures a graph gets silently stale values on replay: the same failure shape as #350 itself, in a
different file.

None of this is a request. It is a note that the two land in the same place.

### 2. An offer against your draft-exit criteria

You list *"reproduce the replay suite on RDNA3"*, *"complete repeated graph rebuild/replay and
long-running serving stability checks"*, and NVIDIA CI for the preserved CUDA path.

We can help with parts of the middle one, on **RDNA4** rather than RDNA3. We run **two** Radeon AI
PRO R9700 (`gfx1201`) on Linux under ROCm 7.14 / PyTorch 2.11, serving a 100 GiB-class MoE
continuously, and we can run your targeted safety and replay suite
(`tests/engine/test_cpu_moe_graph_safety.py` and the `tests/moe/` graph-replay cases) on both
cards and report the numbers the way we did on #132.

We can also exercise it at **TP=2**. To be exact about what is new there: we have reported TP=2
serving before, on #135, so the gap is not TP=2 as such — it is that nobody has run the
**CPU/Hybrid graph-replay** path at TP>1, on any card. That is worth closing, because from our own
experience a caught per-rank exception behaves very differently at TP=2 than at TP=1. That
difference is a scheduler problem rather than yours, and we have written it up separately.

Three honest caveats before you count on any of it.

**We cannot give you the "long-running serving stability" half.** We serve `--moe-backend
offload`, not `cpu` or `hybrid`, so a serving soak on this code path is not something our
production traffic would ever produce — it would be a rig we stand up specially, and it would not
be representative of a real workload the way our `offload` row is. Repeated rebuild/replay and the
suite are the parts we can do well.

**We cannot promise a date.** The box is a shared machine with a live soak on it. Say what you
want run and we will fit it in.

**It is a second gfx1201 datapoint, not a clean-`main` CI result.** RDNA4, on a tree that is
upstream `4b94bdc3` plus a local patch series, with the `BLOCK_N` clamp in the Platform note
below.

No quality claim from us in any case — we have no fidelity instrument on this box, which is why
#350 was reported as a numerical-deviation finding with a relative error rather than as a quality
regression.

---

### Platform

2 × AMD Radeon AI PRO R9700 (`gfx1201`, 32 GiB each), Linux, ROCm 7.14, PyTorch 2.11, 6 physical
cores / 122 GiB host RAM. Serving `qwen4_exp` / Qwen3.8-Flash-Next NVFP4 with `--moe-backend
offload`; the CPU/Hybrid finding in #350 came from running this repo's kernel suite on gfx1201,
not from this serving configuration.

Our tree is upstream `4b94bdc3` plus a local patch series, whose build asserts tree **`c4b96fc1`**
for the 0001..0009 stack every §1 number was measured on. The QSA prefill tile is halved until `2
· head_dim · block_n · itemsize` fits a 60 KiB LDS budget (#349) — at this checkpoint's `head_dim
= 256` in bf16 that lands on `BLOCK_N = 32`. So every prefill number here is on the clamped tile
and is not comparable to a CUDA run of the same engine.

**Box labelling, since this box has two cards:** **every number in §1 — decode, TTFT, the 47.7 GiB
and the `/proc/diskstats` figures alike — is half-box**, one R9700 at TP=1, on tree `c4b96fc1`
(upstream `4b94bdc3` plus our patches 0001..0009), with `--ple-backend` the only flag changed
between the arms. Anything we quote at `--tp-size 2` is **full-box**. Every number is named
against the tree it was measured on, and none of them is measured on upstream `main`.

zihaomu:
Rebased this PR onto the refreshed AMD foundation (#132, head `c2af4c6`) and updated it for the current `moe_strategy` / quantization interfaces.

Conflict resolution kept the current upstream Engine/CpuMoeExecutor contracts (`sample.alpha`, `sample.limit`, `quant_method.cpu_format`) while layering the ROCm graph-safe HIP signal/memop path on top.

Validation:
- `tests/engine/test_cpu_moe_graph_safety.py`: 14 passed
- CPU MoE suites: 28 GPU-dependent cases collected/skipped in the no-GPU test container
- all native extensions, including `_cpu_moe`, compile with ROCm 7.14 modular SDK
- `git diff --check` passes

New head: `94f77248886029c8c0072960c3d00a901b60deaa`.

Issue #29, opened 2026-08-22
Tensor parallelism (TP>1) unsupported for Qwen3.5/3.6 MoE: weight loader gated to TP=1


alogotron:
### Summary

`--tensor-parallel-size 2` fails for Qwen3.5/3.6 MoE checkpoints because the weight loader is gated to TP=1 at every entry point. Since Qwen3.6-35B-A3B is the model featured in the launch announcement, multi-GPU users hit this immediately.

This provides the technical detail behind #15.

### Reproduction

```bash
ft serve --tensor-parallel-size 2 --model Qwen/Qwen3.6-35B-A3B-FP8
```

```
ERROR    Backend supervisor: NotImplementedError: qwen3_5_moe weight loading currently supports TP=1 only
ERROR    Backend worker is gone and cannot be restarted; stopping the API server
```

Startup is otherwise clean — `DistributedInfo(rank=0, size=2)` is parsed correctly and PyNCCL initialises. The failure occurs at weight load.

### Location

`freetoken/models/qwen3_5_moe/weight.py` raises at four sites, covering every loader path:

| Line | Function |
|---|---|
| 214 | `iter_weights` |
| 443 | `_iter_weights_attn_fp8` |
| 578 | `_iter_weights_compressed_tensors` |
| 678 | `iter_weights_parallel` |

```python
tp_info = get_tp_info()
if tp_info.size > 1:
    raise NotImplementedError("qwen3_5_moe weight loading currently supports TP=1 only")
```

No quantisation format bypasses the gate. Verified with both `Qwen/Qwen3.6-35B-A3B-FP8` and `nvidia/Qwen3.6-35B-A3B-NVFP4`.

### Scope: per-family gap, not an engine limitation

The TP machinery is present and used by most families:

| TP=1 restricted | Supports sharding |
|---|---|
| `qwen3_5_moe` | `deepseek_v4`, `glm4_moe`, `glm_moe_dsa` |
| `gemma4` | `gpt_oss`, `llama`, `mistral` |
| `muse_glimmer` | `qwen2`, `qwen3`, `qwen3_moe`, `minimax_m2` |
| `minimax_m3` (attention assert) | |

Notably `qwen3_moe` supports sharding while `qwen3_5_moe` does not, which suggests the sharded loader simply has not been ported to the newer family rather than being blocked by an architectural constraint.

### Impact

[...]

PR #385, opened 2026-09-04
feat(qwen4_exp): tensor parallelism for Qwen3.8-Flash-Next (offload backend)

gdevenyi:
## What this adds

`ft serve --tp-size 2` for `qwen4_exp` (Qwen3.8-Flash-Next) on the offload MoE backend. Upstream refuses TP>1 for this architecture; with two 48 GiB cards the model runs as two independent TP=1 instances, each keeping 37% of the NVFP4 experts resident and streaming the rest over PCIe every step.

Per rank, the patch:

- **Shards the dense weights at load** (`weight.py::_shard`): attention `qkv_proj` by head (`[q|gate]` per head; kv heads split, or replicated when there are fewer than ranks), GDN `in_proj` as its six parts (`[q | k | v | z | b | a]`) plus the matching `conv1d` channels and `A_log` / `dt_bias`, shared-expert `gate_up_proj` per part; `o_proj`, `out_proj` and the shared `down_proj` row-parallel with the all-reduce inside; `embed_tokens` / `lm_head` by vocab rows. Router, QSA indexer, norms, hyper-connections and PLE stay replicated, so every rank selects the same sparse blocks and n-gram rows.
- **Shards the NVFP4 expert banks along the intermediate axis** (`nvfp4_banks.py`, I=640 -> 320 per rank): packed codes, the 16-wide scale blocks and the per-row globals for the gate/up rows and the down columns. The offload cache then holds half the experts per rank.
- **One all-reduce per MoE layer** (`moe.py`): the routed and shared partial sums are combined as `routed + sigmoid(gate) * shared` before a single reduce, instead of one reduce each.
- `LinearColParallelMerged(local_output_sizes=)` for the kv-replicated case (the same shape as #104's hunk), and `distributed_timeout` 60 s -> 1800 s: behind a 100+ GiB load the ranks reach their first collective minutes apart, and 60 s kills the launch.

Everything else in the engine (scheduler, KV pool, PLE table, CUDA graphs) is untouched.

## Measurements

2 x RTX 6000 Ada (48 GiB, sm_89, PCIe Gen4 x16, no NVLink), 2 x Xeon Gold 6526Y, 503 GiB RAM. `RadixArk/Qwen3.8-Flash-Next-NVFP4`, `--moe-backend offload --ple-backend pinned --num-tokens 262144 --memory-ratio 0.94 --moe-prefill-hit-d2d`.

| run (both GPUs, KV 262,144 tokens) | single-stream tok/s | 8 concurrent tok/s | expert residency | TTFT, 1k prompt |
|---|---:|---:|---:|---:|
| two TP=1 instances (previous layout) | 55.9 / 56.7 | 266.1 aggregate | 36.6% each | 2.05 s |
| **TP=2, 8 running requests** | **88.6** | **298.8** | 100% | 1.55 s |
| TP=2, 16 running / graph bs 16 | 89.8 | 324.5 | 94.5% | 0.84 s |

A 262,144-token prompt reaches its first token in 74 s at TP=2 (116 s at TP=1). The decode step is dominated by the bf16 dense read; TP=2 halves it per GPU and removes the PCIe expert gather because everything fits.

**Correctness.** An 8-question probe gives identical answers at TP=1 and TP=2. Three raw prompts decoded greedily for 256 tokens: the ~1k-token prompt (QSA over many blocks, GDN state, PLE context) is word-for-word identical across every run; the two short prompts diverge after 13 and 29 words between TP=1 and TP=2, but the TP=2 server diverges from *itself* at the same points on a second pass (bf16 atomics in the expert kernels), so that is run-to-run noise rather than a sharding error. Greedy output at TP=2 is not bit-exact between passes.

## Limits

Offload backend with bf16 dense projections only: `fp8_block` / `nvfp4` dense checkpoints raise under TP (row-parallel FP8 / NVFP4 linears do not exist yet, the same gap #104 has). The hybrid / CPU MoE backends are not sharded.

Related: #62, #29 (TP for offloaded MoE), #104 (TP for qwen3_5_moe, which this reuses the merged-linear hunk from).

## Testing

- `tests/models/qwen4_exp/test_tp_shard.py`: the per-head / per-part row sharding of every fused projection reassembles to the original (CPU).
- `tests/models/test_nvfp4_banks_tp.py`: the bank placer's per-rank slices of codes, scales and globals cover the intermediate axis exactly once (CPU).
- Served for a day on the machine above at TP=2 with the numbers in the table. `tests/models/qwen4_exp` on one of its GPUs: 97 passed, 3 failed; the same 3 (`test_chunked_prefill_matches_one_shot[*]`, a bit-exact assertion off by bf16 noise on this torch 2.11 / flashinfer 0.6.18 / triton 3.6 stack) fail on plain `main` there too.

### DIFF
--- a/python/freetoken/engine/config.py
+++ b/python/freetoken/engine/config.py
@@ -79,7 +79,7 @@ class EngineConfig:
     # ratio default above. A runtime cache rebuild sets this (num_swa_pages) to pin the window
     # regardless of the full anchor; the ratio is the startup default and the fallback.
     swa_num_pages_override: int | None = None
-    distributed_timeout: float = 60.0
+    distributed_timeout: float = 1800.0  # ranks reach the first collective minutes apart on a 100+ GiB offload load
     use_dummy_weight: bool = False
     use_pynccl: bool = True
     max_seq_len_override: int | None = None
--- a/python/freetoken/layers/linear.py
+++ b/python/freetoken/layers/linear.py
@@ -83,13 +83,17 @@ class LinearColParallelMerged(_LinearTPImpl):
         input_size: int,
         output_sizes: List[int],
         has_bias: bool,
+        local_output_sizes: List[int] | None = None,
         *,
         quant_config: QuantConfig | None = None,
         prefix: str = "",
     ):
-        # check that all output sizes are divisible by tp_size
+        # check that all output sizes are divisible by tp_size (a caller that replicates
+        # GQA kv heads across ranks passes the per-rank sizes explicitly)
         tp_info = get_tp_info()
-        tp_output_sizes = [div_even(size, tp_info.size) for size in output_sizes]
+        if local_output_sizes is None:
+            local_output_sizes = [div_even(size, tp_info.size) for size in output_sizes]
+        tp_output_sizes = local_output_sizes
         output_size = sum(output_sizes)
         tp_output_size = sum(tp_output_sizes)
         super().__init__(
--- a/python/freetoken/layers/quantization/moe/nvfp4.py
+++ b/python/freetoken/layers/quantization/moe/nvfp4.py
@@ -41,7 +41,13 @@ class TritonNvfp4MoEKernel(MoEKernel):
     cpu_format = "nvfp4"
 
     def unusable_reason(self, cfg: MoEConfig) -> str | None:
-        reason = self._common_reject(cfg, resident_ok=False, tp_ok=False, cpu_ok=True, plain_silu_only=False)
+        # tp_ok: the source stream is sliced along the intermediate axis per rank
+        # (nvfp4_banks._tp_shard) and this kernel sizes its banks from
+        # cfg.local_intermediate, so a rank holds and reads only its own half. The routed
+        # output is then a partial sum, which the MoE layer reduces (_maybe_all_reduce, or
+        # one combined all-reduce in qwen4_exp's block). marlin and b12x stay off: their
+        # pack() repacks the rows and neither has been checked against a sharded bank.
+        reason = self._common_reject(cfg, resident_ok=False, tp_ok=True, cpu_ok=True, plain_silu_only=False)
         if reason:
             return reason
         reason = gated_epilogue_reason(cfg)
--- a/python/freetoken/layers/quantization/moe/nvfp4.py
+++ b/python/freetoken/layers/quantization/moe/nvfp4.py
@@ -48,7 +48,7 @@ class TritonNvfp4MoEKernel(MoEKernel):
         return f"triton nvfp4 MoE kernel: {reason}" if reason else None
 
     def layout(self, cfg: MoEConfig) -> dict[str, BankSpec]:
-        i, h = cfg.intermediate, cfg.hidden
+        i, h = cfg.local_intermediate, cfg.hidden
         return {
             "gate_up": BankSpec((2 * i, h // 2), torch.uint8),
             "gate_up_scale": BankSpec((2 * i, h // GROUP), FP8),
@@ -61,7 +61,7 @@ class TritonNvfp4MoEKernel(MoEKernel):
     def pack(self, pieces, cfg: MoEConfig, out):
         out["gate_up"].copy_(fused_piece(pieces, "gate_up"))
         out["gate_up_scale"].copy_(fused_piece(pieces, "gate_up_scale"))
-        out["gate_up_global"].copy_(fused_global(pieces, cfg.intermediate))
+        out["gate_up_global"].copy_(fused_global(pieces, cfg.local_intermediate))
         out["down"].copy_(pieces["down"])
         out["down_scale"].copy_(pieces["down_scale"])
         out["down_global"].copy_(global_rows(pieces["down_global"], cfg.hidden))
@@ -241,7 +241,7 @@ class MarlinNvfp4MoEKernel(MoEKernel):
         return (8, 0) <= backend.device_capability() < (10, 0)
 
     def layout(self, cfg: MoEConfig) -> dict[str, BankSpec]:
-        i, h = cfg.intermediate, cfg.hidden
+        i, h = cfg.local_intermediate, cfg.hidden
         return {
             "gate_up": BankSpec((h // GROUP, 4 * i), torch.int32),
             "gate_up_scale": BankSpec((h // GROUP, 2 * i), FP8),
@@ -252,7 +252,7 @@ class MarlinNvfp4MoEKernel(MoEKernel):
         }
 
     def pack(self, pieces, cfg: MoEConfig, out):
-        i, h = cfg.intermediate, cfg.hidden
+        i, h = cfg.local_intermediate, cfg.hidden
         device = torch.device("cuda")
         gu, gus, gug = fused_piece(pieces, "gate_up"), fused_piece(pieces, "gate_up_scale"), fused_global(pieces, i)
         dn, dns, dng = pieces["down"], pieces["down_scale"], global_rows(pieces["down_global"], h)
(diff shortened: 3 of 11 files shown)


gdevenyi:
`tests/models/qwen4_exp` on a RTX 6000 Ada with this branch: 97 passed, 3 failed; plain `main` (af71ba4) on the same GPU: 94 passed, the same 3 failed (`test_qsa_backend.py::test_chunked_prefill_matches_one_shot[*]`, a bit-exact chunked-vs-one-shot assertion that is off by bf16 noise with torch 2.11.0+cu130 / flashinfer 0.6.18 / triton 3.6.0). So nothing here regresses the package; the second commit fixes the one regression the first had (`test_weight.py` feeds `iter_weights` a synthetic checkpoint without `model_type`, and the loader now reads the config only when TP > 1).

gdevenyi:
Rebased onto `fb7f732` (main after #418 / #427 / #426). The model-side half rebased; **the expert-bank half was a rewrite**, and the new version is smaller and more general, so it is worth describing rather than just re-pushing.

**What changed and why.** The original sharded inside `load_nvfp4_expert_source_banks` with a local `_tp_slice` + `_Placer` that wrote each checkpoint tensor into its bank slot. That function no longer exists: #418 turned bank building into `model -> expert piece stream -> moe/expert_banks.build_expert_banks -> the kernel's pack`, and `nvfp4_banks.py` now only yields pieces.

So the sharding moved to the two places the new architecture actually owns:

1. `nvfp4_banks._tp_shard` slices each checkpoint tensor to this rank's slice of the expert intermediate axis as it streams (gate/up on the row axis, down on the column axis — halved for the packed FP4 codes, sixteenthed for the fp8 block scales; the per-tensor `_global` has no I axis and is kept whole).
2. `layers/quantization/moe/nvfp4.py` sizes its banks from `cfg.local_intermediate` instead of `cfg.intermediate`.

**`MoEConfig` already had `tp_rank`, `tp_size` and `local_intermediate`** — the NVFP4 method simply was not using them (`unquantized` and `mxfp4` already do). So this is less new machinery than the original PR, and because it sits in the shared method rather than in qwen4_exp, it now applies to any NVFP4 model on the offload path rather than just this one. At TP=1 `local_intermediate == intermediate`, so nothing changes.

`tests/models/test_nvfp4_banks_tp.py` was rewritten against that contract: for each role and tp_size in {1,2,4}, the ranks' slices must concatenate back to the original tensor exactly, on the correct axis — which is the property the all-reduce depends on. It also asserts a shard that would split a 16-wide scale block is rejected, and that `MoEConfig.local_intermediate` agrees with the slice.

**Two other resolutions worth flagging:**

- `o_proj` stays `LinearOProj`. Main uses `LinearReplicated`, which is right today and wrong the moment this lands: `qkv_proj` is column-parallel, so each rank's attention output is its own head slice and `o_proj` has to take the sharded input dim and all-reduce. At TP=1 the two are identical (`div_even(x, 1) == x`, reduction skipped). Same point as my note on #392.
- The GDN `in_proj` guard is now phrased against #426's scheme test (`quant_config.scheme_for(...)`) rather than the deleted `attn_quant`/`expert_quant` strings, and the fp8 branching this PR used to carry is gone — #418's scheme system owns it.

**Testing.** Full `pytest tests` with CUDA hidden (deterministic — with the GPUs visible this box is serving a model and 65-95 GPU tests fail on *main itself*, so only the set difference means anything): main 1205 passed / 0 failed, this branch **1214 passed / 0 failed**, and the set of failures new relative to main is **empty**. The +9 are the rewritten TP tests.

gdevenyi:

**Result: 1214 passed, 350 skipped, **no new failures**.**

The +9 over main are this PR's own tests, and they ran (not skipped).

gdevenyi:
Pushed `d2ad39b`: **let `TritonNvfp4MoEKernel` serve a tensor-parallel shard.**

Integrating this branch with the NVFP4 expert path after the #418/#427 quantization refactor, the engine refused to start:

```
KernelSelectionError: no usable kernel in table; triton: TP > 1 is not supported
```

Every NVFP4 MoE kernel carried `tp_ok=False`. For marlin and b12x that is right — they index the full intermediate. The triton kernel does not: once `MoEConfig` gained `tp_rank`/`tp_size`/`local_intermediate` it already has everything it needs, and its seven remaining uses of `cfg.intermediate` were simply the un-migrated spelling of `cfg.local_intermediate`. Fixed those and flipped `tp_ok=True` for the triton kernel only.

Worth a second pair of eyes on one thing: this made the bank-sizing bug below *silent* rather than fatal, so I want the sizing assertion reviewed as much as the flag.

**Related, and the more interesting failure.** A merge of upstream main into the deployment line silently dropped this same `local_intermediate` change. `layers/quantization/moe/nvfp4.py` was **new in main**, so the merge took upstream's copy wholesale with no conflict to resolve — the file did not exist on the other side to conflict with. At TP=2 the expert bank came out `(1280, 1280)` instead of `(640, 1280)`: twice the memory, wrong shard, no error. I found it by checking all 42 files the rebase touched for the "new in main, so no conflict" pattern; it was the only one. If you take this PR, that class of loss is worth knowing about for any future rebase across the refactor.

Also on this branch since the last comment: `tests/models/qwen4_exp/test_build_cpu.py`, which builds the whole model on `torch.device("meta")` and asserts both layer families, `lm_head`, and MoE weights on more than one layer. It exists because **every** model-construction test in `tests/models/qwen4_exp` is behind `requires_cuda`, so a constructor signature change (`Qwen4ExpMoE.__init__() got an unexpected keyword argument 'prefix'`) passed CI on a green CPU suite and only surfaced on hardware. Verified it fails without the fix it guards.

gdevenyi:
**End-to-end validation of this PR at TP=2, as a controlled A/B against the branch it replaces.**

Our deployment branch (upstream `fb7f732` + this PR + #386 + #389 + a few local fixes) was benchmarked against the previous production branch in one session, same harness, both servers settled before measuring, same 300 GSM8K questions in the same order. 2 × RTX 6000 Ada, `nvidia/Qwen3.8-Flash-Next-NVFP4`, offload MoE, fp8 KV, 2,097,152-token pool.

| | previous branch | with this PR |
|---|---|---|
| single-stream decode | 105.80 / 105.69 tok/s | 105.06 / 105.68 |
| aggregate @ 8 concurrent | 353.36 / 357.37 | 354.23 / 357.31 |
| TTFT, 1k prompt (3 runs) | 0.79 / 0.73 / 0.70 s | **0.680 / 0.683 / 0.680** |
| expert residency | 0.717 | 0.717 |
| GSM8K-300 greedy, 1536 cap | 97.67% (293/300) | **98.00% (294/300)** |
| genuine cap-independent misses | 7 | **5** |

Throughput and residency are identical. TTFT is ~7% better on the mean, and more interestingly it is **flat at 0.68 from the first measurement** where the old branch drifts 0.79 → 0.70 as its expert cache settles. Accuracy is one question higher with two fewer genuine misses — both inside noise at n=300, but nothing regressed.

Two methodology notes, because they cost me real time and may save yours:

**Do not benchmark this box inside ~10 minutes of start.** The aggregate reads 2–3% low while the expert cache fills. My first run on the new branch showed 343.7 against a 352.3 baseline and I nearly reported a 2.4% regression that did not exist; settled, it is 354–357.

**GSM8K at the default 768-token cap measures truncation as much as reasoning.** At 768 the two branches score 95.67% and 96.33%, which looks like a regression; the difference is entirely that one truncated 10 answers and the other 6. At 1536 the ordering reverses. Per-question flip analysis: 9 of 10 truncated answers become correct when the cap is lifted.

Since the last comment this branch also carries `1c09cee` — `TritonNvfp4MoEKernel` now serves a tensor-parallel shard, without which TP=2 raises `KernelSelectionError: no usable kernel in table` on an NVFP4 checkpoint.

gdevenyi:

madbrain76:
Thanks for this PR. I adapted it locally for TP=4 . It gave me a massive boost with my TR Pro 3955WX, 128 GB DDR4-3200, and 4 x 5060 Ti 16GB. I can run at the quant you listed at around 50 token/s single stream. And 100-150 with concurrency. Not too shabby for a box I spent < $5000 on. I ran out of my ChatGPT Pro Lite subscription for the next 4 days. So, I pointed Codex CLI at this model. It's surprisingly capable.

My instrumentation shows there is a ton of room to improve also. Neither the CPU utilization, GPU watts, RAM bandwidth or PCIe usage are peaked. I'm still trying to figure out the bottleneck.

I hope official TP>1 support comes to Freetoken soon.

PR #389, opened 2026-09-04
feat(qwen4_exp): load-time per-tensor FP8 dense projections (W8A8 via _scaled_mm)

gdevenyi:
## What this adds

`FREETOKEN_FP8_DENSE=1`: load-time per-tensor FP8 for the bf16 attention / GDN projections of `qwen4_exp`, run as cuBLASLt W8A8 GEMMs (`torch._scaled_mm`). Opt-in, default off, no checkpoint change: the weight reader quantizes `qkv_proj`, `o_proj`, GDN `in_proj` (q|k|v|z; the b|a gate rows stay bf16 as `in_proj_ba`, as in the block-fp8 checkpoints and in sglang / vLLM) and `out_proj` to e4m3 with one fp32 scale each, after TP sharding. `layers/fp8_dynamic.py` holds the op: a dynamic per-tensor activation scale (one fused Triton launch at decode sizes: amax pass, then the cast; a torch reduction plus a cast kernel above 64k elements), the `_scaled_mm` call, and the all-reduce for the row-parallel case. No host sync anywhere, so the decode path is CUDA-graph safe; the branch between the two quant paths is on the tensor shape, never on its values. Requires sm_89+ (`_scaled_mm`'s floor).

**Stacked on #385** (the TP commits): the column-merged / row-parallel classes shard the same way its bf16 ones do.

## Why

On 2 x RTX 6000 Ada (sm_89, torch 2.11.0+cu130, flashinfer 0.6.18) these projections are 2.67 GB of the ~4 GB a TP=2 rank reads per decode token. Micro-benchmark at the per-rank shapes (12 x [6656x2560] + [2560x3072], 36 x [8192x2560] + [2560x3072]), weight rotations larger than the 96 MB L2, activation quantization outside the timed region for the raw rows:

| route, 48 layers per rank | M=1 | M=8 | M=16 |
|---|---:|---:|---:|
| bf16 cuBLAS | 3.20 ms | 3.38 ms | 3.38 ms |
| FreeToken block-FP8 Triton (`fp8_block_linear.py`) | 3.78 | 4.59 | 4.58 |
| FreeToken per-row W8A16 (`fp8_pertensor_linear.py`) | 3.70 | 3.07 | 3.06 |
| FreeToken per-tensor W8A8 wrapper (same file) | 3.62 | 3.59 | 3.58 |
| raw `torch._scaled_mm`, per-tensor scales | **1.94** | **1.93** | **1.92** |

The FP8 tensor cores are fine here; the existing Triton FP8 kernels reach 20-60% of bf16's bandwidth on this stack and the per-tensor wrapper spends ~24 us per call before cuBLASLt, so the FP8 checkpoints are *slower* than bf16 at decode on this card. This path takes the direct route.

## Measurements

A/B on the machine above (`RadixArk/Qwen3.8-Flash-Next-NVFP4`, TP=2, `--moe-backend offload --ple-backend pinned --num-tokens 262144 --memory-ratio 0.94 --moe-prefill-hit-d2d --max-running-requests 16 --cuda-graph-max-bs 16`, vision tower loaded), both runs from the same build in the same session:

| | bf16 dense | `FREETOKEN_FP8_DENSE=1` |
|---|---:|---:|
| single-stream decode, median of 3 x 256 tokens | 89.8 tok/s | **99.2 tok/s (+10.4%)** |
| 8 concurrent, aggregate | 323.7 tok/s | 331.9 tok/s (+2.5%) |
| TTFT, 1.8k-token prompt | 0.84 s | 0.85 s |
| expert residency (`--moe-cache-auto`) | 91.9% (22,594 slots) | **95.8% (23,539 slots)** after 3f8f249 (a separate restart, same flags; single-stream on that build 99.6 tok/s), 91.4% before (see below) |
| 8-question probe, thinking off | 6/8 | 6/8, same answers |
| greedy 256 tokens: essay / code / 1k-prompt summary | reference | coherent; first divergence after 13 / 44 / 7 words |

For the greedy rows the bf16 run-to-run floor at TP=2 is 84 words / identical / identical, so FP8 changes the sampling trajectory more than the noise does, as expected from different numerics; the outputs stay on topic (the code case continues the same memoised Fibonacci past where bf16 stopped). The +10% matches the micro-benchmark: ~1.3 ms per step per rank (4 GEMMs per layer, 48 layers) of an ~11 ms step.

Residency first *fell* (22,594 -> 22,458 slots) although the reader emits 1.25 GiB less per rank: `in_proj_ba.weight` was produced as `t[qkvz:].contiguous()`, and `.contiguous()` on an already-contiguous row slice returns the view, so every GDN layer's 48 bf16 gate rows kept the whole sharded bf16 `in_proj` alive next to its fp8 copy (36 x 42 MB = 1.5 GiB per rank). 3f8f249 clones the slice (regression assertion in the test); the planner then resolves 23,539 slots, 95.8% residency, with the same 2.6 GiB headroom.

## Limits

- Per-tensor dynamic activation scaling across the batch is the coarsest scheme there is (one outlier channel in one row sets the scale for all rows). Row-wise scales were measured on torch 2.12.1+cu130 (where the sm_89 row-wise path is on-stream; pytorch@252bb4a is in v2.12.1), same shapes, GEMM only, 12 weight rotations. Absolute times in this measurement come from that separate torch build and differ from the table above (2.44 vs 1.94 ms for the same per-tensor route; cause not identified), so compare within the row only: M=1 / 8 / 16 per-tensor 2.44 / 2.39 / 2.40 ms per step per rank vs row-wise 2.41 / 2.40 / 2.38 (same), M=2048 18.5 vs 67.4 ms (the CUTLASS row-wise kernel is 3.6x slower than cuBLASLt at prefill sizes), relative error vs bf16 at M=16 with outlier channels 0.034-0.037 for both schemes. So row-wise buys nothing on this card, and it needs torch >= 2.12, which the `torch<2.12` pin (sglang-kernel 0.4.5) blocks anyway; per-tensor stays.
- The quality gate is a smoke test: the API has no logprobs, so it is the 8-question probe plus greedy comparisons, not perplexity. Numbers above.
- Only the four projection families; shared expert, hyper-connections, lm_head and PLE stay bf16. A block-FP8 checkpoint through this path is double quantization (block-FP8 dequantized, then re-quantized per tensor). Measured on lovedheart's MIXED_PRECISION build (NVFP4 experts, block-FP8 dense) on the same box, TP=1 on one card, 8 running: dequantize-to-bf16 58.4 tok/s / 135.9 at 8 concurrent / 36.6% residency, #392's native block-FP8 69.9 / 150.2 / 40.5%, this path 71.7 / 153.4 / 40.4%, 8-question probe 7/8 for all three. At TP=2 with the production flags this path gives 99.8 / 336.7 / 95.8%, the same as the RadixArk bf16-dense build (99.2-99.6 / 331.9). Probe 6/8 there. The 6/8 at TP=2 is the prime-count question (primes in 10..30, answered 5): the RadixArk build also misses it at TP=2 in bf16 and in FP8, the bf16-dequant lovedheart run at TP=2 scored 7/8, and every TP=1 run scored 7/8. It is a borderline question that flips with the numeric path (TP=2 reduction order, FP8), and this is a one-question smoke test, not a perplexity measurement.

## Testing

- `tests/models/qwen4_exp/test_fp8_dense.py`: loader round trip within e4m3 tolerance, the `in_proj` split per rank (fp8 q|k|v|z rows + bf16 b|a rows equal to the source), op state-dict contract (CPU); the op against `F.linear` on the dequantized weight at M=1 / 16 / 300 and the zero-input scale floor (GPU).
- `tests/models/qwen4_exp` + `tests/scheduler` on the box: 153 passed, 104 skipped (CPU).

### DIFF
--- a/python/freetoken/engine/config.py
+++ b/python/freetoken/engine/config.py
@@ -79,7 +79,7 @@ class EngineConfig:
     # ratio default above. A runtime cache rebuild sets this (num_swa_pages) to pin the window
     # regardless of the full anchor; the ratio is the startup default and the fallback.
     swa_num_pages_override: int | None = None
-    distributed_timeout: float = 60.0
+    distributed_timeout: float = 1800.0  # ranks reach the first collective minutes apart on a 100+ GiB offload load
     use_dummy_weight: bool = False
     use_pynccl: bool = True
     max_seq_len_override: int | None = None
--- a/python/freetoken/layers/embedding.py
+++ b/python/freetoken/layers/embedding.py
@@ -121,6 +121,17 @@ class ParallelLMHead(VocabParallelEmbedding):
             return super().state_dict(prefix=prefix, result=result)
         return {} if result is None else result
 
+    def _logits(self, x: torch.Tensor) -> torch.Tensor:
+        """The local vocab-shard GEMM; the seam a quantized head overrides."""
+        module = self.tied_embedding or self
+        return F.linear(x, module.weight, self.bias)
+
+    def _logits(self, x: torch.Tensor) -> torch.Tensor:
+        """The local vocab-shard GEMM; the seam a quantized head overrides."""
+        if self.tied_embedding is not None:
+            return F.linear(x, self.tied_embedding.weight, self.bias)
+        return self.quant_method.apply(self, x)
+
     @nvtx_annotate("LMHead")
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         ctx = get_global_ctx()
@@ -131,10 +142,7 @@ class ParallelLMHead(VocabParallelEmbedding):
             x = x[indices].contiguous()
             del indices
 
-        if self.tied_embedding is not None:
-            logits = F.linear(x, self.tied_embedding.weight, self.bias)
-        else:
-            logits = self.quant_method.apply(self, x)
+        logits = self._logits(x)
         if self.tp_size == 1:
             return logits
         input_shape = logits.shape
--- a/python/freetoken/layers/fp8_dynamic.py
+++ b/python/freetoken/layers/fp8_dynamic.py
@@ -1,0 +1,154 @@
+"""Load-time per-tensor FP8 dense linear (W8A8 through cuBLASLt ``torch._scaled_mm``).
+
+The weight is e4m3 ``[out_local, in_local]`` with one fp32 ``weight_scale`` (shape ``()``),
+produced by the model's weight reader from the bf16 checkpoint tensor (after TP sharding).
+The activation is quantized per call with a dynamic per-tensor scale: one fused Triton launch
+(amax pass, then cast) at decode sizes, a torch reduction plus a cast kernel above that. No
+host sync anywhere (the scales stay on the device), so the decode path is CUDA-graph safe;
+the branch between the two paths is on the tensor *shape*, never on its values.
+
+Measured on an RTX 6000 Ada (sm_89, torch 2.11.0+cu130) at the qwen4_exp TP=2 shapes, per
+decode step per rank over 48 layers: bf16 cuBLAS 3.2-3.4 ms, this path 1.9 ms at M=1/8/16.
+Requires sm_89+ (``_scaled_mm``'s floor; Ampere has no FP8 tensor cores).
+"""
+
+from __future__ import annotations
+
+from typing import List
+
+import torch
+import triton
+import triton.language as tl
+from freetoken.distributed import DistributedCommunicator, get_tp_info
+from freetoken.utils import div_even
+
+from .base import BaseOP
+
+FP8 = torch.float8_e4m3fn
+E4M3_MAX = 448.0
+_FUSED_MAX_ELEMENTS = (
+    65536  # one program handles the whole tensor below this (decode sizes)
+)
+_SCALE_FLOOR = 1e-12  # an all-zero activation (graph warmup buffers) must not give 1/0
+
+
+@triton.jit
+def _quant_fused_kernel(x_ptr, out_ptr, scale_ptr, n, BLOCK: tl.constexpr):
+    """One program: max|x| over the tensor, then the cast. Decode-sized inputs only."""
+    acc = tl.zeros([BLOCK], dtype=tl.float32)
+    for start in range(0, n, BLOCK):
+        offs = start + tl.arange(0, BLOCK)
+        v = tl.load(x_ptr + offs, mask=offs < n, other=0.0).to(tl.float32)
+        acc = tl.maximum(acc, tl.abs(v))
+    amax = tl.maximum(tl.max(acc, axis=0), 1e-12)
+    tl.store(scale_ptr, amax / 448.0)
+    inv = 448.0 / amax
+    for start in range(0, n, BLOCK):
+        offs = start + tl.arange(0, BLOCK)
+        mask = offs < n
+        v = tl.load(x_ptr + offs, mask=mask, other=0.0).to(tl.float32) * inv
+        v = tl.minimum(tl.maximum(v, -448.0), 448.0)
+        tl.store(out_ptr + offs, v.to(tl.float8e4nv), mask=mask)
+
+
+@triton.jit
+def _quant_cast_kernel(x_ptr, out_ptr, scale_ptr, n, BLOCK: tl.constexpr):
+    """Cast under a scale already on the device (prefill sizes; the amax is a torch reduction)."""
+    offs = tl.program_id(0) * BLOCK + tl.arange(0, BLOCK)
+    mask = offs < n
+    inv = 1.0 / tl.load(scale_ptr)
+    v = tl.load(x_ptr + offs, mask=mask, other=0.0).to(tl.float32) * inv
+    v = tl.minimum(tl.maximum(v, -448.0), 448.0)
+    tl.store(out_ptr + offs, v.to(tl.float8e4nv), mask=mask)
+
+
+def quant_per_tensor(x: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
+    """``(x_fp8, scale)`` with ``x ~= x_fp8 * scale``; ``x`` contiguous, ``scale`` fp32 ``()``."""
+    n = x.numel()
+    out = torch.empty_like(x, dtype=FP8)
+    if n <= _FUSED_MAX_ELEMENTS:
+        scale = torch.empty((), dtype=torch.float32, device=x.device)
+        _quant_fused_kernel[(1,)](x, out, scale, n, BLOCK=4096, num_warps=8)
+        return out, scale
+    amax = torch.linalg.vector_norm(x, ord=float("inf")).float()
+    scale = amax.clamp_min_(_SCALE_FLOOR).div_(E4M3_MAX)
+    _quant_cast_kernel[(triton.cdiv(n, 4096),)](
+        x, out, scale, n, BLOCK=4096, num_warps=4
+    )
+    return out, scale
+
+
+def fp8_dynamic_linear(
+    x: torch.Tensor, weight: torch.Tensor, weight_scale: torch.Tensor
+) -> torch.Tensor:
+    """``x @ (weight * weight_scale)^T`` in W8A8; ``weight`` [N, K] e4m3 row-major, whose ``.t()``
+    is the column-major operand cuBLASLt wants (a stride change, never a copy)."""
+    *lead, k = x.shape
+    x2 = x.reshape(-1, k).contiguous()
+    x8, scale = quant_per_tensor(x2)
+    y = torch._scaled_mm(
+        x8, weight.t(), scale_a=scale, scale_b=weight_scale, out_dtype=x.dtype
+    )
+    return y.reshape(*lead, weight.shape[0])
+
+
+class Fp8DynamicLinear(BaseOP):
+    """Per-tensor FP8 linear over the local shard; ``all_reduce`` adds the TP sum (row-parallel)."""
+
+    def __init__(self, local_isize: int, local_osize: int, *, all_reduce: bool = False):
+        assert local_isize % 16 == 0 and local_osize % 16 == 0, (
+            local_isize,
+            local_osize,
+        )
+        self.local_input_size = local_isize
+        self.local_output_size = local_osize
+        self.weight = torch.empty(local_osize, local_isize, dtype=FP8)
+        self.weight_scale = torch.empty((), dtype=torch.float32)
+        self._comm = (
+            DistributedCommunicator() if all_reduce and get_tp_info().size > 1 else None
+        )
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        y = fp8_dynamic_linear(x, self.weight, self.weight_scale)
+        if self._comm is not None:
+            y = self._comm.all_reduce(y)
+        return y
+
+
+class Fp8DynamicColMerged(Fp8DynamicLinear):
+    """Drop-in for ``LinearColParallelMerged``: one weight concatenating several projections
+    along the output dim; the caller splits the output by the local sizes as before."""
+
+    def __init__(
+        self,
+        input_size: int,
+        output_sizes: List[int],
+        local_output_sizes: List[int] | None = None,
+    ):
+        tp = get_tp_info()
+        if local_output_sizes is None:
+            local_output_sizes = [div_even(size, tp.size) for size in output_sizes]
+        self.output_sizes = list(output_sizes)
+        self.local_output_sizes = list(local_output_sizes)
+        super().__init__(input_size, sum(local_output_sizes))
+
+
+class Fp8DynamicRowParallel(Fp8DynamicLinear):
+    """Drop-in for ``LinearOProj`` / ``LinearRowParallel``: the input dim is sharded, the
+    all-reduce runs after the local GEMM (each rank scales its own shard)."""
+
+    def __init__(self, input_size: int, output_size: int):
+        super().__init__(
+            div_even(input_size, get_tp_info().size), output_size, all_reduce=True
+        )
+
+
+__all__ = [
+    "FP8",
+    "E4M3_MAX",
+    "Fp8DynamicColMerged",
+    "Fp8DynamicLinear",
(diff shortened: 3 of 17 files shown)


gdevenyi:
3f8f249: the expert-cache planner now sees the halved dense bytes. `in_proj_ba.weight` came out of `t[qkvz:].contiguous()`, which is a view of the sharded bf16 `in_proj` (a contiguous row slice is already contiguous), so each GDN layer kept 42 MB of bf16 alive next to its fp8 copy, 1.5 GiB per TP=2 rank. With a clone: 22,458 -> 23,539 slots (91.4% -> 95.8% residency) (a separate restart with the same flags, not the paired A/B runs in the body). Single-stream re-measured on that build: 99.6 tok/s; vision, long-context and text probes pass. Body updated.

gdevenyi:
Two follow-up commits pushed onto this branch, both measured end to end on 2 x RTX 6000 Ada (sm_89, TP=2, `nvidia/Qwen3.8-Flash-Next-NVFP4`, torch 2.11).

**`perf(fp8): quantize activations with a block-parallel amax`**

`quant_per_tensor` walked the whole tensor in a single Triton program, so its cost was linear in the input: 9.8 us for an `[8, 10240]` activation, 68.6 us at `[64, 10240]`. Above 16384 elements it now splits into a block-parallel partial amax plus a reduce+cast. Same arithmetic, so a given tensor quantizes bit for bit as before (there is a test for that); the old three-launch torch reduction path above 65536 elements goes away.

| n | before | after |
|---:|---:|---:|
| 20,480 | 2.59 us | 2.22 us |
| 81,920 | 9.83 us | 2.44 us |
| 655,360 | 68.60 us | 3.06 us |

The partial count reaches the kernels as a runtime argument, not a constexpr, and pass 1 is grid-strided and capped, so both kernels have exactly one compiled variant. Passing it as a constexpr instead costs a fresh compilation per distinct input length, which in a server means unbounded variant growth and a compile stall mid-generation — that cost ~7 tok/s on the first benchmark run before it was fixed.

**`feat(layers): optional per-tensor FP8 lm_head (FREETOKEN_FP8_LMHEAD=1)`**

The vocab-parallel head is the last large bf16 read on the decode path: 0.64 GB per step per rank after the TP=2 vocab split, 0.80 ms of a 9.9 ms step in an nsys trace. `ParallelLMHead.forward` gains a `_logits()` seam and `Fp8ParallelLMHead` overrides only that, so the all_gather above it is untouched. Untied embeddings only.

Its own env flag rather than riding on `FREETOKEN_FP8_DENSE`, because this one moves the logits directly rather than feeding a norm or a sigmoid, so it should carry its own quality gate.

Measured against this branch as the baseline, production flags, warm single-stream (the benchmark's first run is always cold):

| | single-stream | 8 concurrent | TTFT 1k | GSM8K-300 | 8-question probe |
| baseline | 100.95 tok/s | 323.3 | 0.90 s | 97.00% | 7/8 |
| `FREETOKEN_FP8_LMHEAD=1` | **103.39** | 324.5 | 0.87 s | **97.00%** | 7/8 |

The aggregate barely moves, and that is expected: at 8 concurrent the step is ~24.7 ms rather than ~10 ms because the routed-expert read scales with the batch, so a fixed dense-side saving is a much smaller fraction of it. This is a single-stream lever.

Tests: `tests/models/qwen4_exp` + `tests/layers` give 120 passed / 50 skipped on this branch, the same 3 `test_qsa_backend.py::test_chunked_prefill_matches_one_shot` failures as the branch point (those are what #388 fixes) and no new ones.

One thing I tried on top and am **not** proposing: the same treatment for the hyper-connection mixers, which look like the obvious next target (`LinearReplicated`, so every rank re-reads all 1.30 GB of them every step). It benchmarks flat end to end — 100.95 -> 100.56 tok/s. The FP8 GEMMs do save 5.7 us per mix block cold, but the dynamic per-tensor quant hands 4.6 us of that straight back in three extra launches. The quant cost is per call, not per byte, which is why it pays on one 636 MB GEMM and not on 97 GEMMs of 6.9 MB. Worth knowing before someone repeats it; making it land needs the activation quant fused into the norm that already writes the tensor.

gdevenyi:
Rebased onto `fb7f732`. The quantization refactor removed the hook this PR hung on, so the selection logic changed shape — the kernels and the loader are untouched.

**What this PR is, restated against the new architecture.** #418 gives every module its scheme from the checkpoint's `QuantConfig`, and `layers/quantization/linear/fp8_tensor.py` already serves per-tensor FP8 — but only when the checkpoint *declares* it: `TorchFp8TensorLinearKernel.unusable_reason` returns "the checkpoint has no static activation scale" without an `input_scale`. That is **static** W8A8 for a quantized checkpoint.

This PR is the other case: a checkpoint that ships its dense side in **bf16**, quantized per-tensor at load with a **dynamic** activation scale. Qwen3.8-Flash-Next NVFP4 is exactly that — a 292-entry modelopt `ignore` list leaves attention, GDN, the shared expert, HC, PLE and lm_head unquantized. So the two do not overlap, and the flag stays opt-in.

**Selection.** #426 deleted the `attn_quant` / `dense_quant` / `lm_head_quant` block from `qwen4_exp.parse_config`, which is what used to carry `"fp8_dynamic"`. `ModelConfig.attn_quant` still exists, so `parse_config` now sets just the one line:

```python
attn_quant = "fp8_dynamic" if fp8_dense_enabled() else "none"
```

and every consumer additionally requires the checkpoint to declare nothing for that module:

```python
fp8_dyn = config.attn_quant == "fp8_dynamic" and (
    config.quant is None or config.quant.scheme_for(f"{prefix}.qkv_proj") is None
)
```

so a quantized checkpoint always wins and goes through `QuantConfig`; the synthetic path only fills in where there is no scheme. GDN does the same (`self._dynamic_fp8 = attn_quant == "fp8_dynamic" and not self._split_in_proj`), and the FP8 lm_head additionally requires `config.quant is None`.

I think the cleaner long-term shape is to express this as a real scheme + `LinearMethod` in `layers/quantization/` — "synthesize a scheme for a bf16 checkpoint when the operator asks for it" — so it stops being a parallel selection path. I did not do that here because it changes the PR from a rebase into a redesign, and I would rather agree the shape first. Happy to do it if you prefer that.

**Other resolutions:** `ParallelLMHead.forward` gained a `_logits()` seam around upstream's exact body (`tied_embedding` -> `F.linear`, else `quant_method.apply`), which `Fp8ParallelLMHead` overrides; the old `lm_head_quant == "nvfp4"` branch is gone with the scheme system; `models/quant_linear.py` and the `Fp8PerTensorColMerged` / `Fp8BlockColMerged` branching this PR carried are deleted upstream and gone from here.

**Testing.** Full `pytest tests`, CUDA hidden (deterministic; with GPUs visible this box serves a model and 65-95 GPU tests fail on main itself): main 1205 passed / 0 failed, this branch **1219 passed / 0 failed**, **zero** failures new relative to main.

gdevenyi:

**Result: 1219 passed, 359 skipped, **no new failures**.**

9 more tests skip here than on main: this PR adds that many CUDA-only tests the hidden-GPU run does **not** exercise. So the result above says it merges, imports and leaves every CPU-reachable path intact — it says nothing about the kernels themselves. I am running those on free cards in a maintenance window and will post the numbers here.

gdevenyi:
Two correctness fixes pushed onto this branch (`d377a20`), both found while integrating it with TP=2 rather than by the test suite. Worth flagging because the first is a real trap in this design.

**1. The FP8 lm_head predicate was evaluated twice, and the copies drifted.**

`model.py` decided "build a `Fp8ParallelLMHead`" and `weight.py` decided "emit `lm_head.weight_scale`" from separate expressions. On the shipping NVIDIA NVFP4 checkpoint they disagreed and startup died:

```
RuntimeError: Unexpected keys in state_dict: ['lm_head.weight_scale']
```

The builder's extra condition was `config.quant is None`. That is the wrong question. What matters is whether the checkpoint declares a scheme **for lm_head**, not whether it has a `QuantConfig` at all — this checkpoint ships NVFP4 routed experts (so `quant` is not `None`) with `lm_head` in the modelopt `ignore` list (so it has no scheme, and the synthetic FP8 head is exactly what is wanted). Both sides now call one `config.use_fp8_lmhead()` that owns the flag, the tied-embedding exclusion and the scheme test.

Generalising: any load-time quantization has this shape. The builder allocates slots, the loader emits keys, and `load_state_dict` is strict — so the predicate has to be a single named function, not an expression repeated at both ends.

**2. `use_fp8_lmhead` referenced `fp8_lmhead_enabled` without importing it**, so it raised `NameError` under *every* flag setting including the default. Caught before it reached a GPU.

**How #2 was caught, and the check I would suggest upstream adopt.** Build the model on `torch.device("meta")`, take `set(model.state_dict())` minus the routed experts, run the checkpoint's `weight_map` through the loader's own `_rename`/`_FUSIONS`/FP8 renames, and diff the two name sets. No GPU, no tensors, ~1 min. Result on `nvidia/Qwen3.8-Flash-Next-NVFP4` at TP=2:

| `FREETOKEN_FP8_DENSE` | `FREETOKEN_FP8_LMHEAD` | model slots | loader keys | |
|---|---|---|---|---|
| 0 | 0 | 794 | 794 | match |
| 0 | 1 | 795 | 795 | match |
| 1 | 0 | 926 | 926 | match |
| 1 | 1 | 927 | 927 | match |

The reason this class of bug survives CI is that every model-construction test in `tests/models/qwen4_exp` sits behind `requires_cuda`, so the CPU suite stays green through a constructor that cannot even be called. A meta-device build test costs nothing and closes that hole; I added one (`tests/models/qwen4_exp/test_build_cpu.py`) on the TP branch (#385) and confirmed it fails without the fix it guards.

**Known limitation, not fixed here.** The model side gates the dense FP8 path per layer — `attention.py` tests `config.quant.scheme_for(f"{prefix}.qkv_proj") is None`, `gdn.py` the same for `in_proj_qkvz` — but `weight.py` gates on the model-wide `config.attn_quant == "fp8_dynamic"` with no `scheme_for` test. On every checkpoint that exists today the skip list covers all dense projections, so the two agree and the table above is clean. A checkpoint that quantized *some* dense attention projection would reproduce failure 1 in the dense path. I have left it alone rather than guess at a name mapping I cannot exercise; a reviewer may prefer a loud refusal in the loader over a silent mismatch.

gdevenyi:
Third fix on this branch (`bd106c9`), and this one is the reason I am writing rather than just pushing: it cost a GPU window, and the check that catches it is one command.

The rebase onto the quantization refactor dropped a single line from `GatedDeltaNet.forward`:

```python
nk, nv = self._local_num_k_heads, self._local_num_v_heads
```

and kept all eleven uses of `nk` and `nv` below it. The server loaded the weights, selected its kernels, built the expert banks, and then died at the first CUDA-graph capture, five minutes in:

```
File "python/freetoken/models/qwen4_exp/gdn.py", line 189, in forward
  b, a = torch.split(ba, [nv, nv], dim=-1)
NameError: name 'nv' is not defined
```

**`ruff check` reports this in under a second.** F821 is in ruff's default rule set; no configuration is needed. I had run the CPU test suite, a meta-device build test, and a loader-vs-model key diff — all green, because none of them execute `forward`. A name undefined inside a method body survives every construction test ever written and is invisible until the first real token.

That suggests something worth considering for the repo rather than just for me: there is no ruff configuration and no lint job in `.github/workflows/`. `ruff check --select F821 python/ tests/` on current `main` returns **7 hits** in three files:

```
python/freetoken/kvcache/cache_status.py        (5)
python/freetoken/models/glm5_next/attention.py  (1)
python/freetoken/models/glm_moe_dsa/attention.py (1)
```

I have not looked at whether those are live paths or dead branches, and some may be deliberate forward references. But each one is a `NameError` waiting for whoever first executes that line, and a CI step that runs `ruff check` would have turned my five-minute outage into a red tick on a pull request. Happy to open that as a separate PR if it is wanted.

The same commit removes a duplicated `from freetoken.distributed import get_tp_info` two lines below the first, left by the same merge.

**Where this branch now stands on hardware.** 2 x RTX 6000 Ada, TP=2, `nvidia/Qwen3.8-Flash-Next-NVFP4`, production flags: the server starts, and single-stream decode is 105.49 tok/s across three samples spanning 0.37 tok/s, against 105.6 on the branch this replaces — level. Expert residency 71.7%, geometry unchanged. GSM8K-300 greedy is still being separated from a token-limit artifact (8 of 13 misses hit the 768-token cap rather than answering wrongly), and I will post the number once it is clean rather than a number that flatters or maligns the branch by accident.

gdevenyi:
**Accuracy result for the load-time FP8 path, measured properly this time.**

Earlier I reported GSM8K on this stack as "97.00% unchanged" from an older note. That figure was stale and I should not have quoted it. Here is a controlled A/B run in one session, same harness, both servers settled, same 300 questions in the same order, on 2 × RTX 6000 Ada at TP=2 with `FREETOKEN_FP8_DENSE=1 FREETOKEN_FP8_LMHEAD=1`:

| | without this branch | with FP8 dense + FP8 lm_head |
|---|---|---|
| GSM8K-300 @ 1536 cap | 97.67% (293/300) | **98.00% (294/300)** |
| genuine cap-independent misses | 7 | **5** |
| single-stream decode | 105.80 / 105.69 tok/s | 105.06 / 105.68 |
| aggregate @ 8 concurrent | 353.36 / 357.37 | 354.23 / 357.31 |
| TTFT, 1k prompt | 0.79 / 0.73 / 0.70 s | **0.680 / 0.683 / 0.680** |

So per-tensor FP8 on the dense projections and the lm_head costs **no measurable accuracy** on this checkpoint — one question better, two fewer genuine misses, which is noise at n=300 but is certainly not a regression.

One caution for anyone reproducing this. **At the default 768-token cap the comparison inverts**: 95.67% with FP8 against 96.33% without, which reads as a 2-question regression. It is not. The FP8 run truncated 10 answers at the cap and the baseline truncated 6, and a per-question flip analysis shows 9 of the 10 truncated answers become correct at 1536. Quoting GSM8K on this model without stating the cap is close to meaningless; the untruncated number is the one that means something.

The branch is also now free of the two startup faults reported above — the `use_fp8_lmhead` predicate drift and the missing import — and has run a full benchmark plus two 300-question GSM8K passes without incident.

gdevenyi:

PR #390, opened 2026-09-05
fix(models): better support for mixed-precision compressed-tensors NVFP4 

Sam-Izdat:
## Summary

Adds support for compressed-tensors checkpoints that mix per-tensor
**FP8** attention/shared-expert projections with **NVFP4** routed
experts - a layout produced by `llm-compressor` and found in the wild.

Without this, such checkpoints either crash on load (FP8 attn gets
`torch.cat`'d with bf16 weights) or produce garbled output (the dequant
kernel multiplies by the per-row global scale, but `llm-compressor`
stores the QUANT-side scale rather than the DEQUANT-side divisor that
`modelopt` stores).

The fix is in five files, +236/-63 lines. The key idea is to use the
**on-disk tensor naming** as the ground truth for which convention is
in use (`weight_packed` -> llm-compressor, `weight_scale_2` -> modelopt)
rather than relying on the config-side `format: nvfp4-pack-quantized`
string (which both exporters set).

## Test models

| HF repo | Layout | Result |
|---------|--------|--------|
| `nvidia/Qwen3.6-35B-A3B-NVFP4` | FP8 attn + NVFP4 shared-expert + NVFP4 experts (modelopt mixed_precision) | regression-clean, coherent |
| `primitive-ai/Ornith-1.5-35B-A3B-agentic-NVFP4-FP8` | FP8 attn + FP8 shared-expert + NVFP4 experts (compressed-tensors mixed-precision) | coherent (was: garbled output) |

Tested with `--moe-backend offload --expert-load parallel` on RTX 3060
12 GB (`nvfp4_backend='triton'`). Ornith decoding at ~30-40+ tok/s with minimal tuning.

## Fixed

1. **Crash on mixed-precision attn+shared-expert** — the
   compressed-tensors iter doesn't know about FP8. Routing
   compressed-tensors MoE to the modelopt iter (which handles FP8
   attn) when the shared-expert is in the FP8 group avoids the
   cat-of-fp8-and-bf16 crash.

2. **Missing global-reciprocal for llm-compressor NVFP4 experts** —
   the dequant kernel multiplies by the per-row global; `modelopt`
   stores the DEQUANT-side divisor directly, but `llm-compressor`
   stores the QUANT-side scale that must be reciprocated. A
   per-checkpoint flag on the NVFP4 source spec picks the right
   convention.

3. **Unknown NVFP4 kind `input_global_scale` (parallel build)** —
   the bank dispatch only knew `weight_scale_2` (modelopt). Alias
   `weight_global_scale -> weight_scale_2` via `spec.kind_map`;
   skip `input_scale` / `input_global_scale` (activation scales,
   not bank tensors).

4. **Per-expert naming regex too narrow for parallel build** — the
   expert key pattern only matched `weight | weight_scale |
   weight_scale_2`. Extended to cover the full `llm-compressor`
   per-expert naming set so the parallel bank builder's
   `weight_info` populates correctly.

5. **Iter dispatch routed mixed-precision MoE to the
   compressed-tensors iter**, which holds more GPU state per-shard
   than the default iter (caused OOM on 12 GB). Reverted to the
   default iter for that case.

6. **The on-disk-naming heuristic for reciprocal was too coarse.**
   `format: nvfp4-pack-quantized` is set by BOTH `llm-compressor` and
   `modelopt` re-exports. Now probes the safetensors index for
   `weight_packed` (llm-compressor signature) vs `weight_scale_2`
   (modelopt signature) and picks the convention from the data, not
   the config claim.

## Files changed (5, +236/-63)

- `python/freetoken/models/config.py` — `_nvfp4_global_reciprocal`
  heuristic probing the safetensors index;
  `ModelConfig.nvfp4_global_reciprocal` field
- `python/freetoken/models/nvfp4_banks.py` — bank kind dispatch
  handles `weight_global_scale` (via `spec.kind_map`) and skips
  activation scales; serial and parallel paths updated in lockstep
- `python/freetoken/models/qwen3_5_moe/config.py` — `_has_moe_experts`
  helper; `_attn_quant` and `_expert_quant` extended for
  compressed-tensors; `dense_quant` probe for FP8 shared-expert;
  parse_config wiring
- `python/freetoken/models/qwen3_5_moe/moe.py` — `_SharedExpert`
  refactored to use the `quant_linear` factory, with
  `attn_quant='fp8_pertensor'` dispatch for the mixed-precision
  shared-expert path
- `python/freetoken/models/qwen3_5_moe/weight.py` — per-checkpoint
  `_spec_for` with `kind_map`; `_load_maybe_quantized` and
  `_nvfp4_parts` accept llm-compressor naming; `_PT_FP8_FUSE` and
  `_CT_NVFP4_FUSE` extended for shared-expert gate|up fuse; expanded
  `_NVFP4_EXPERT_KEY_RE` regex

## Backwards compatibility

- NVIDIA `Qwen3.6-35B-A3B-NVFP4` (modelopt) is regression-clean.
  No NVFP4 reciprocal applied (the data shows `weight_scale_2` only,
  which signals the modelopt convention).
- The default iter (the one most production single-group NVFP4 MoE 
  exports use) is unchanged in behavior.

## Notes

The heuristic that disambiguates `llm-compressor` from `modelopt`
probes the safetensors index for a routed-expert tensor's sibling
suffixes. This is a one-time read at model-load time; no per-tensor
overhead. If both `weight_packed` and `weight_scale_2` are absent
(e.g. some other compressed-tensors flavor), the heuristic falls
back to the safe default of "no reciprocal" same as
`global_reciprocal=False` upstream.

### DIFF
--- a/python/freetoken/models/config.py
+++ b/python/freetoken/models/config.py
@@ -91,6 +91,66 @@ def detect_compressed_tensors_nvfp4(hf_config: Any) -> bool:
     return saw_nvfp4
 
 
+def _nvfp4_global_reciprocal(hf_config: Any) -> bool:
+    """Whether the routed-experts' per-tensor global scale needs to be reciprocated
+    at load time.
+
+    llm-compressor stores the QUANT-side scale (the multiplier baked into the local
+    fp8 scales before the FP4 cast); the dequant kernel divides by it, so the loader
+    must reciprocate 1/x. Modelopt's NVFP4 stores the DEQUANT-side divisor directly
+    and the kernel multiplies (no transform).
+
+    The ``format: nvfp4-pack-quantized`` string on the config_group is set by BOTH
+    exporters, so it alone can't disambiguate. The ground-truth signal is the on-disk
+    tensor naming: llm-compressor uses ``weight_packed`` + ``weight_global_scale`` +
+    ``input_global_scale``; modelopt uses ``weight`` + ``weight_scale_2`` + ``input_scale``.
+    We probe the safetensors index for one routed-expert tensor and pick the convention
+    from its sibling suffixes.
+
+    Returns True (reciprocate) for llm-compressor naming, False (use as-is) for
+    modelopt naming. Returns False when nothing is found (caller can't decide, so the
+    bank loader keeps the on-disk value verbatim -- the modelopt default)."""
+    # Cheap pre-check: ``quant_method`` should be compressed-tensors for either case.
+    quant = getattr(hf_config, "quantization_config", None)
+    if not quant:
+        return False
+    get = quant.get if isinstance(quant, dict) else (lambda k, d=None: getattr(quant, k, d))
+    if str(get("quant_method") or "").lower() != "compressed-tensors":
+        return False
+
+    # Probe the safetensors index for an expert tensor and inspect its sibling suffixes.
+    # Cheap read (one json parse) and bounds-checked (only the first hit is needed).
+    try:
+        from pathlib import Path as _Path
+        from freetoken.utils.hf import download_hf_weight as _download
+        folder = _Path(_download(hf_config._name_or_path
+                                 if hasattr(hf_config, "_name_or_path") else None))
+        idx_path = folder / "model.safetensors.index.json"
+        if not idx_path.exists():
+            return False  # single-file or no index -- fall back to safe default
+        import json as _json
+        weight_map = _json.loads(idx_path.read_text())["weight_map"]
+    except Exception:
+        return False  # any read error -- don't reciprocate (modelopt default is safer)
+
+    # Find one expert key in any shard; the suffix names reveal the layout.
+    has_packed = False       # llm-compressor writes weight_packed
+    has_scale2 = False       # modelopt writes weight_scale_2
+    for name in weight_map:
+        if ".mlp.experts." in name and ".gate_proj." in name:
+            has_packed = has_packed or name.endswith(".weight_packed")
+            has_scale2 = has_scale2 or name.endswith(".weight_scale_2")
+            if has_packed or has_scale2:
+                break
+    # Disambiguate: llm-compressor uses weight_packed (and weight_global_scale); modelopt
+    # uses weight + weight_scale_2 directly. A safetensors checkpoint that has weight_packed
+    # for routed experts is llm-compressor; one that has weight_scale_2 is modelopt.
+    if has_packed:
+        return True
+    return False
+
+
+
 @dataclass(frozen=True)
 class RotaryConfig:
     head_dim: int
@@ -270,6 +330,11 @@ class ModelConfig:
     expert_quant: str = "none"
     # NVFP4 routed-expert GEMM backend (--nvfp4-backend); injected from EngineConfig.
     nvfp4_backend: str = "triton"
+    # Reciprocate the per-tensor NVFP4 expert global scale at load (llm-compressor NVFP4
+    # stores the QUANT-side scale; the dequant kernel multiplies by the DEQUANT-side value,
+    # so we read 1/x from disk). False for modelopt NVFP4 (dequant-side value is on disk).
+    # See ``_nvfp4_global_reciprocal`` for the heuristic.
+    nvfp4_global_reciprocal: bool = False
     # Block size (out, in) for block-wise weight quantization (fp8_block: (128, 128)).
     weight_block_size: tuple[int, int] | None = None
     # Weight quantization of the *dense* attention / GatedDeltaNet projections (separate
--- a/python/freetoken/models/nvfp4_banks.py
+++ b/python/freetoken/models/nvfp4_banks.py
@@ -130,9 +130,13 @@ def load_nvfp4_expert_source_banks(
             raise ValueError(f"{spec.desc}: unknown NVFP4 expert projection {proj!r}")
         kind = _canon_kind(spec, match.group("kind"))
         if kind == "weight_scale_2":
+            # modelopt's ``weight_scale_2`` and llm-compressor's ``weight_global_scale``
+            # (aliased via spec.kind_map) both carry the per-row global.
             global_shards[shard].append((name, match, bank_layer))
         elif kind in {"weight", "weight_scale"}:
             weight_shards[shard].append((name, match, bank_layer))
+        elif kind in {"input_scale", "input_global_scale"}:
+            pass  # activation scale (W4A8 / W8A8); consumed with its .weight
         else:
             raise ValueError(f"{spec.desc}: unknown NVFP4 expert tensor kind {kind!r}")
 
@@ -254,9 +258,13 @@ def load_nvfp4_expert_source_banks_parallel(
             continue
         kind = _canon_kind(spec, match.group("kind"))
         if kind == "weight_scale_2":
+            # modelopt names the per-row global ``weight_scale_2``; llm-compressor names it
+            # ``weight_global_scale`` (after ``_canon_kind`` maps via spec.kind_map, or alias).
             global_names_by_shard[shard].append(name)
         elif kind in {"weight", "weight_scale"}:
             weight_info[name] = (match, bank_layer)
+        elif kind in {"input_scale", "input_global_scale"}:
+            pass  # activation scale (W4A8 / W8A8); consumed with its .weight, not a bank tensor
         else:
             raise ValueError(f"{spec.desc}: unknown NVFP4 expert tensor kind {kind!r}")
 
--- a/python/freetoken/models/qwen3_5_moe/config.py
+++ b/python/freetoken/models/qwen3_5_moe/config.py
@@ -7,7 +7,9 @@ from freetoken.models.config import (
     LinearGatedDeltaGroupConfig,
     ModelConfig,
     RotaryConfig,
+    _nvfp4_global_reciprocal,
     detect_compressed_tensors_nvfp4,
+    detect_expert_quant,
 )
 
 
@@ -39,6 +41,15 @@ def _fp8_block_quant(hf_config: Any) -> tuple[str, tuple[int, int] | None]:
     return "none", None
 
 
+def _has_moe_experts(hf_config: Any) -> bool:
+    """True if the model declares any routed experts (``text_config.num_experts > 0``).
+    Used to disambiguate dense compressed-tensors NVFP4 (Qwen3.6-27B) from MoE mixed-precision
+    compressed-tensors (Ornith: FP8 attn/shared_expert + NVFP4 experts) so per-group quant
+    detection isn't overridden by a blanket "everything is NVFP4" assumption."""
+    text = getattr(hf_config, "text_config", hf_config)
+    return (getattr(text, "num_experts", 0) or 0) > 0
+
+
 def _expert_quant(hf_config: Any) -> str:
     """Quantization format of the *routed* experts (the only weights served from the
     offload cache). The nvidia/modelopt checkpoints are either plain NVFP4 (``quant_algo``
@@ -60,6 +71,12 @@ def _expert_quant(hf_config: Any) -> str:
                     return "nvfp4"
                 if "fp8" in expert_algo:
                     return "fp8"
+    if str(get("quant_method") or "").lower() == "compressed-tensors":
+        # llm-compressor: the routed experts' group carries ``format: nvfp4-pack-quantized``.
+        # ``detect_expert_quant`` (top-level) already returns "nvfp4" for this -- mirror it
+        # here so the offload expert bank activates.
+        if detect_expert_quant(hf_config) == "nvfp4":
+            return "nvfp4"
     return "none"
 
 
@@ -108,17 +125,31 @@ def _attn_quant(hf_config: Any) -> str:
     """Per-tensor FP8 on the *dense* attention/GDN projections. The modelopt
     ``MIXED_PRECISION`` checkpoints tag ``self_attn.{q,k,v,o}_proj`` and
     ``linear_attn.{in_proj_qkv,in_proj_z,out_proj}`` with ``quant_algo`` ``FP8`` (fp8-e4m3
-    weight + a scalar ``weight_scale``; W8A16). Returns ``"fp8_pertensor"`` when present,
-    else ``"none"`` (NVFP4 dense weights -- shared_expert/lm_head -- stay dequant-at-load)."""
+    weight + a scalar ``weight_scale``; W8A16). llm-compressor mixed-precision checkpoints
+    (Ornith) use a per-tensor-FP8 ``config_groups`` entry targeting the same projections.
+    Returns ``"fp8_pertensor"`` when present, else ``"none"``."""
     get = _quant_accessor(hf_config)
     if get is None:
         return "none"
     layers = get("quantized_layers") or {}
-    if not isinstance(layers, dict):
-        return "none"
-    for name, spec in layers.items():
-        algo = str((spec or {}).get("quant_algo", "")).lower()
-        if algo == "fp8" and (".self_attn." in name or ".linear_attn." in name):
+    if isinstance(layers, dict):
+        for name, spec in layers.items():
+            algo = str((spec or {}).get("quant_algo", "")).lower()
+            if algo == "fp8" and (".self_attn." in name or ".linear_attn." in name):
+                return "fp8_pertensor"
+    # llm-compressor path: a config_group with format "float-quantized" and per-tensor
+    # strategy (``group_size is None and strategy == "tensor"``) targeting ``.self_attn.``
+    # or ``.linear_attn.`` is per-tensor FP8 (W8A16). Targets are regex strings
+    # (``re:.*\.self_attn\.``), so the substring check uses the unescaped leaf path.
+    for g in (get("config_groups") or {}).values():
+        if not g:
+            continue
+        if not any((".self_attn" in t or "linear_attn" in t) for t in (g.get("targets") or [])):
+            continue
+        w = g.get("weights") or {}
+        if str(w.get("type", "")).lower() != "float" or int(w.get("num_bits", 0) or 0) != 8:
+            continue
+        if w.get("group_size") is None and w.get("strategy") == "tensor":
             return "fp8_pertensor"
     return "none"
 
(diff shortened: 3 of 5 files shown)


salekseev:
Thank you for this change. We used it as the base for per-channel compressed-tensors FP8 support, and we found one gap.

The new branch in `_attn_quant` tests `strategy == "tensor"`. A checkpoint that uses `strategy: "channel"` does not match the test, so it still gets `"none"`.

We checked `unsloth/Qwen3.6-35B-A3B-NVFP4-Fast` against your head commit:

| tree | attn_quant |
|---|---|
| main af71ba4 | `none` |
| #390 head | `none` |

The value does not change. This value is a routing key, so the checkpoint never reaches `_iter_weights_attn_fp8`.

The correction is one token:

```python
if w.get("group_size") is None and w.get("strategy") in ("tensor", "channel"):
```

This is safe. `Fp8PerTensorLinear.weight_scale` is per output row. A true per-tensor weight holds the same scalar in every row. Both strategies use the same layer, and no kernel change follows.

One more change is necessary with it. `_per_row_scale` in `weight.py` does `scale.reshape(1)`. A per-channel scale is `[rows, 1]`, so the reshape fails:

```
RuntimeError: shape '[1]' is invalid for input of size 248320
```

We have both changes on a branch with tests. Tell us if you want them in this pull request, and we will send them. We will not open a competing pull request.

The full analysis and the measurements are in #252.



---

The change is on a branch, if you want to cherry-pick it.

- Commit: [`f4e3270`](https://github.com/salekseev/FreeToken/commit/f4e3270)
- Branch: [`upstream/ct-fp8-channel-strategy`](https://github.com/salekseev/FreeToken/tree/upstream/ct-fp8-channel-strategy)
- Base: this pull request's head (`6b78a6c`)

The commit holds the one-token change and a new test file, `tests/models/test_qwen3_5_moe_config.py`. The test covers the geometries that must route to the per-tensor fp8 layer, and the geometries that must not: a group or block scale, a `group_size` with a per-channel name, 4-bit weights, int weights, and a group that targets only the experts.

Use it, change it, or ignore it, as you prefer.



---

One more observation, for your consideration. It is not a defect in this pull request.

The new branch does not consult the checkpoint's `ignore` list. compressed-tensors gives `ignore` precedence over a group's `targets`, and llm-compressor often pairs a broad target with an `ignore` list that carves modules back out. A module in `ignore` is bf16 on disk. If the branch claims such a module as fp8, the loader builds an fp8 layer for a bf16 weight, and the load stops on the dtype check.

`models/qwen4_exp/config.py` already does this for its own family. It reads `get("ignore")` and derives each flag from a probe module name. The comment there gives the reason: derive every flag from that list instead of assuming the split.

The matcher must differ, though. `qwen4_exp` uses `fnmatch`, which suits ModelOpt's plain module names. compressed-tensors writes regex entries such as `re:.*lm_head`, and `fnmatch("lm_head", "re:.*lm_head")` returns `False`. So a compressed-tensors check needs its own match, not a reuse of that helper.

Neither checkpoint we tested is affected. Neither lists these modules in `ignore`, and we have not seen a checkpoint that hits this. We mention it only because the same gap applies to every group the new branch inspects.

salekseev:
The change is on a branch, if you want to cherry-pick it:

- Commit: [`f4e3270`](https://github.com/salekseev/FreeToken/commit/f4e3270)
- Branch: [`upstream/ct-fp8-channel-strategy`](https://github.com/salekseev/FreeToken/tree/upstream/ct-fp8-channel-strategy)
- Base: this pull request's head (`6b78a6c`)

The commit holds the one-token change and a new test file, `tests/models/test_qwen3_5_moe_config.py`. The test covers the geometries that must route to the per-tensor fp8 layer, and the geometries that must not: a group or block scale, a `group_size` with a per-channel name, 4-bit weights, int weights, and a group that targets only the experts.

Use it, change it, or ignore it, as you prefer.

Romeriz:
Hi -- thanks for this. Before I open a complementary PR, let me flag the overlap and a
couple of gaps so we don't ship conflicting changes in the same functions.

**What overlaps (same base `main`, so the hunks collide):**
- `_CT_NVFP4_FUSE`: I also add the identical `.mlp.shared_expert.gate_up_proj` gate|up entry.
- `_expert_quant`: I also recognize compressed-tensors MoE -> `nvfp4`.
- Per-expert llm-compressor bank naming: I map `weight_packed` -> `weight`,
  `weight_global_scale` -> `weight_scale_2` with `global_reciprocal` too (I use a separate
  per-expert CT spec + a stacked spec selected by probing the weight map, rather than one
  widened regex + a config reciprocal flag).
- `iter_weights` MoE-CT routing is where we differ architecturally: you gate the CT reader
  to `not _has_moe_experts` (MoE CT goes to the fp8-aware modelopt branch); I keep MoE CT
  in the CT reader and teach it to skip experts, dequant an fp8/bf16 GDN, dequant NVFP4
  `embed_tokens`, and keep `lm_head` native when `lm_head_quant == "nvfp4"`.

**Why I don't think either patch alone covers the other's models:**
- The exports I fixed (AEON-7/Qwen3.6-35B-A3B-heretic-NVFP4, Ttimms/KAT-Coder-V2.5-Dev-REAP-50-NVFP4A16,
  doth4580/Kwaipilot-KAT-Coder-V2.5-Dev-NVFP4-MIXED) keep the whole dense side NVFP4-
  `.weight_packed` (attn + shared expert), so routing them to the modelopt reader (which
  reads `.weight`-named NVFP4) doesn't parse them; and AEON is a single 23 GB file with no
  index, where an index-only reciprocal probe returns the modelopt default (no
  reciprocal) -> wrong dequant.
- Your Ornith-style export (FP8-attn config group + NVFP4 experts) needs the fp8 dense
  routing I don't implement.

**Suggestion:** rather than merge overlapping diffs, reconcile the MoE-CT dense routing
per variant (NVFP4-`.weight_packed` dense -> the CT reader; FP8-`config_groups` dense ->
the fp8 reader), and make the reciprocal probe fall back to reading the single-file
safetensors header. I'm happy to rebase my branch on yours once it's in a state you're
happy with, or fold my delta in. Want me to post the precise diff-by-hunk comparison, or
shall I just open my PR referencing this one so the maintainers can compare?

(tagging for visibility; happy to take this to the developer Slack if easier)

Sam-Izdat:
@salekseev @Romeriz - Thank you both for feedback and corrections. I agree that we should consolidate, and either way is fine by me. I can fold in your changes if you like, or you can take what you need from here for one of your PRs, and I can close this one -- just let me know. Just a heads up: I'm sitting on an ancient workstation with a 3060 for compute, so that's the extent of the hardware testing I can do, if you need to target more demanding models.

salekseev:
@Sam-Izdat -- please don't close this on my account. Two of my four commits turned out not to touch anything you or @Romeriz touch, so I've just opened them as standalone PRs against `main`:

- #415 -- `_per_row_scale` accepting a `[rows, 1]` per-channel scale
- #416 -- an `Fp8LMHead`, so a natively-FP8 head loads instead of dying on `Unexpected keys in state_dict: ['lm_head.weight_scale']`

I checked rather than assumed: this PR's diff has no `_per_row_scale`, no `Fp8LMHead` and no `_lm_head_quant` in it, and neither appears in Romeriz's overlap list. They also don't care which way the MoE-CT routing question lands, which is why I pulled them out -- they can be reviewed while the interesting argument continues.

The `strategy: "channel"` token from my earlier comment is a change to *your* branch, not a PR of mine. Take it or leave it; I'm not going to open something that competes with this.

I'm deliberately not weighing in on the routing overlap between you two. I've only exercised this against two checkpoints and I'd be guessing about the exports Romeriz is targeting. That call probably wants @jason-fxz, since as far as I can tell none of the three of us can merge anything here.

### One correctness thing, because it's easy to lose in a structural discussion

Romeriz's point about the index probe is right, and I think it's sharper than it came across. `_nvfp4_global_reciprocal` says the right thing in its own docstring -- the ground truth is the on-disk tensor naming -- but it reads that naming out of `model.safetensors.index.json`:

```python
if not idx_path.exists():
    return False  # single-file or no index -- fall back to safe default
except Exception:
    return False  # any read error -- don't reciprocate (modelopt default is safer)
```

For a single-file llm-compressor export, `False` picks the ModelOpt convention, so the reciprocal is skipped and the weights come out wrong with no error anywhere -- just degraded output. The comment's "safer" holds for ModelOpt but inverts for the exporter this PR exists to support, and a silent wrong-numerics default is a rough one to debug from a bug report. Romeriz's AEON case (single 23 GB file, no index) hits it directly.

Worth considering: you already have the ground truth in your hand at read time. In my commit for the packed shared expert I pick the parts reader because the tensor I'm looking at literally ends in `.weight_packed`, straight out of `f.keys()` on the open file -- no index, no config flag, no heuristic, and it can't be wrong for a single-file export because there's nothing to probe. Reading the safetensors header directly, as Romeriz suggested, would get you the same guarantee while keeping the config flag.

### Offer

I've got an RTX 4080 SUPER (16 GiB, sm_89) with both `unsloth/Qwen3.6-35B-A3B-NVFP4-Fast` and `nvidia/Qwen3.6-35B-A3B-NVFP4` on disk, plus needle-recall, BFCL tool-calling and evalscope harnesses already wired up for this engine. If a 3060 is the constraint, point me at a branch and I'll run the 35B-class checkpoints against it and post numbers. Same offer to you, @Romeriz -- I can't obtain your NVFP4 exports quickly, but if the dense-side routing question comes down to "does this actually load and generate sensibly", I can be the box that answers it.

salekseev:
Ran this on hardware. Took @Romeriz's suggestion and read a header instead of an index -- more on that below, because there's a reproducible bug in it.

## #390 loads and serves a real mixed-precision export

I pulled `primitive-ai/Ornith-1.5-35B-A3B-mixed-NVFP4-FP8` (22.55 GiB) specifically because it exercises **this PR and nothing of mine**: its fp8 group is `strategy: "tensor"` with real module regexes, so your detection matches it as shipped, and its `lm_head` is in `ignore`, so my `Fp8LMHead` never fires. Detector output:

```
attn_quant='fp8_pertensor'  expert_quant='nvfp4'  lm_head_quant='none'  dense_mlp='none'
```

It loads through `Loading mixed-fp8 weights`, comes up on the `fi` attention backend, and generates coherently. RTX 4080 SUPER, 16 GiB, sm_89, TP=1, 3600 expert slots, fp8 KV at 262,144 tokens.

| | this checkpoint | `unsloth/...-NVFP4-Fast` | `nvidia/...-NVFP4` |
|---|---|---|---|
| prefill @8k | **3688 tok/s** (8/8) | 3311 | 3248-3618 |
| decode @1200 | 106.95 | 121.26 (n=4) | 123.4 (n=4) |
| longctx @32k | 101.63 | -- | 117.1 |
| needle recall | **9/9** to 176k | 9/9 to 248k | 9/9 to 248k |
| BFCL parallel | 92.50 | 95.00 (mean of 2) | 94.50 (mean of 2) |
| free VRAM | 1.46 GiB | 2.13 GiB | -- |

Prefill is the fastest of the three. BFCL 92.50 against unsloth's 95.00 is 5 flipped entries out of 200, inside the 8-flip noise floor I've measured on this category, and it's a single run where the others have two or three -- so I'd call that unresolved rather than a regression. It's also a different base model (Ornith-1.5, not Qwen3.6), scored with its own tokenizer and chat template.

**Edit:** the unsloth column was a single 120.05 run when I first posted this. I've since run it four times -- 120.05 / 123.3 / 121.21 / 120.48, mean 121.26 -- so the table now shows that. Worth flagging because the original numbers could be read as "the compressed-tensors fp8 path costs ~2.7% decode", and that isn't supported: 121.26 against 123.4 is 1.7% on the means with overlapping ranges (unsloth's best run beats nvidia's mean). Treat the two as a wash on decode.

Two notes so nobody over-reads the table. The gap to *this* checkpoint is mostly the bf16 `lm_head` -- 0.947 GiB read per decoded token here versus unsloth's native-fp8 0.474 GiB. The only other difference is the shared expert (fp8 here, NVFP4 there, worth roughly +52 MiB/token), so the head is about 90% of the ~525 MiB traffic delta. One consistency check: ΔTPOT is ~1.10 ms (9.35 vs ~8.25), which against ~525 MiB implies ~480 GB/s effective -- the right order for this card at its ~53% achieved bandwidth. That supports the traffic explanation without proving it; primitive-ai is still n=1, and these are two checkpoints differing in more than one variable rather than a controlled experiment. And my build also carries an unrelated fp8-KV change, so don't read these as clean-tree numbers.

## The reciprocal probe is wrong for single-file exports, and here's a repro

Romeriz was right, and it's worse than "falls back to a default" -- the fallback silently picks the *opposite* convention with no error anywhere. Your own docstring already names the ground truth ("the on-disk tensor naming"); the problem is only that `_nvfp4_global_reciprocal` reads that naming out of `model.safetensors.index.json` rather than out of the tensors.

Using the checkpoint above, which is genuinely llm-compressor (`weight_packed` present, `weight_scale_2` absent), I symlinked the six shards into a temp dir with the config and **no index**:

```
expert gate_proj suffixes on disk:
  ['input_global_scale', 'weight', 'weight_global_scale', 'weight_packed', 'weight_scale']
  -> llm-compressor naming (weight_packed): True
  -> ModelOpt naming (weight_scale_2):      False

WITH index:    _nvfp4_global_reciprocal = True     <- correct
WITHOUT index: _nvfp4_global_reciprocal = False    <- wrong
  (dir holds 6 shards, so the ground truth is present -- it is just never read)
```

Same tensors both times. Only the index file differs. `False` means "don't reciprocate", i.e. ModelOpt's convention applied to an llm-compressor checkpoint -- degraded output, nothing raised, and nothing in a bug report to point at. That's Romeriz's AEON case (single 23 GB file, no index) reproduced on a published checkpoint.

Worth noting the `except Exception: return False` path has the same shape, so a permissions error or an unreadable index lands in the same place. If it's useful I'm happy to send a patch that reads the safetensors header directly and keeps the config flag as an override -- but it's your PR, so say the word rather than me pushing at it.

## Where I've landed

@Sam-Izdat -- #415 and #416 are open against `main` and don't touch anything in here, so nothing you do with this PR is blocked on them. Still not opening anything that competes with this one, and still not going to referee the routing question between you and Romeriz; that wants @jason-fxz.

Offer stands on the hardware. I've got this box plus three 35B-class checkpoints on disk and the harnesses wired up, so if either of you wants a branch run through load + needle + BFCL, point me at it.

Romeriz:
@Sam-Izdat — let's consolidate, and I'll take you up on "take what you need." Concretely:

- I'll rebase #413 onto `main` and cherry-pick your compressed-tensors commits onto it as the base, keeping your authorship intact (the work is yours; folding it into my PR is just to give maintainers a single branch to review). Then you close #390 pointing at #413.
- The one piece neither of us should keep as-is is the MoE-CT routing: I'll replace both routing carve-outs with a per-variant dispatch on the dense-side storage (`.weight_packed` NVFP4 dense MoE-CT -> `_iter_weights_compressed_tensors`; FP8 `config_groups` / per-tensor-scale dense MoE-CT -> the fp8 reader). That is the reconciliation from my earlier comment and the thing that lets both export families load off one branch.
- salekseev's reciprocal repro settles it: keep #413's header-based reciprocal (spec-level `global_reciprocal`, read via `_weight_map` from the safetensors header when there's no index) and drop the index-probe config flag, so that single-file bug class doesn't survive the merge.
- #415/#416 stay standalone (no overlap with this route) - I'll review/+1 them.

If that arrangement works, I'll post the consolidated branch here and we can ask salekseev to run it on his 4080 SUPER. If you'd rather fold my delta into #390 and keep that as the vehicle, say so - either way works for me, I just don't want two overlapping PRs sitting in front of the maintainers.

Romeriz:
@salekseev - thank you for the repro; a same-tensors/index-vs-no-index demo makes the point much more crisply than my AEON case alone. To be explicit about what it means for the consolidation Sam and I are arranging: if we fold #390 into #413, we keep the header-based reciprocal (spec-level `global_reciprocal`, read through `_weight_map` from the file header when there is no index) and drop the index-probe config flag - so that silent wrong-convention default does not survive the merge.

And yes to the hardware offer, once the consolidated branch exists. Your 4080 SUPER is the only box in this thread that can exercise the FP8/mixed side (`unsloth/Qwen3.6-35B-A3B-NVFP4-Fast`, the Ornith mixed export) plus the nvidia-modelopt regression. When it is ready I'll post the branch here and would welcome your load + needle + BFCL numbers, with the caveat noted that the branch will carry the per-variant dense routing I described.

Agreed on #415/#416 standing alone - no overlap with the routing work, and I'll review both.

salekseev:
@Romeriz @Sam-Izdat -- before you spend effort on the consolidation, `main` moved under all three of us and it changes what #413 needs to carry. #418 and #427 landed a quantization refactor with a new `layers/quantization/` package, and a lot of what we've been arguing about is now handled there.

I probed `CompressedTensorsConfig` from today's `main` (`fb7f732`) against the checkpoints in this thread:

```
unsloth/Qwen3.6-35B-A3B-NVFP4-Fast
  self_attn.q_proj            -> QuantScheme(fp8_tensor, group=(1, -1))     # per-CHANNEL
  lm_head                     -> QuantScheme(fp8_tensor, group=(1, -1))
  mlp.shared_expert.gate_proj -> QuantScheme(nvfp4, group=(1, 16))

primitive-ai/Ornith-1.5-35B-A3B-mixed-NVFP4-FP8
  self_attn.q_proj            -> QuantScheme(fp8_tensor, group=(-1, -1))    # per-TENSOR
  lm_head                     -> None                                       # excluded via `ignore`
```

Relevant to what each of us was carrying:

- **`strategy: "channel"`** is a first-class scheme now (`FP8_CHANNEL`). The one-token widening I was offering for this PR is obsolete.
- **`ignore` beats `targets`**, via `ct_set(..., class_names=False)`, and it understands class-name targets like `"Linear"`. That's more complete than the hardening I'd written, and it's the thing I flagged on #252 as the actual blocker for `ornith-ai/Ornith-1.5-35B-A3B-FP8` -- which uses `targets: ["Linear"]` and got `attn='none'` from every version of the old detector.
- **Per-row weight scales** are native: `fp8_tensor.py` allocates `weight_scale` as `[out_features]` fp32.
- **The reciprocal** -- I haven't found an index-probe equivalent in the new config layer at all, which is the outcome you both wanted. Worth checking against `_weight_map` before you port anything, but it looks like the bug class we were discussing may simply not exist there.

`qwen3_5_moe` hasn't been migrated onto any of this. Its `config.py` still computes `_expert_quant` / `_attn_quant` / `_lm_head_quant` (lines 42/71/107, consumed at 169/172/180) and `_iter_weights_attn_fp8` still routes off those strings. On today's `main` the legacy detectors return `attn='none'` for **both** compressed-tensors checkpoints above, so the reader never reaches the fp8 dense path even though the config layer describes it perfectly. `qwen4_exp` is fully migrated with zero legacy detector references and reads like the intended template.

So my read is that the per-variant dense routing you were going to write, @Romeriz, may not need writing as a routing carve-out at all -- the variant is already a `QuantScheme` per module. It might reduce to migrating `qwen3_5_moe`'s config + reader onto `config.quant` and letting the scheme decide, which would cover the NVFP4-`.weight_packed` dense family and the FP8-`config_groups` family through one path without either carve-out.

I've closed #416 on the back of this -- its scaffolding is gone from `main` (the layer classes moved out of the kernel modules, and `model.py` already passes `quant_config`+`prefix` to a single `ParallelLMHead`), and the new config layer delivers what it was for. #415 I've left open with a note asking whether it's wanted as a stopgap for the still-live legacy reader or should be folded into the migration.

Offer unchanged and now easier to take up: 62 GB RAM as of today, a 4080 SUPER, and `unsloth/...-NVFP4-Fast`, `nvidia/...-NVFP4` and the Ornith mixed export all on disk with load/needle/BFCL harnesses wired up. Point me at a branch and I'll run all three, modelopt regression included. And if nobody has claimed the `qwen3_5_moe` migration I'm glad to take it, since I can validate it against every family in this thread -- say so and I won't duplicate whatever you two are already doing.

I have filed the migration as #437, so it is not buried in this thread and we do not duplicate each other -- @Romeriz if the per-variant dense routing you described is something you'd rather own, say so there and I'll stay out of `weight.py`. It's scoped to fp8/NVFP4 with the legacy detectors kept as the AWQ/int4 fallback, because `layers/quantization` has no dialect for those yet (three of my checkpoints raise `NotImplementedError` on `QuantConfig.from_hf` -- details in #437, and the int4 side is #396).

salekseev:
Filed the migration as #437 so it isn't just buried in this thread, and so we don't duplicate each other — @Romeriz if the per-variant dense routing you described is something you'd rather own, say so there and I'll stay out of `weight.py`. It's scoped to fp8/NVFP4 with the legacy detectors kept as the AWQ/int4 fallback, because `layers/quantization` has no dialect for those yet (three of my checkpoints raise `NotImplementedError` on `QuantConfig.from_hf` — details in #437, and the int4 side is #396).

PR #391, opened 2026-09-05
fix(tokenizer): accept the OpenAI developer role (map to system for templates without it)

gdevenyi:
## What this fixes

A request whose system prompt uses the OpenAI `developer` role (the current spelling of `system`; pi sends it by default for `openai-completions` providers, `compat.supportsDeveloperRole: true`) fails with `400 could not encode request: Unexpected message role.` on every model whose chat template does not spell the role, which is most of them (Qwen3.8-Flash-Next, Qwen3.x, GLM, ...). The template raises, the server relays it.

`TokenizeManager._render` now maps `developer` to `system` before `apply_chat_template`, unless the template mentions `developer` itself (gpt-oss renders the two roles differently and keeps its input). The mapping copies the message; the request is not mutated.

## Testing

- `tests/tokenizer/test_tokenize.py::test_developer_role_maps_to_system_unless_the_template_knows_it`.
- Live on a Qwen3.8-Flash-Next TP=2 server: `{"role": "developer", ...}` went from the 400 above to the same `prompt_tokens` as the `system` spelling.

### DIFF
--- a/python/freetoken/tokenizer/tokenize.py
+++ b/python/freetoken/tokenizer/tokenize.py
@@ -46,6 +46,19 @@ def resolve_thinking_mode(chat_template_kwargs: dict[str, Any] | None, tools: An
 _EFFORT_PROBE_MESSAGES = [{"role": "user", "content": "ping"}]
 
 
+def _map_developer_role(
+    messages: list[dict[str, Any]], chat_template: str | None
+) -> list[dict[str, Any]]:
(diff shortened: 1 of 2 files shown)


gdevenyi:

**Result: 1206 passed, 350 skipped, **no new failures**.**

The +1 over main are this PR's own tests, and they ran (not skipped).

gdevenyi:

PR #393, opened 2026-09-05
feat(server): OpenAI protocol coverage -- sampling extras, n, echo/suffix, usage details, tokenize/detokenize/metrics

gdevenyi:
## Summary

Closes the most common gaps between FreeToken's OpenAI-compatible surface and vLLM / SGLang / llama.cpp's server. `docs/openai_api.md` has the four-way table of endpoints, request parameters and response fields (honoured / accepted / 400), and lists what is deliberately not implemented.

**Sampling extras** (`/v1/chat/completions` and `/v1/completions`): `min_p`, `presence_penalty`, `frequency_penalty` (over generated tokens), `repetition_penalty` (prompt + generated, HF semantics), `logit_bias` (token id → bias, clamped to [-100, 100]), `min_tokens` (no EOS / stop token before N generated tokens), `stop_token_ids`, `include_stop_str_in_output`, `skip_special_tokens` (per request; default off here because the reasoning and tool parsers read the tags). Out-of-range values answer 400 with the field named.

Implemented as logits processors (`engine/sample.py`): `Sampler.prepare` builds a per-batch `LogitsPlan` on the host from the requests' `SamplingParams` and token histories, only for the rows that asked; `apply_logits_processors` applies it to a float32 copy of the logits before the sampling kernel (repetition → presence → frequency → logit_bias → min_tokens mask → min_p, vLLM's order). A batch without them takes exactly the old path, and sampling already runs outside the CUDA graph, so no capture changes. Cost measured on an Ada card with a 3k-token history: 0.97 ms per step for a batch that uses them (0.32 ms on the device), 0.08 ms on the plain path.

**`n` (1..16)** on both routes: one generation per choice submitted together (the prefix cache serves the shared prompt), results gathered; streams interleaved into one SSE response with the choice index, one usage chunk (prompt counted once, completions summed) and one `[DONE]`; a client disconnect aborts every uid of the fan-out.

**Completions**: `echo` (the prompt leads the text, or the first chunk), `suffix` as a fill-in-the-middle prompt on models whose vocabulary has the FIM tokens (Qwen family; 400 otherwise).

**Chat**: `continue_final_message` (the final assistant message is continued, no generation prompt), `request_id` echoed as the response id, `seed` / `user` accepted, `system_fingerprint: null` in responses and chunks.

**Usage**: `completion_tokens_details.reasoning_tokens` — the detokenizer counts the tokens up to and including the reasoning end tag and reports it on the finished reply.

**Routes**: `POST /tokenize` and `/detokenize` (also under `/v1/`; the vLLM / SGLang shape, `messages` render through the chat template so `count` equals a generation's `prompt_tokens`), `GET /metrics` (Prometheus text of `/v1/stats`), `GET /version`.

Not in this PR, on purpose: constrained decoding (`response_format` json / json_schema, grammars: needs a grammar engine on the sampling path), prompt logprobs and `echo` + `logprobs` (prefill logits), a per-request `seed` (one batched sampling kernel), embeddings / rerank / score / audio.

## Test plan

- [x] `tests/engine/test_logits_processors.py` (pure torch on the CPU: each processor, the plan builder, the greedy path), `tests/tokenizer/test_detokenize_extras.py` (stop-string keep, per-request `skip_special_tokens`, reasoning token count), `tests/server/test_openai_extras.py` (n fan-out non-stream and stream, echo, suffix with and without FIM tokens, prompt-major choice order, the extras reaching `SamplingParams`, 400s, `continue_final_message`, `request_id`, usage details, tokenize / detokenize, metrics exposition)
- [x] `tests/server`, `tests/tokenizer`, `tests/scheduler`, `tests/engine`: 790 passed on the CPU
- [x] GPU micro-check of the processors and the sampler with the real vocabulary (248,320) on RTX 6000 Ada
- [x] Served Qwen3.8-Flash-Next (RadixArk NVFP4, offload backend, TP=1) from this branch and ran a live probe of every field and route: 24/24 (penalties change a repetitive output, logit_bias steers the answer, min_tokens forces 2 → 35 tokens, stop_token_ids on `.` stops at 18 tokens, include_stop_str_in_output keeps the stop word, min_p / seed / user accepted, n=2 non-stream and stream, echo, suffix as FIM (`return a + b`), continue_final_message, request_id as the response id, reasoning_tokens 39 of 44 completion tokens with thinking on, tokenize / detokenize round trip and a rendered-messages count equal to the generation's prompt_tokens, /metrics, /version, five 400s), 8-question quality probe unchanged (6/8), decode unchanged (61.7 tok/s single-stream / 166 at 8 concurrent, the same as main on that card)

### DIFF
--- a/docs/openai_api.md
+++ b/docs/openai_api.md
@@ -1,0 +1,101 @@
+# OpenAI-compatible API: endpoints and parameters
+
+What `ft serve` accepts on its OpenAI-shaped routes, what it ignores, and what it refuses,
+next to the same surface in vLLM, SGLang and llama.cpp's server. "honoured" means the value
+changes the generation; "accepted" means the request is not rejected but the value has no
+effect; "400" means the server refuses the request with a clear message.
+
+## Endpoints
+
+| route | FreeToken | vLLM | SGLang | llama.cpp |
+|---|---|---|---|---|
+| `POST /v1/chat/completions` | yes | yes | yes | yes |
+| `POST /v1/completions` | yes | yes | yes | yes |
+| `GET /v1/models`, `GET /v1/models/{id}` | yes | yes | yes | yes |
+| `POST /v1/responses` (+ get, cancel) | yes | yes | yes | yes |
+| `POST /v1/messages` (Anthropic), `count_tokens` | yes | yes | no | yes |
+| `POST /tokenize`, `POST /detokenize` (also under `/v1/`) | yes | yes | yes | yes (`/tokenize`, `/detokenize`) |
+| `GET /metrics` (Prometheus) | yes | yes | yes | yes |
+| `GET /version` | yes | yes | yes | no (`/props`) |
+| `GET /health` | yes | yes | yes | yes |
+| `GET /v1/stats`, `/v1/cache/*`, `/v1/requests`, `/v1/admin/*` | yes (FreeToken control plane) | `/metrics` only | `/get_server_info`, `/flush_cache` | `/slots`, `/props`, `/metrics` |
+| `POST /v1/embeddings`, `/rerank`, `/score`, `/pooling` | no (generative models only) | embedding / reranker models | embedding / reranker models | yes, any model with pooling |
+| `POST /v1/audio/*`, `/v1/files`, `/v1/batches` | no | transcriptions, batch runner (offline) | transcriptions, files, batches | no |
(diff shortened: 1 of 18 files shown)


gdevenyi:

**Result: 1235 passed, 350 skipped, **no new failures**.**

The +30 over main are this PR's own tests, and they ran (not skipped).

gdevenyi:

PR #399, opened 2026-09-06
fix(cpu-moe): honor padded fp8 scale strides, add avx512f tier and float64 parity (builds on #36)

ChenyuHeee:
## Relationship to #36

**This builds directly on @gdevenyi's #36 and includes his commit unchanged** (`4a04c9b`, +360 lines — the format definition, bank plumbing, ABI probe and benchbw integration). My three commits on top are +262 lines. If you would rather land #36 first, these three cherry-pick cleanly onto it and I am happy to send them there instead — I opened this separately only because the stride bug below is a correctness issue that would otherwise ship silently.

Credit for the feature is his. What follows is a bug fix and test hardening.

## 1. The scale bank's trailing dimension is padded (correctness)

`kernel/aot_models.py` pads the scale bank's last dimension so per-expert rows land on a 16-byte boundary for the fused copy:

```python
def fp8_block_scale_pad(rows: int, cols: int) -> int:
    """Trailing scale-bank dim padded so per-expert row bytes are 16B-aligned (fused copy)."""
    while (rows * cols * 2) % 16:
        cols += 1
    return cols
```

So the row stride is **not** `ceil(K/128)`. #36 recomputes it from the unpadded block count in two places (`cpu_moe_ext.cpp` and the shape assert in `cpu_executor.py`), which reads scales at the wrong offset whenever padding applies.

**Qwen3.8 is the case already documented in the tree** — `models/qwen3_5_moe/weight.py` records `down_scale` as `20x5` bf16 = 200 B, and `200 % 16 != 0` pads cols 5 → 6. The assert demands `(E, 20, 5)` and fails; with asserts off the kernel silently produces wrong numbers.

Neither the benchbw fp8 workload (H=2048, I=512) nor DSV4-Flash (H=7168, I=2048) trips it — both happen to give `rows*cols` divisible by 8, which is why #36's own tests are green.

The fix reads the stride off the tensor and threads it through to the kernel. Reverting just the C++ half turns 7 tests red (4 on a CPU-only host).

Worth noting: the Triton kernel was **already** stride-aware. Feeding it padded banks keeps the GPU cross-check green, so the CPU path was the only one getting this wrong.

## 2. The parity test could not see accumulator precision loss

The existing check compares two fp32 accumulations (CPU vs Triton) at `rel < 2e-2`, so common-mode error cancels and a kernel with ~1e-3 error passes unnoticed.

Added a float64 reference — `torch.float8_e4m3fn` dequant with the bf16 block scales, GEMV in float64 — over mixed-sign full-range e4m3 weights, zero-mean activations, and ragged K. Measured error is 2.674e-3 against a 5e-3 bound (the executor stores bf16, so the floor is not tighter than that).

No existing assertion was weakened; the `2e-2` check is untouched.

Also addressed the two Copilot comments on #36: the ISA-tier test now actually sets `FREETOKEN_CPU_MOE_ISA`, and `test_subnormal_and_sign_decode_exactly`'s docstring no longer claims to validate the C++ decoder when it only exercises a PyTorch round-trip.

## 3. An avx512f tier for CPUs without AVX512-BF16

#36's fast path is `_mm512_dpbf16_ps`, so a CPU with AVX-512 but no BF16 extension falls all the way to the AVX2 tier — 8-wide fp32, 512-bit registers unused. That is Skylake-SP, Cascade Lake and Ice Lake-SP.

Added an `avx512f` tier between them, following the existing per-function `__attribute__((target(...)))` + runtime dispatch structure. All three reachable tiers are bit-identical across all 256 e4m3 codes.

Measured on Xeon Platinum 8368 (Ice Lake-SP, no `avx512_bf16`), three repetitions each:

| | avx2 | avx512f | |
|---|---|---|---|
| single-thread, cache-resident | 3.31 GB/s | 3.94 GB/s | **+19%**, intervals disjoint |
| `ft bench bw --dtype fp8`, 75 threads | 60.4 GB/s | 62.7 GB/s | +4%, intervals **overlap** |

At 75 threads the kernel is memory-bound and the tier is largely washed out — the project's own benchmark cannot resolve it. The single-thread number is the honest figure.

## 4. Tests now cover the code they ship

Before this, deleting the entire `avx512f` branch from `select_fp8dot` left every test passing, because `isa_name()` reports `select_dot()` (the bf16 chain) rather than the fp8 dispatch — and no test ran numerics under a forced tier. Injecting a ragged-tail bug into `dot_fp8_block_avx2` was invisible for the same reason.

Each reachable tier now runs the float64 parity GEMV, and the fp8 dispatch is observable. Verified by mutation:

| mutation | before | after |
|---|---|---|
| drop the `avx512f` dispatch branch | all green | `assert 'avx2' == 'avx512f'` |
| ragged-tail scale bug in `dot_fp8_block_avx2` | all green | rel = 0.257 |
| kernel ignores the passed stride | — | 7 failed (4 CPU-only) |
| break only `scalar` | — | caught, that tier only |
| break only `avx512f` | — | caught, that tier only |

The module-level CUDA guard also moved onto the tests that need a GPU — all ten were skipping on a CPU-only host, which is exactly where a CPU MoE kernel matters most.

## Known limits, stated up front

- **The `avx512bf16` fp8 tier has never executed here.** This host lacks the ISA and I have no SDE, so that path is unvalidated on hardware — and the code comment says it is the tier that matters. A mutation that halves its output goes undetected on this machine. Someone with Emerald Rapids or Sapphire Rapids should confirm it; @gdevenyi's 6526Y would do.
- **Tier verification is name-mediated.** A tier that is numerically correct but silently does not use AVX-512 would pass. Correctness tests cannot distinguish that; only the throughput number argues the tier earns its place.
- CPU-only runs exercise `num_threads=1` only. Row splitting is format-independent, so the risk is low, but multi-threaded fp8 is covered solely by the GPU cross-check.

## Test results

```
CUDA:      12 passed, 1 skipped     (avx512bf16 tier skipped — host lacks the ISA)
CPU-only:   9 passed, 4 skipped
tests/moe: 130 passed, 7 skipped, 1 failed
```

The one failure is `test_cpu_moe_q4_0.py::test_cpu_decode_q4_0_matches_ggml_mmvq`, a GGUF JIT compile error under gcc 15 (`need 'typename' before ...` in torch headers). **It fails identically on `main`** — unrelated to this PR.

Built with gcc 15.2.1 / binutils 2.44. Function-level AVX-512 target attributes need binutils >= 2.36; the gcc 11 / binutils 2.35 on RHEL 9 cannot assemble them.

### DIFF
--- a/python/freetoken/kernel/csrc/cpu_moe/cpu_moe_ext.cpp
+++ b/python/freetoken/kernel/csrc/cpu_moe/cpu_moe_ext.cpp
@@ -263,6 +263,134 @@ inline float e4m3_decode(uint8_t v) {
   return sign * (1.0f + man / 8.0f) * std::ldexp(1.0f, (int)exp - 7);
 }
 
+// ------------------------ block-FP8 (W8A16) dequant ------------------------
+// DeepSeek-V3-style block-fp8 experts (Qwen3.5/3.6-FP8, GLM, DSV4 dense): fp8-e4m3
+// weights row-major [rows, K] plus one bf16 scale per 128x128 weight block. The tensor
+// is named ``weight_scale_inv`` but it multiplies; the reference is the Triton decode
+// GEMV (kernel/triton/fp8_blockscale_moe.py):
+//
+//     acc += sum(w * a  over one 128-wide K block) * scale[row/128][kb]
+//
+// K is contiguous per output row -- unlike mxfp4's transposed bank -- so this maps
+// straight onto a bf16 dot product. e4m3 -> bf16 is *exact* (3 mantissa bits into 7,
+// and bf16's exponent range covers e4m3's whole 2^-9..448 span), so the AVX-512 path
+// widens the weights in-register and feeds _mm512_dpbf16_ps against the bf16
+// activations. That reproduces the reference's fp32 products exactly; only the
+// summation order differs, the same latitude the bf16 dot already takes.
+//
+// Bit math for the normal range (exp != 0), from byte b = s<<7 | e<<3 | m:
+//   bf16 = (b & 0x80) << 8  |  (((b & 0x7F) << 4) + (120 << 7))
+// because e4m3 bias 7 -> bf16 bias 127 is a constant +120 on the exponent field once
+// the magnitude is shifted into place. exp == 0 is subnormal (value = m * 2^-9) and
+// does not follow that rule, so those lanes are blended in from a small table.
+using fp8dot_fn = float (*)(const uint8_t*, const bf16_t*, const bf16_t*, int, const float*);
+
+constexpr int FP8_BLK = 128;  // K elements covered by one weight scale
+
+// e4m3 subnormals: m * 2^-9, as bf16 bit patterns (index = mantissa, 0..7).
+alignas(64) const uint16_t kE4M3SubBf16[32] = {
+    0x0000, 0x3B00, 0x3B80, 0x3BC0, 0x3C00, 0x3C20, 0x3C40, 0x3C60,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+// Same values as fp32, for the AVX2 path.
+alignas(32) const float kE4M3SubF32[8] = {
+    0.0f, 1.0f / 512, 2.0f / 512, 3.0f / 512, 4.0f / 512, 5.0f / 512, 6.0f / 512, 7.0f / 512};
+
+float dot_fp8_block_scalar(const uint8_t* w, const bf16_t* x, const bf16_t* s, int K,
+                           const float* e4m3) {
+  float acc = 0.0f;
+  for (int b = 0, k0 = 0; k0 < K; ++b, k0 += FP8_BLK) {
+    const int k1 = std::min(K, k0 + FP8_BLK);
+    float blk = 0.0f;
+    for (int k = k0; k < k1; ++k) blk += e4m3[w[k]] * bf16_to_f32(x[k]);
+    acc += blk * bf16_to_f32(s[b]);
+  }
+  return acc;
+}
+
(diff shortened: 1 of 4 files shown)


jason-fxz:
@ChenyuHeee 

**Request changes.** The kernel checks out: I ran the suite on a Sapphire Rapids (avx512_bf16) + H100 box, so the bf16 tier you could not run passed here, and all 256 e4m3 codes decode exactly at every tier. The problems are integration with `main`, not the kernel.

1. **Rebase onto current `main` first.** After the quant refactor, `layers/quantization/moe/fp8_block.py` still has `cpu_ok=False` and no `cpu_format`, so `--moe-strategy cpu|hybrid` raises `KernelSelectionError` and this PR is unreachable from the CLI. Meanwhile adding `fp8_block` to `benchbw._CPU_MOE_FORMATS` makes a fresh `ft bench bw` recommend hybrid, so a default `ft serve` of a block-fp8 checkpoint fails to boot. Fix: `cpu_format = "fp8_block"` and `cpu_ok=True` in that file, same commit as the benchbw change. I tried it on the merged tree: builds and passes.

2. **`cpu_moe_ext.cpp:322-356, :1205`:** `dot_fp8_block_avx512bf16` / `as_bh` and the `select_fp8dot` branch are outside `#ifdef CPU_MOE_HAS_AVX512BF16`. On gcc < 10 the whole extension fails to compile. (`select_dot` and your `fp8_isa_name` guard it correctly.)

3. **Pick up #36's head 19e9382** (ISA x bs parametrisation of the GPU parity test, `ex.isa.split("+")[0]`). Those are the 12 cases @gdevenyi reported failing; they do not reproduce here (CPU vs GPU differs by one bf16 ulp at every tier), but the IDs should survive. Cherry-pick keeps his authorship.

4. **`benchbw.py:798` still writes `"version": 4`** and the reader only checks the GPU name, so existing profiles keep `fp8_block -> offload` forever and explicit hybrid gets a fetch cap of 1. Bump the version and ignore older profiles.

PR #408, opened 2026-09-07
feat(kvcache): add nvfp4 kv quantization

ArqAlice:
## Summary

Add opt-in `--kv-cache-dtype nvfp4` KV-cache storage for paged MHA/GQA, hybrid-SWA, and QSA sparse attention.

K/V use packed E2M1 values, one E4M3 scale per 16 values, and one FP32 row scale per token/KV head. QSA keeps its index, pending-ring, and scratch tiers in BF16; only paged K/V rows are quantized.

At head_dim 128, each K/V row uses 76 bytes versus 256 bytes for BF16. The implementation includes pool budgeting/rebuild, CUDA Graph-safe KV writes, Triton restore paths, configuration validation, and a microbenchmark.

## Validation

Hardware: RTX 5090 (32607 MiB), driver 591.86, WSL2/Docker, PyTorch 2.11.0+cu130, Triton 3.6.0.

```sh
python -m pytest tests/engine tests/kvcache \
  tests/scheduler/test_scheduler_kv_usage.py \
  tests/kernels/test_kv_fp8.py \
  tests/kernels/test_kv_nvfp4.py \
  tests/kernels/test_qsa_fp8.py \
  tests/kernels/test_qsa_nvfp4.py \
  tests/kernels/test_triton_attention.py \
  -q --tb=short -m "not slow"

Result: 503 passed, 2 skipped, 2 failed.
The two failures are the existing FP8 parametrizations of
test_extend_paged_attention_decodes_fp8_scales; they reproduce on the unmodified local base with the same mismatch counts and maximum differences.
Also verified with RadixArk/Qwen3.8-Flash-Next-NVFP4 using CPU MoE offload: QSA sparse attention was selected, CUDA Graph capture completed, and OpenWebUI served a 1M-token context successfully.
NVFP4 reduces KV capacity cost, but the current Triton restore path is slower than BF16/FP8 in the included synthetic decode microbenchmark.
```


gdevenyi:
Ran this on **sm_89** (2 × RTX 6000 Ada, TP=2, `nvidia/Qwen3.8-Flash-Next-NVFP4`, qwen4_exp, QSA sparse attention, offload MoE, 2,097,152-token pool). It works, the memory claim holds almost exactly, and the throughput does not move. Numbers against the same server on `--kv-cache-dtype fp8`:

| | fp8 KV | nvfp4 KV |
|---|---|---|
| KV pool | 13.69 GiB/rank | **8.44 GiB/rank** |
| expert residency | 71.7% | **88.2%** |
| single-stream decode | 105.6 tok/s | 105.51 / 105.59 |
| aggregate @ 8 concurrent | 352.3 tok/s | 352.27 / 357.02 |
| TTFT, 1k prompt | 0.68 s | 0.71–0.74 s |
| GSM8K-300, greedy, 1536 max | 98.00% (294/300) | **98.00% (294/300)** |

**Accuracy is free.** 294/300 both ways, same questions, greedy. On this model 4-bit KV costs nothing measurable, which is the result I was least confident about going in and the one most worth having.

**The capacity claim is almost exactly right.** Before running it I sized the win from your 76 B/row figure and predicted 6.04 GiB/rank freed and 88.1% residency. Measured: 5.25 GiB and 88.2%.

**The throughput win is zero, and I think that is architectural on Ada.** Our decode is PCIe-bound on the offload backend, so a 71.7% → 88.2% residency jump cuts expert fetch traffic ~58% and should have shown up. It does not, because nvfp4 dequantises on every attention read and sm_89 has no native FP4 math — the Triton unpack-and-scale in the hot path costs about what the residency saves. On Blackwell, where the dequant is cheaper or native, I would expect your gain to survive; this is not an argument against the PR, it is a note that the benefit is arch-dependent and a reader on Ada should size it as **capacity, not speed**.

That framing is still valuable to us: 5.25 GiB/rank at flat tok/s and identical accuracy is budget we can spend on a larger KV pool or more concurrent requests. Worth saying explicitly in the PR text, because "3.4× smaller KV" reads as a throughput claim and on at least one architecture it is not one.

**Two things for the PR itself.**

It merges onto a current qwen4_exp deploy line with exactly one conflict, a comment in `tests/kernels/test_triton_attention.py` describing the same stub two ways. Trivial, but it will hit anyone carrying #354.

And a caution from my own mistake, in case it saves someone else the run: my first attempt reported that #408 crashed at `_warmup_prefill` on sm_89 with a device-side assert. That was **not** this PR — I had cut the eval branch from a commit carrying an unrelated local bug of mine, and `IndexKernel.cu:111` pointed at my line, not yours. Rebuilt clean, it starts first time. Apologies if that reached you before this did.

PR #411, opened 2026-09-07
fix(server): honor configured output default across APIs

### LINKED ISSUE #395: [Bug] --max-output-tokens is ignored by /v1/chat/completions and /v1/messages (only /v1/responses honours it)
## Summary

`--max-output-tokens` is documented as "Default max output tokens for requests that omit one
(default 32k)", but that default is applied **only by the Responses API**. On
`/v1/chat/completions` and `/v1/messages` a request that omits `max_tokens` gets the
hardcoded `DEFAULT_MAX_OUTPUT_TOKENS = 32768` instead of the configured value, so the flag
silently does nothing on the two OpenAI/Anthropic-compatible paths.

On a server with `--max-running-requests 1` this is not just a cosmetic gap: one runaway
generation occupies the single slot for its full 32k budget and every queued request waits
behind it.

## Where

`resolve_sampling` hardcodes the constant and never sees the server config
(`freetoken/server/generation.py:398,426`):

```python
DEFAULT_MAX_OUTPUT_TOKENS = 32768
...
max_tokens=DEFAULT_MAX_OUTPUT_TOKENS if max_tokens is None else max_tokens,
```

Callers:

| path | passes the configured default? |
|---|---|
| `freetoken/server/responses_api.py:153` | **yes** — `default_max = getattr(state.config, "max_output_tokens", None) or DEFAULT_MAX_OUTPUT_TOKENS` |
| `freetoken/server/openai_api.py:71,521` | no — `max_tokens=req.max_tokens` straight through |
| `freetoken/server/anthropic_api.py:312` | no (the file has no reference to `max_output_tokens` at all) |

`/v1/messages` matters less in practice because `max_tokens` is required by that API, so
clients always send one. `/v1/chat/completions` treats it as optional, so this is where it
bites.

## Repro

Server started with `--max-output-tokens 4096`. `ignore_eos` is used only to guarantee the
generation would otherwise exceed the limit, so the test is deterministic rather than
dependent on the model choosing to ramble:

```bash
ft serve --model <any> --max-output-tokens 4096 --max-running-requests 1 ...

# max_tokens OMITTED -> should fall back to the configured 4096
curl -s localhost:1919/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model": "<served-name>",
  "messages": [{"role": "user", "content": "Say hello."}],
  "ignore_eos": true
}' | jq '{completion_tokens: .usage.completion_tokens, finish_reason: .choices[0].finish_reason}'
```

Expected `4096`. Actual:

```json
{ "completion_tokens": 32767, "finish_reason": "length" }
```

Controls on the same server, same endpoint:

| request | completion_tokens | finish_reason |
|---|---|---|
| `ignore_eos: true`, no `max_tokens` | **32767** | length |
| explicit `"max_tokens": 4096` | 4095 | length |
| no `max_tokens`, model stops on its own | 1327 | stop |

So an explicit `max_tokens` is honoured and the engine enforces limits correctly — only the
configured *default* is not wired into this path. (32767 rather than 32768 is what both this
repro and the benchmark runs below report.)

## Observed impact

Found while benchmarking, not synthetically. BFCL v4 `parallel` (200 entries, 4 client
threads, `--max-running-requests 1`, `--max-output-tokens 4096`); BFCL's OpenAI handler does
not send `max_tokens`. Across two 200-entry runs of the same model family, **3 of 400 entries
emitted 32767 completion tokens**:

| run | entry | completion tokens | latency |
|---|---|---|---|
| A | `parallel_71` | 32767 | 550.7 s |
| A | `parallel_122` | 32767 | 586.6 s |
| B | `parallel_183` | 32767 | 530.6 s |

Consequences with a single running slot:

* **Head-of-line blocking.** In run B the three entries queued behind `parallel_183`
  completed in 525.5 s / 519.8 s / 512.5 s while emitting only 450 / 598 / 860 tokens each —
  ~8.5 minutes of wall clock spent waiting on one runaway.
* **Accuracy loss.** Run A's `parallel_122` runaway is one of that run's 13 scored failures.
* Median completion length in these runs is 858 tokens, so 32767 is ~38x the typical
  request; the configured 4096 would have been ample.

## Suggested fix

Thread the configured default into `resolve_sampling` the way `responses_api.py` already
does — e.g. give it a `default_max_tokens: int = DEFAULT_MAX_OUTPUT_TOKENS` parameter and
pass `state.config.max_output_tokens` from the OpenAI and Anthropic paths. That keeps the
constant as the fallback and makes the flag mean what its help text says on every API.

Happy to send a PR if the shape above is what you would want.

## Environment

FreeToken `0.1.2+gaf71ba432` (nightly wheel from the `beta` channel, commit `af71ba432`);
torch 2.11.0+cu130, CUDA 13.3; RTX 4080 SUPER 16 GB (sm_89); Qwen3.6-35B-A3B 4-bit,
`--moe-backend offload --moe-cache-auto --max-running-requests 1 --max-output-tokens 4096
--sampling-defaults model --reasoning-parser qwen3 --tool-call-parser qwen3_coder`.
Local patches present but unrelated (routed-expert int4 repack paths); the three files above
are unmodified from upstream.

earlvanze:
Closes #395.

Threads the server-level `--max-output-tokens` fallback through the Chat Completions and Completions adapters, as the Responses adapter already does. Explicit request limits still take precedence; absent server configuration retains the 32k fallback.

Validation:
- `python3 -m py_compile` for changed modules
- `git diff --check`
- Focused pytest suite is not runnable in this checkout because pytest/runtime dependencies are not installed.

### DIFF
--- a/python/freetoken/server/anthropic_api.py
+++ b/python/freetoken/server/anthropic_api.py
@@ -33,6 +33,7 @@ from .anthropic_models import (
     AnthropicUsage,
 )
 from .generation import (
+    DEFAULT_MAX_OUTPUT_TOKENS,
     KEEPALIVE,
     ContentDelta,
     GenDone,
@@ -116,6 +117,9 @@ async def handle_anthropic_messages(
         spec = convert_anthropic_to_genspec(
             req, model_sampling,
             reasoning_parser=getattr(state.config, "reasoning_parser", None),
+            default_max_tokens=(
+                getattr(state.config, "max_output_tokens", None) or DEFAULT_MAX_OUTPUT_TOKENS
+            ),
         )
         uid = await submit_generation(spec, state)
     except ValueError as exc:
@@ -303,6 +307,7 @@ def convert_anthropic_to_genspec(
     req: AnthropicMessagesRequest,
     model_sampling: dict[str, Any],
     reasoning_parser: str | None = None,
+    default_max_tokens: int = DEFAULT_MAX_OUTPUT_TOKENS,
 ) -> GenSpec:
     messages, template_tools, parser_tools, ctk = convert_anthropic_prompt(
         req, reasoning_parser=reasoning_parser
@@ -317,6 +322,7 @@ def convert_anthropic_to_genspec(
             ignore_eos=False,
             model_sampling=model_sampling,
             stop=req.stop_sequences,
+            default_max_tokens=default_max_tokens,
         ),
         chat_template_kwargs=ctk,
         template_tools=template_tools,
--- a/python/freetoken/server/generation.py
+++ b/python/freetoken/server/generation.py
@@ -163,6 +163,7 @@ def resolve_sampling(
     ignore_eos: bool,
     model_sampling: dict[str, Any],
     stop: str | list[str] | None = None,
+    default_max_tokens: int = DEFAULT_MAX_OUTPUT_TOKENS,
 ) -> SamplingParams:
     """Map a protocol's sampling fields onto the engine's neutral SamplingParams,
     filling unspecified fields from the checkpoint's recommended defaults."""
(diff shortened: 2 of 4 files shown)


jason-fxz:
Tested end-to-end on top of current main: with `--max-output-tokens 64`, `/v1/chat/completions` and `/v1/completions` without `max_tokens` now stop at the configured limit (63 tokens vs 2032 on main, clamped from the 32k constant); explicit limits unchanged, server tests pass. LGTM.

PR #412, opened 2026-09-07
fix(bench): show how to use custom bandwidth profiles

### LINKED ISSUE #308: ft bench bw -o writes a profile that hybrid never loads (silent fetch cap of 1)
## Summary

`ft bench bw -o /some/path.json` writes a profile that `ft serve --moe-backend hybrid` never finds. The serve logs one WARNING and starts anyway with a **fixed fetch cap of 1** — a different backend configuration from the one the bench recommended — unless `FREETOKEN_BENCHBW_PATH` is also exported.

**Environment:** FreeToken git `4b94bdc` (HEAD 2026-08-30, no local patches), `pip install --no-build-isolation "freetoken[accel] @ git+…@4b94bdc"`; torch 2.11.0+cu130, CUDA 13.0, driver 595.71.05; RTX 3090 24 GB (sm_86), PCIe 4.0 x16 (`ft bench bw`: H2D 23.8 GB/s), 32-core host, 503 GiB RAM; `dealignai/Qwen3.8-Flash-Next-ABLITERATED-NVFP4` @ `be794b99` (compressed-tensors NVFP4, same layout as the RadixArk stock checkpoint).

`ft serve --max-seq-len-override 262144 --kv-reserve-tokens 262144 --moe-backend offload --moe-cache-auto --nvfp4-backend triton --max-running-requests 1 --max-prefill-length 8192 --sampling-defaults model --reasoning-parser qwen3 --tool-call-parser qwen3_coder` (the serve as above but `--moe-backend hybrid`)

## Repro

1. `ft bench bw --dtype nvfp4 -o /workspace/bench-bw.json` → `nvfp4 … 2.62x hybrid; overlapped: CPU-MoE 43.4 + PCIe 21.6 GB/s -> hybrid fetches 33.2% of misses`, `saved: /workspace/bench-bw.json`
2. `ft serve … --moe-backend hybrid` → `WARNING --moe-hybrid-max-fetch auto: no usable ft bench bw profile for 'nvfp4' experts; using a fixed fetch cap of 1`
3. `FREETOKEN_BENCHBW_PATH=/workspace/bench-bw.json ft serve … --moe-backend hybrid` → `INFO --moe-hybrid-max-fetch auto: fetching 33.2% of each decode step's expert misses over PCIe … the rest on the CPU`

`moe/bench_profile.py` looks at `FREETOKEN_BENCHBW_PATH`, then `benchbw/<gpu_uuid>.json`, then the legacy `benchbw.json` — so `-o` opts the profile out of discovery, and nothing at `bw` time says so.

## Measured consequence

Short-prompt decode on this box: offload 24.5 tok/s, hybrid with the profile 31.7, hybrid with the silent fetch cap 28.3–30.5 — close enough that a user would never notice they are running the un-benchmarked configuration; at long context the cap-1 run was the unstable one (9.5 / 29.1 tok/s on two runs vs 12.4–13.1 with the profile).

## Ask (any of these; happy to PR whichever fits)

- `bw -o` prints the `export FREETOKEN_BENCHBW_PATH=…` line (or also writes the default location), or
- `--moe-backend hybrid` refuses to start without a usable profile unless `--moe-hybrid-max-fetch` is given explicitly, or
- a `--benchbw-path` flag on `ft serve` so the profile is part of the command line.

earlvanze:
Closes #308.

When `ft bench bw -o PATH` writes outside normal profile discovery, the report now prints the exact shell-safe `FREETOKEN_BENCHBW_PATH` export needed before `ft serve`. The `--out` help text makes the relationship explicit, and the hint is covered by a focused unit test.

Validation:
- `python3 -m py_compile` for changed modules
- `git diff --check`
- Focused pytest suite is not runnable in this checkout because pytest/runtime dependencies are not installed.

### DIFF
--- a/python/freetoken/moe/benchbw.py
+++ b/python/freetoken/moe/benchbw.py
@@ -39,6 +39,7 @@ import contextlib
 import json
 import logging
 import os
+import shlex
 import socket
 import statistics
 import threading
@@ -688,6 +689,11 @@ def _atomic_write_json(path: str, obj: dict) -> None:
     os.replace(tmp, path)
 
 
+def custom_profile_hint(path: str) -> str:
+    """Shell command required for a non-default bench profile to affect serving."""
+    return f"export FREETOKEN_BENCHBW_PATH={shlex.quote(path)}"
+
+
 def run_benchbw(
     out_path: str | None = None,
     threshold: float = 2.0,
@@ -703,6 +709,7 @@ def run_benchbw(
     kernel_cpu_iters: int = 64,
     kernel_pcie_iters: int = 20,
 ) -> dict:
+    custom_out_path = out_path is not None
     if not torch.cuda.is_available():
         raise RuntimeError(
             "benchbw needs a CUDA device to measure PCIe bandwidth (both offload and "
@@ -801,6 +808,7 @@ def run_benchbw(
     out_path = os.path.expanduser(out_path or default_out_path(gpu["uuid"]))
     _atomic_write_json(out_path, result)
     result["out_path"] = out_path
+    result["custom_out_path"] = custom_out_path
     if prog_on:
         print(f"FTBENCH_OUT {out_path}", flush=True)
     return result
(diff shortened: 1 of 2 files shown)


jason-fxz:
Thanks. LGTM.

One request: please drop `tests/moe/test_benchbw_cli.py`. It checks a one-line f-string and has to import torch (via `freetoken.moe.benchbw`) to do it, which is not worth a test file. With that removed this is good to merge.

PR #416, opened 2026-09-08
feat(qwen3_5_moe): keep a natively-fp8 lm_head native instead of failing to load

salekseev:
> [!NOTE]
> Stacked on #415 -- that's the first commit here. Review the second commit, or merge #415 first and this becomes a single commit.

A compressed-tensors mixed-precision checkpoint can put `lm_head` in its FP8 group (`unsloth/Qwen3.6-35B-A3B-NVFP4-Fast` does). Right now that combination can't load at all, and the reason is a mismatch between two places that each look locally correct.

The FP8 branch in `_iter_weights_attn_fp8` emits *any* FP8 `.weight` natively -- `lm_head` included. But the model always builds the bf16 `ParallelLMHead` for a head that isn't NVFP4, and that class has no `weight_scale` buffer. So:

```
Unexpected keys in state_dict: ['lm_head.weight_scale']
```

The easy fix would be to dequantize the head at load and move on. I went the other way and added the missing layer class, because nothing existing can consume a head that's already FP8 on disk -- `Nvfp4LMHead` is FP4, and the FP8 heads in `glm_moe_dsa`/`glm5_next` subclass the bf16 `ParallelLMHead` and quantize during load. `Fp8LMHead` is just `Fp8PerTensorLinear` plus `ParallelLMHead`'s prefill slice; the weight buffers, the `input_scale` handling and the uniform-scale/segment precompute all come from the parent, so the subclass is the slice and its docstring.

`_lm_head_quant` gains the matching `"fp8"` verdict, gated on the same weights geometry `_attn_quant` already uses, and the loader only keeps the head native when the model actually built an `Fp8LMHead` -- otherwise it falls through and dequantizes as before.

### Why this is more than a nice-to-have on small cards

On `[248320, 2048]` it's 0.474 GiB native vs 0.947 GiB dequantized. On a 16 GiB card that isn't just decode bandwidth: with the bf16 head there was 118 MiB free, and the GDN chunked-prefill workspace then OOM-killed the backend on an 8192-token prefill. Keeping it native is what makes that prefill fit.

### One thing I'd flag for review

`_lm_head_quant` also skips a head that the checkpoint lists in `ignore`. compressed-tensors gives `ignore` precedence over a group's `targets`, and the common llm-compressor shape is a broad target plus an `ignore` list carving modules back out -- such a head is bf16 on disk, so claiming it as FP8 would build an `Fp8LMHead` for a bf16 weight and die on the dtype check.

Worth knowing: the match there is a substring test, not `fnmatch`. `models/qwen4_exp/config.py` consults `ignore` via `fnmatch`, which suits ModelOpt's plain module names, but compressed-tensors writes regex entries like `re:.*lm_head` and `fnmatch("lm_head", "re:.*lm_head")` is `False`. So reusing that helper here would silently not match. Happy to change the approach if you'd rather it were shared.

### Reachability

Same caveat as #415: on `main` you can't get here, because the only checkpoints reaching `_iter_weights_attn_fp8` are ModelOpt ones and their `lm_head` is NVFP4, not FP8. It becomes reachable once compressed-tensors per-channel detection lands (#390 plus accepting `"channel"`).

### Tests

`tests/models/test_qwen3_5_moe_fp8_lm_head.py` -- detection for both FP8 strategies, plus the cases that must *not* be claimed: a head outside the FP8 group, a 4-bit group, a grouped/block scale, and an ignored head. Also asserts the layer's buffers match the per-output-row contract the loader feeds it.

### Testing done

`unsloth/Qwen3.6-35B-A3B-NVFP4-Fast` loads and serves on an RTX 4080 SUPER (16 GiB, sm_89, TP=1): 121.26 tok/s decode at 1200 tokens of context (mean of 4 runs, 120.05-123.3), 3311-3418 tok/s prefill at 8192 (8/8), 95.00 on BFCL parallel tool calls, 9/9 exact-match needle recall. `nvidia/Qwen3.6-35B-A3B-NVFP4` is unaffected -- its four quant verdicts are unchanged. Note that build also carries an unrelated FP8 KV-cache change, so don't read those throughput numbers as coming from a clean tree. More detail in #252.

### DIFF
--- a/python/freetoken/kernel/triton/fp8_pertensor_linear.py
+++ b/python/freetoken/kernel/triton/fp8_pertensor_linear.py
@@ -423,8 +423,45 @@ class Fp8PerTensorColMerged(Fp8PerTensorLinear):
         super().__init__(in_features, sum(output_sizes), has_bias)
 
 
+class Fp8LMHead(Fp8PerTensorLinear):
+    """FP8 (W8A16) LM head: ``Fp8PerTensorLinear`` with ``ParallelLMHead``'s prefill
+    behaviour at TP=1 -- slice to the last token per sequence, then the W8A16 GEMV/GEMM
+    instead of a bf16 ``F.linear`` over the dequantized weight. Weight buffers, the
+    ``input_scale`` dance and the uniform-scale/segment precompute are all inherited; the
+    prefill slice is the only new code.
+
+    Exists because a checkpoint can store ``lm_head`` as fp8 natively and no head class could
+    consume that: ``Nvfp4LMHead`` is FP4, and glm_moe_dsa/glm5_next's fp8 heads subclass the
+    bf16 ``ParallelLMHead`` and quantize at load. So a natively-fp8 head had to be
+    dequantized at load. For unsloth/Qwen3.6-35B-A3B-NVFP4-Fast that matrix is
+    ``[248320, 2048]``: 0.474 GiB kept native versus 0.947 GiB as bf16. On a 16 GiB card the
+    difference is not only decode traffic, it is the headroom an 8k prefill needs -- the
+    dequantized head left 118 MiB free and the GDN chunked-prefill workspace then OOM'd.
+
+    ``weight_scale`` is per output row ``[vocab]``, which covers both compressed-tensors
+    strategies: "channel" ships one scalar per row, "tensor" ships one for the whole matrix
+    and the loader broadcasts it. TP=1 and untied embeddings, same as ``Nvfp4LMHead``."""
+
+    def __init__(self, num_embeddings: int, embedding_dim: int):
+        self.num_embeddings = num_embeddings
+        self.embedding_dim = embedding_dim
+        super().__init__(in_features=embedding_dim, out_features=num_embeddings)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        from freetoken.core import get_global_ctx
+
+        batch = get_global_ctx().batch
+        if batch.is_prefill:
+            # Only the last position of each sequence produces logits, so slicing here keeps
+            # the [vocab, hidden] GEMM at M=batch rather than M=prompt_tokens.
+            indices = batch.attn_metadata.get_last_indices(batch.size)
+            x = x[indices].contiguous()
+        return super().forward(x)
+
+
 __all__ = [
     "FP8",
+    "Fp8LMHead",
     "Fp8PerTensorLinear",
     "Fp8PerTensorColMerged",
     "fp8_pertensor_linear",
--- a/python/freetoken/models/qwen3_5_moe/config.py
+++ b/python/freetoken/models/qwen3_5_moe/config.py
@@ -71,7 +71,9 @@ _compressed_tensors_nvfp4 = detect_compressed_tensors_nvfp4
 def _lm_head_quant(hf_config: Any) -> str:
     """Whether the checkpoint stores ``lm_head`` as NVFP4. modelopt MIXED_PRECISION lists it in
     the per-layer ``quantized_layers`` map (``W4A16_NVFP4``); pure-NVFP4 checkpoints have no
-    per-layer map and leave lm_head bf16. Returns ``"nvfp4"`` or ``"none"``."""
+    per-layer map and leave lm_head bf16. llm-compressor mixed-precision instead puts
+    ``lm_head`` in an fp8 ``config_groups`` entry. Returns ``"nvfp4"``, ``"fp8"``
+    or ``"none"``."""
     get = _quant_accessor(hf_config)
     if get is None:
         return "none"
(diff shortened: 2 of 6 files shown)


salekseev:
Some independent evidence for this one turned up while I was testing a different checkpoint for #390.

`primitive-ai/Ornith-1.5-35B-A3B-mixed-NVFP4-FP8` is a compressed-tensors mixed-precision export that puts `lm_head` in `ignore`, so its head stays **bf16** -- the exact situation this PR avoids. Same engine, same 3600 expert slots, same fp8 KV at 262,144 tokens, same box:

| | bf16 head (primitive-ai) | native fp8 head (unsloth) |
|---|---|---|
| decode @1200 | 106.95 | **121.26** (n=4, range 120.05-123.3) |
| free VRAM | 1.46 GiB | 1.91-2.13 GiB |
| prefill @8k | 3688 (8/8) | 3311-3418 (8/8) |

That's 14.3 tok/s, and the head accounts for most of it: 0.947 GiB read per decoded token versus 0.474. The only other difference between the two is the shared expert (fp8 vs NVFP4, worth roughly +52 MiB/token), so the head is about 90% of the ~525 MiB traffic delta. ΔTPOT is ~1.10 ms, which against ~525 MiB implies ~480 GB/s effective -- the right order for this card, so the traffic explanation at least hangs together.

Caveat it properly: two checkpoints differing in more than one variable, and primitive-ai is a single run. It is not a controlled experiment. But it's an independent measurement of the thing this PR is for, on a checkpoint I didn't pick for the purpose.

Also updating one number in the description: the decode figure there was a single 120.05 run. Four runs give a mean of 121.26. Same reason I corrected it on #390 -- the original could be read as the fp8 path costing decode, and it doesn't; against `nvidia/Qwen3.6-35B-A3B-NVFP4`'s 123.4 (n=4) the ranges overlap.

For what it's worth, I've promoted this checkpoint to my serving default on the strength of the tool-calling score and the checkpoint-calibrated KV scales, so the `Fp8LMHead` path is what I'm running daily rather than something I measured once.

salekseev:
Closing this -- the quantization refactor in #418/#427 replaced the mechanism this PR is built on, and did it better.

Every piece of scaffolding this hangs off is gone from `main`:

- `Fp8PerTensorLinear` was removed from `kernel/triton/fp8_pertensor_linear.py` (−81 lines), and `Nvfp4LMHead` from `nvfp4_linear.py` (−123). The layer classes now live under `layers/quantization/linear/`. So `Fp8LMHead` has no parent to subclass and no module to live in.
- `qwen3_5_moe/model.py` no longer branches on a `lm_head_quant` verdict. It's one unconditional `ParallelLMHead(..., quant_config=config.quant, prefix="lm_head")`, and the head resolves its own scheme through `quant_method`. The `elif` this PR adds has nowhere to go.

And the behaviour it was chasing now falls out of the new config layer on its own. Probing `CompressedTensorsConfig` against the checkpoint that motivated this:

```
unsloth/Qwen3.6-35B-A3B-NVFP4-Fast
  lm_head       -> QuantScheme(fp8_tensor, WeightDesc(elem='e4m3', group=(1, -1), ...))
  q_proj        -> QuantScheme(fp8_tensor, group=(1, -1))          # per-channel
  shared_expert -> QuantScheme(nvfp4, group=(1, 16))

primitive-ai/Ornith-1.5-35B-A3B-mixed-NVFP4-FP8
  lm_head       -> None                                            # correctly excluded via `ignore`
```

`group=(1, -1)` is the per-channel case, and `FP8_CHANNEL` is a first-class scheme now. That covers all three things I raised here: the natively-fp8 head, per-channel scales, and `ignore` taking precedence over a group's `targets`. The `ignore` handling in particular is more complete than what I wrote -- `ct_set(..., class_names=False)` understands class-name targets like `"Linear"`, which is the thing I'd flagged on #252 as the real blocker for `ornith-ai/Ornith-1.5-35B-A3B-FP8`.

Nice work; this is a better design than bolting another head class on beside `Nvfp4LMHead`.

**One thing that is still open, and it's the reason I'm not just closing this quietly.** `qwen3_5_moe` isn't finished migrating. Its `config.py` still computes `_expert_quant` / `_attn_quant` / `_lm_head_quant` (lines 42/71/107, consumed at 169/172/180), and `_iter_weights_attn_fp8` still routes off those strings. On today's `main` the legacy detectors return `attn='none'` for both compressed-tensors checkpoints above, so the reader never reaches the fp8 dense path even though the config layer describes it correctly. `qwen4_exp` is fully migrated with zero legacy detector references and looks like the template.

I'd rather spend the effort there than rebase this. I have `unsloth/Qwen3.6-35B-A3B-NVFP4-Fast`, `nvidia/Qwen3.6-35B-A3B-NVFP4` and the Ornith mixed export on disk with a 4080 SUPER and load/needle/BFCL harnesses wired up, so I can validate a migration against all three including the modelopt regression. Happy to open that as an issue or a PR if it's wanted -- say which and I'll get on with it.

#415 I'll leave open for now, since it fixes a live bug in the legacy `_per_row_scale` that the migration would eventually delete; there's a note on it.

Issue #437, opened 2026-09-10
[Feature] Migrate qwen3_5_moe onto layers/quantization: the new config layer resolves compressed-tensors checkpoints correctly, the legacy detectors don't (offering a PR)


salekseev:
### Applies to

`qwen3_5_moe` (engine), following the quantization refactor in #418 / #427. Verified against `main` @ `fb7f732`.

### What problem does this solve

`layers/quantization/` already describes the compressed-tensors checkpoints people keep filing issues about -- correctly -- but `qwen3_5_moe` doesn't consume it, so those checkpoints still don't load.

Probing `QuantConfig.from_hf(...).scheme_for_name(...)` on today's `main`:

```
unsloth/Qwen3.6-35B-A3B-NVFP4-Fast
  self_attn.q_proj            -> fp8_tensor, group=(1, -1)     # per-channel
  lm_head                     -> fp8_tensor, group=(1, -1)
  mlp.shared_expert.gate_proj -> nvfp4,      group=(1, 16)

primitive-ai/Ornith-1.5-35B-A3B-mixed-NVFP4-FP8
  self_attn.q_proj            -> fp8_tensor, group=(-1, -1)    # per-tensor
  lm_head                     -> None                          # correctly excluded via `ignore`

nvidia/Qwen3.6-35B-A3B-NVFP4   (modelopt)
  self_attn.q_proj / experts / shared_expert / lm_head -> fp8_tensor / nvfp4 / nvfp4 / nvfp4
```

All three resolve exactly right. But `qwen3_5_moe/config.py` still computes its own verdicts with hand-rolled detectors -- `_expert_quant` (L42), `_lm_head_quant` (L71), `_attn_quant` (L107), consumed at L169/172/180 -- and `_iter_weights_attn_fp8` in `weight.py` still routes off those strings. Those detectors return `attn_quant='none'` for **both** compressed-tensors checkpoints above, so the reader never reaches the fp8 dense path even though the config layer describes it perfectly.

`model.py` is already migrated for the head -- it passes `quant_config=config.quant, prefix="lm_head"` to a single `ParallelLMHead` -- so the model half and the config half currently disagree about the same checkpoint.

This is the shared root cause under a cluster of open reports: #252 (per-channel FP8, `targets: ["Linear"]`), #381 (`model.language_model` prefix), #344, #410, and the routing discussion on #390 / #413.

### Proposed solution

[...]

Issue #410, opened 2026-09-07
[Model request] KAT-Coder-V2.5-Dev NVFP4 community exports (compressed-tensors MoE): dense-branch misroute + GDN granularity


minatoAI:
### Before you start

- [x] I have read the [FAQ](https://github.com/FlashML-org/FreeToken/issues/84) and my problem is not answered there.
- [x] I have read the [Roadmap](https://github.com/FlashML-org/FreeToken/issues/79) and this is not already planned there.
- [x] I have searched [existing issues](https://github.com/FlashML-org/FreeToken/issues?q=is%3Aissue) and found no duplicate.
- [x] I am on the latest release, or on a freshly rebuilt `main` when building from source.

## Checkpoints
- `sakamakismile/KAT-Coder-V2.5-Dev-NVFP4` — uniform NVFP4 (llm-compressor recipe
  targets `Linear`, bf16 lm_head, single `model.safetensors`)
- `sahilchachra/KAT-Coder-V2.5-Dev-NVFP4A16` — NVFP4A16 (recipe `ignore` carries
  `re:.*\.linear_attn\..*`, so all GDN projections stay bf16)

Both are Qwen3.5-MoE (`Qwen3_5MoeForConditionalGeneration`, hidden 2048, 40
layers, 256 experts / top-8, MoE-512), `quant_method: compressed-tensors`,
`format: nvfp4-pack-quantized`.

## Environment
- GPU: RTX 5070Ti 16GB (Blackwell) / driver 610 / Windows 11
- FreeToken Desktop 0.1.2+g141c31a8d (Windows bundle)
- Commands: `ft checkpoint` (via Desktop) then
  `ft serve --model <id> --moe-backend auto`

## Problem 1: MoE+compressed-tensors falls into the dense loader
`_expert_quant()` only recognizes ModelOpt `NVFP4` / `MIXED_PRECISION`, so
`expert_quant` falls back to `none` and `iter_weights()` takes the
dense (`no routed experts`) branch. Same bug class as #381.

## Problem 2: one attn knob cannot cover both GDN layouts
After routing experts to the NVFP4 banks, serve fails per-variant:
- A16 (`sahilchachra`): `KeyError: 'model.layers.0.linear_attn.out_proj.weight_scale'`
  (`nvfp4_linear.py load_state_dict` pops a scale the bf16 GDN never had).
- uniform (`sakamakismile`): `AssertionError` at
[...]

PR #438, opened 2026-09-10
feat(qwen3_5_moe): read every checkpoint layout through the QuantConfig

### LINKED ISSUE #238: [BUG] qwen3_5 dense: mixed-precision NVFP4 crashes in ct_bf16_fuse (Float8 × BFloat16 promotion)

### LINKED ISSUE #344: [Windows beta] Compressed-tensors NVFP4 MoE checkpoints: per-expert tensors leak into the dense state_dict
### Environment
- Windows 11 (10.0.26200), RTX 5060 Laptop 8GB, driver 596.49 (CUDA 13.2)
- freetoken 0.1.2+g932e87e8a (cp312-win_amd64 wheel from the `FreeToken-Web` `beta` tag) + freetoken_kernel_cache-0.1.2+cu130
- torch 2.11.0+cu130, triton-windows 3.7.1.post27, installed via `engine/install.ps1` (desktop bundle)

### Model
`sakamakismile/Huihui-Qwen3.6-35B-A3B-abliterated-NVFP4` — a compressed-tensors `nvfp4-pack-quantized` **MoE** checkpoint with per-expert unfused keys:

```
model.language_model.layers.N.mlp.experts.E.{gate,up,down}_proj.weight_packed   # uint8 [O, K//2]
model.language_model.layers.N.mlp.experts.E.{...}.weight_scale                  # fp8-e4m3 [O, K//16]
model.language_model.layers.N.mlp.experts.E.{...}.weight_global_scale           # f32 scalar (quant-side, reciprocal)
model.language_model.layers.N.mlp.experts.E.{...}.input_global_scale            # f32 scalar
```

### What works
After a byte-level safetensors transform I wrote (fuse shared-expert gate/up into `gate_up_proj` keeping per-part global scales; dequantize GDN `in_proj_{a,b,qkv,z}` to bf16 via `E2M1[code] * block_scale / weight_global_scale`; keep everything else as-is), the dense pass of `_iter_weights_compressed_tensors` loads 100% correctly — shared_expert fusion, GDN `in_proj` bf16 assembly, attention projections, lm_head, norms all consume cleanly.

### What fails
The routed experts never reach the offload bank. The dense pass yields per-expert packed tensors as standalone NVFP4 projections, so they land in the dense state_dict and the MoE module's strict check rejects them:

```
RuntimeError: Unexpected keys in state_dict: ['model.layers.0.mlp.experts.0.down_proj.weight', ...]
```

Renaming the per-expert keys to modelopt kinds (`weight` / `weight_scale_2` / `input_scale`) does not help either — the ct iter has no expert skip in either naming, and the bank builder then reports:

```
ValueError: Missing MoE expert source layers: {'gate_up': [0..39], 'down': [0..39]}
```

because it probes for per-layer **stacked** `...experts.gate_up_proj` / `...experts.down_proj` keys, which this checkpoint does not use.

### Ask
Main appears to have the pieces for per-expert modelopt-kind expert sources (`_NVFP4_EXPERT_KEY_RE`, `_NVFP4_SOURCE_SPEC` in `nvfp4_banks`), but the published Windows beta wheel predates that. Could you either:
1. ship an updated Windows beta wheel with MoE + compressed-tensors per-expert support (so `_iter_weights_compressed_tensors` routes `experts.<int>.*` to the offload bank instead of the dense state_dict), or
2. document the expected expert key layout for compressed-tensors MoE checkpoints on Windows?

Dense compressed-tensors checkpoints (Qwen3.6-27B class) load fine with the same wheel — the gap is MoE-specific. Happy to test any preview wheel.

### LINKED ISSUE #381: [Bug]: qwen3_5_moe fails to convert/load compressed-tensors NVFP4 checkpoints with model.language_model prefix
### Description
When attempting to convert or load a `compressed-tensors` (llm-compressor) NVFP4 export of a Qwen3.5 / Agents-A1 MoE model (e.g. `Agents-A1-NVFP4` with `model.language_model.*` tensor hierarchy), `ft checkpoint` fails during expert bank conversion.

Depending on prefix handling, the conversion halts with:
1. `ValueError: Missing MoE expert source layers: {'gate_up': [0..39], 'down': [0..39]}` (when `_NVFP4_EXPERT_KEY_RE` does not match the `compressed-tensors` naming scheme)
2. `KeyError: (0, 0, 'down_proj')` or assertion failure in `_iter_weights_compressed_tensors` when encountering mixed precision (BF16 linear attention + NVFP4 full attention/experts).

### Root Cause Analysis

1. **Quant Detection (`freetoken/models/qwen3_5_moe/config.py`):**
   `_expert_quant()` checks for `quant_algo == "NVFP4"` (ModelOpt style) or `quant_algo == "MIXED_PRECISION"`. For checkpoints produced by `llm-compressor`, the top-level field is `quant_method: "compressed-tensors"`, which results in `expert_quant` falling back to `"none"`.

2. **Regex & Suffix Mismatch (`freetoken/models/qwen3_5_moe/weight.pyd`):**
   `_NVFP4_EXPERT_KEY_RE` expects keys matching `(?P<kind>weight|weight_scale|weight_scale_2)`. In `compressed-tensors` checkpoints:
   - Packed weights are named `*.weight_packed` instead of `*.weight`.
   - Global scale scalars are named `*.weight_global_scale` / `*.input_global_scale`.
   - The tensor tree carries a `model.language_model.layers...` prefix rather than `model.layers...`.

3. **Hybrid Attention/GDN Quantization:**
   In `compressed-tensors` MoE checkpoints for hybrid architectures (e.g. Qwen3.5 with GatedDeltaNet + full attention), only the full attention layers and routed experts are quantized to NVFP4, while the linear recurrent (GDN) projections (`conv1d`, `A_log`, `in_proj_*`, `out_proj`) remain in BF16. The compiled loader assumes all dense projections in a `compressed-tensors` checkpoint share a uniform quantization scheme.

### Steps to Reproduce
1. Obtain a `compressed-tensors` NVFP4 checkpoint of `Agents-A1` or `Qwen3.5-MoE`.
2. Run:
   ```bash
   ft checkpoint --model <checkpoint_dir> --out <ftw_dir> --dtype bfloat16 --moe-backend offload
   ```
3. Observe `ValueError: Missing MoE expert source layers` or failure during weight iteration.

### Environment
- OS: Windows 11
- FreeToken Version: 0.1.2+g0ee42b72c
- GPU: NVIDIA GeForce RTX 4060

jason-fxz:
## What

`qwen3_5_moe` picked one of four weight readers by string flags (`attn_quant`, `dense_quant`, `lm_head_quant`) that its config derived from the HF `quantization_config` on its own. Since #418 every layer builds its buffers from `config.quant.scheme_for(prefix)`, so the flags and the readers disagreed with the model on every compressed-tensors export (#238, #263, #344, #381, #410, #437, and the PRs #208, #275, #296, #390, #413, #415).

One reader now asks the same `scheme_for` for every Linear module and fills exactly the buffers the model declared:

- `qwen3_5_moe/weight.py`: one `_DenseReader` replaces the four readers and their fusion tables. Tensors are matched to the roles of the module's scheme, packed projections are concatenated per role, dtypes and scale shapes are checked against the scheme, per-tensor scales are broadcast per row (the shape handling from #415). A module the config calls bf16 but whose tensors are quantized is rejected; a module the family serves bf16 (the routers) is dequantized.
- `qwen3_5_moe/config.py`: the string detectors are gone; `expert_quant` comes from the `QuantConfig`, as in `qwen4_exp`.
- Each dialect declares `STORAGE`, the checkpoint tensor behind every role of every kind it exports (`weight_packed` / `weight_global_scale` for llm-compressor, `weight` / `weight_scale_2` for ModelOpt), with the reciprocal convention on the entry. The NVFP4 expert reader spec is derived from it.
- `ct_set`: a compressed-tensors `targets` / `ignore` name covers that module alone, not its children, as in llm-compressor and vLLM. Exports list every unquantized module including containers (`layers.N.linear_attn`, `mlp.experts.N`), and the old ancestor match turned the quantized projections under them into bf16. Target class names other than `Linear` fail closed.
- `Nvfp4LinearMethod` declares `input_scale` when the scheme carries it, so W4A4 checkpoints load in full; no kernel reads it yet. The gemma4, glm4_moe and muse_glimmer readers load it too.
- `loader.safetensors_weight_map` reads the index or, without one, the shard headers; it replaces the hard index read in `nvfp4_banks.py` that failed on single-file checkpoints.
- `glm5_next` keeps its own copy of `_fp8_block_quant` instead of importing it from `qwen3_5_moe`.

## Compatibility

An FTW written by an older build from an NVFP4 dense checkpoint (for example nvidia/Qwen3.6-27B-NVFP4, nvidia/Qwen3.6-35B-A3B-NVFP4, RadixArk/Qwen3.8-27B-NVFP4, nvidia/Gemma-4-31B-IT-NVFP4, nvidia/GLM-4.7-NVFP4) lacks the `input_scale` tensors the model now declares and fails with a `KeyError`. `scripts/ftw_hotfix.py --ftw <dir> --source <hf dir>` adds them.

## Tested

H100 80 GB, driver 580.95, CUDA 13.0, torch 2.11, on top of `main` @ fb7f732. `tests/e2e/test_aime.py` (AIME25 problem 1, pass@3) with `FREETOKEN_TEST_MODEL=<dir>`, plus `FREETOKEN_TEST_MOE_CACHE_AUTO=1` for MoE.

| checkpoint | layout | result |
|---|---|---|
| sakamakismile/Qwen3.6-27B-NVFP4 | compressed-tensors NVFP4, GDN quantized | pass (`main`: `KeyError in_proj_qkvz`) |
| unsloth/Qwen3.6-35B-A3B-NVFP4-Fast | compressed-tensors channel-fp8 + NVFP4, fp8 lm_head | pass |
| RedHatAI/Qwen3.6-35B-A3B-NVFP4 | compressed-tensors NVFP4 | pass |
| primitive-ai/Ornith-1.5-35B-A3B-mixed-NVFP4-FP8 | compressed-tensors static fp8 + NVFP4 | pass |
| kyaky/Qwen3.6-35B-A3B-Uncensored-NVFP4 | compressed-tensors block-fp8 + NVFP4, single file | pass |
| unsloth/Qwen3.8-27B-NVFP4 | compressed-tensors channel-fp8 + NVFP4, dense | pass |
| nvidia/Qwen3.6-35B-A3B-NVFP4 | ModelOpt FP8 + NVFP4 | pass |
| nvidia/Qwen3.6-27B-NVFP4 | ModelOpt FP8 + NVFP4, dense | pass |
| RadixArk/Qwen3.8-27B-NVFP4 | ModelOpt FP8 + NVFP4, dense | pass |
| Qwen/Qwen3.6-35B-A3B-FP8 | block-fp8 | pass |
| Qwen/Qwen3.8-27B-FP8 | block-fp8, dense | pass |
| Qwen/Qwen3.6-27B | bf16, dense | pass |
| Qwen/Qwen3.5-2B | bf16, dense, tied embeddings | pass |
| Qwen/Qwen3.6-35B-A3B | bf16 | pass |
| nvidia/Gemma-4-31B-IT-NVFP4 | ModelOpt NVFP4 dense MLP (gemma4 reader) | pass |
| nvidia/Qwen3.6-27B-NVFP4 FTW from an older build | | `KeyError ... input_scale` as expected; pass after `ftw_hotfix.py --source` |
| RedHatAI/Muse-Glimmer-30B-NVFP4 | compressed-tensors NVFP4 (muse_glimmer reader) | loads; no answer within the 16k-token cap, same as `main` |
| nvidia/GLM-4.7-NVFP4 | ModelOpt NVFP4 shared experts (glm4_moe reader) | loads (dense pass); AIME not run, 215 GB of experts do not fit this host's RAM |

`pytest tests/models tests/moe/test_offload.py tests/engine/test_cache_budget.py`: 255 passed.

Closes #238, closes #344, closes #381. Implements the migration proposed in #437, left open for the int4 discussion. Addresses the compressed-tensors NVFP4 side of #263 (sakamakismile/Huihui-Qwen3.6-35B-A3B-abliterated-NVFP4; the joshebbs export with the nested ModelOpt config is not covered) and of #410 (the sahilchachra, sakamakismile and Ttimms KAT-Coder exports have regular configs; doth4580's format-only config is not covered). Not #252: its ornith-ai/Ornith-1.5-35B-A3B-FP8 quantizes the routed experts per-channel fp8 too, and there is no fp8 per-tensor MoE method yet. Supersedes #208, #275, #296, #390, #415 (its per-channel scale handling is included) and the reader and detection parts of #413 (its KAT-Coder handling is not included).

### DIFF
--- a/python/freetoken/layers/quantization/configs/base.py
+++ b/python/freetoken/layers/quantization/configs/base.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+from dataclasses import dataclass
 from typing import Any, ClassVar
 
 from ..linear import LinearConfig
@@ -36,8 +37,18 @@ def quantization_config_of(hf_config: Any) -> dict[str, Any] | None:
     return dict(vars(q))
 
 
+@dataclass(frozen=True)
+class Stored:
+    """One checkpoint tensor behind a role: its suffix, and whether it holds the quant-side scale whose reciprocal the layer wants."""
+
+    name: str
+    reciprocal: bool = False
+
+
 class QuantConfig(ABC):
     dialect: ClassVar[str]
+    # per kind the dialect exports, role -> the checkpoint tensor suffix (or Stored) that feeds it
+    STORAGE: ClassVar[dict[QuantKind, dict[str, str | Stored]]]
 
     def __init__(self, name_map: NameMap | None = None, unquantized: tuple[str, ...] = ()):
         self.name_map = name_map or NameMap()
@@ -64,6 +75,14 @@ class QuantConfig(ABC):
         scheme = schemes.pop()
         self._schemes[prefix] = scheme
         return scheme
+
+    def stored_tensors(self, kind: QuantKind) -> dict[str, Stored]:
+        """role -> checkpoint tensor for every role the dialect stores for ``kind``."""
+        return {role: entry if isinstance(entry, Stored) else Stored(entry) for role, entry in self.STORAGE[kind].items()}
+
+    def storage(self, scheme: QuantScheme) -> dict[str, Stored]:
+        """role -> checkpoint tensor for one scheme's tensors."""
+        return {role: entry for role, entry in self.stored_tensors(scheme.kind).items() if scheme.has(role)}
 
     def get_quant_method(self, layer: Any, prefix: str):
         scheme = self.scheme_for(prefix)
@@ -101,6 +120,7 @@ class NoQuantConfig(QuantConfig):
     """No ``quantization_config``: every module is bf16."""
 
     dialect = "none"
+    STORAGE: ClassVar[dict[QuantKind, dict[str, str | Stored]]] = {}
 
     @classmethod
     def claims(cls, q: dict[str, Any]) -> bool:
--- a/python/freetoken/layers/quantization/configs/compressed_tensors.py
+++ b/python/freetoken/layers/quantization/configs/compressed_tensors.py
@@ -4,9 +4,9 @@ from typing import Any, ClassVar
 
 from ..names import Matcher, ct_set
 from ..registry import register_dialect
-from ..scheme import QuantScheme
+from ..scheme import QuantKind, QuantScheme
 from ..scheme import fp8_block_scheme, fp8_tensor_scheme, nvfp4_scheme
-from .base import QuantConfig
+from .base import QuantConfig, Stored
 
 
 @register_dialect
@@ -22,6 +22,15 @@ class CompressedTensorsConfig(QuantConfig):
         "FP8_TENSOR_DYNAMIC": fp8_tensor_scheme("bf16"),
         "FP8_CHANNEL": fp8_tensor_scheme("bf16", per_row=True),
         "FP8_BLOCK": fp8_block_scheme("float"),
+    }
+    # the NVFP4 globals are the quant-side scales (vLLM: alpha = 1 / (input_global_scale * weight_global_scale))
+    STORAGE: ClassVar[dict[QuantKind, dict[str, str | Stored]]] = {
+        QuantKind.NVFP4: {
+            "weight": "weight_packed", "weight_scale": "weight_scale",
+            "weight_global": Stored("weight_global_scale", reciprocal=True), "input_scale": Stored("input_global_scale", reciprocal=True),
+        },
+        QuantKind.FP8_TENSOR: {"weight": "weight", "weight_scale": "weight_scale", "input_scale": "input_scale"},
+        QuantKind.FP8_BLOCK: {"weight": "weight", "weight_scale_inv": "weight_scale"},
     }
 
     def __init__(self, q: dict[str, Any], hf_config: Any = None, *, name_map=None, unquantized=()):
--- a/python/freetoken/layers/quantization/configs/fp8.py
+++ b/python/freetoken/layers/quantization/configs/fp8.py
@@ -4,9 +4,9 @@ from typing import Any, ClassVar
 
 from ..names import is_routed_expert, name_set, substr_set
 from ..registry import register_dialect
-from ..scheme import QuantScheme
+from ..scheme import QuantKind, QuantScheme
 from ..scheme import FP8_BLOCK, fp8_block_scheme, fp8_tensor_scheme, mxfp4_scheme
-from .base import QuantConfig, cfg_get
+from .base import QuantConfig, Stored, cfg_get
 
 
 @register_dialect
@@ -22,6 +22,12 @@ class Fp8BlockConfig(QuantConfig):
         "TABLE": fp8_tensor_scheme("float"),
         "EXPERT_MXFP4": mxfp4_scheme(),
     }
+    # transformers' fp8 names; DeepSeek-V4's e8m0 export calls every scale ``scale`` (see storage)
+    STORAGE: ClassVar[dict[QuantKind, dict[str, str | Stored]]] = {
+        QuantKind.FP8_BLOCK: {"weight": "weight", "weight_scale_inv": "weight_scale_inv"},
+        QuantKind.FP8_TENSOR: {"weight": "weight", "weight_scale": "weight_scale"},
+        QuantKind.MXFP4: {"weight": "weight", "weight_scale": "scale"},
+    }
 
     def __init__(self, q: dict[str, Any], hf_config: Any = None, *, name_map=None, unquantized=()):
         super().__init__(name_map, unquantized)
@@ -35,6 +41,12 @@ class Fp8BlockConfig(QuantConfig):
         self.convert_tables = name_set(tuple(q.get("modules_to_convert") or ()))
         self.e8m0 = str(q.get("scale_fmt") or "").lower() == "ue8m0"
         self.expert_fp4 = str(cfg_get(hf_config, "expert_dtype") or "").lower() == "fp4"
+
+    def storage(self, scheme: QuantScheme) -> dict[str, Stored]:
+        names = super().storage(scheme)
+        if self.e8m0 and scheme.kind is QuantKind.FP8_BLOCK:
+            names["weight_scale_inv"] = Stored("scale")
+        return names
 
     def scheme_for_name(self, name: str) -> QuantScheme | None:
         if self.convert_tables(name):
--- a/python/freetoken/layers/quantization/configs/modelopt.py
+++ b/python/freetoken/layers/quantization/configs/modelopt.py
@@ -4,9 +4,9 @@ from typing import Any, ClassVar
 
 from ..names import ancestors, name_set
 from ..registry import register_dialect
-from ..scheme import QuantScheme
+from ..scheme import QuantKind, QuantScheme
 from ..scheme import fp8_block_scheme, fp8_tensor_scheme, mxfp8_scheme, nvfp4_scheme
-from .base import QuantConfig
+from .base import QuantConfig, Stored
 
 
 @register_dialect
@@ -23,6 +23,12 @@ class ModelOptConfig(QuantConfig):
         "FP8_PER_CHANNEL_PER_TOKEN": fp8_tensor_scheme("fp32", per_row=True),
         "FP8_PB_WO": fp8_block_scheme("fp32"),
         "MXFP8": mxfp8_scheme(),
+    }
+    STORAGE: ClassVar[dict[QuantKind, dict[str, str | Stored]]] = {
+        QuantKind.FP8_TENSOR: {"weight": "weight", "weight_scale": "weight_scale", "input_scale": "input_scale"},
+        QuantKind.FP8_BLOCK: {"weight": "weight", "weight_scale_inv": "weight_scale_inv"},
+        QuantKind.MXFP8: {"weight": "weight", "weight_scale_inv": "weight_scale_inv"},
+        QuantKind.NVFP4: {"weight": "weight", "weight_scale": "weight_scale", "weight_global": "weight_scale_2", "input_scale": "input_scale"},
     }
 
     @classmethod
--- a/python/freetoken/layers/quantization/configs/mxfp4.py
+++ b/python/freetoken/layers/quantization/configs/mxfp4.py
@@ -4,9 +4,9 @@ from typing import Any, ClassVar
 
 from ..names import name_set
 from ..registry import register_dialect
-from ..scheme import QuantScheme
+from ..scheme import QuantKind, QuantScheme
 from ..scheme import mxfp4_scheme
-from .base import QuantConfig
+from .base import QuantConfig, Stored
 
 
 @register_dialect
@@ -16,6 +16,8 @@ class Mxfp4Config(QuantConfig):
     dialect = "mxfp4"
 
     SCHEME: ClassVar[QuantScheme] = mxfp4_scheme()
+    # the experts are stacked per layer (``gate_up_proj_blocks`` / ``_scales``), not per-Linear tensors; gpt_oss reads them itself
+    STORAGE: ClassVar[dict[QuantKind, dict[str, str | Stored]]] = {}
 
     def __init__(self, q: dict[str, Any], hf_config: Any = None, *, name_map=None, unquantized=()):
         super().__init__(name_map, unquantized)
--- a/python/freetoken/layers/quantization/linear/nvfp4.py
+++ b/python/freetoken/layers/quantization/linear/nvfp4.py
@@ -120,4 +120,5 @@ class Nvfp4LinearMethod(LinearMethod):
         layer.weight = torch.empty(g.out_features, g.in_features // 2, dtype=torch.uint8)
         layer.weight_scale = torch.empty(g.out_features, g.in_features // GROUP, dtype=FP8)
         layer.weight_global = torch.empty(g.out_features, dtype=torch.float16)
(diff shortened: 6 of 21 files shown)


salekseev:
Ran #438 on sm_89 with MoE offload, since the table in the PR is H100. It works out of the box, `953659d` over the `gfb7f732de` nightly wheels.

RTX 4080 SUPER, 16 GiB, sm_89, TP=1, `--moe-backend offload`, 3600 expert slots, 62 GB host RAM.

| checkpoint | dialect | result |
|---|---|---|
| `unsloth/Qwen3.6-35B-A3B-NVFP4-Fast` | CT, fp8 **per-channel** dense + NVFP4 | healthy in 40 s; decode **123.46 tok/s**; prefill @8k **3648 tok/s, 8/8**; coherent; 2/2 parallel tool calls |
| `nvidia/Qwen3.6-35B-A3B-NVFP4` | ModelOpt MIXED_PRECISION | healthy in 50 s, 2.07 GiB free, generates correctly -- **no regression** |
| `primitive-ai/Ornith-1.5-35B-A3B-mixed-NVFP4-FP8` | CT, fp8 **per-tensor** dense + NVFP4 | healthy in 50 s, 1.50 GiB free, generates correctly |

The 8192-token prefill is the canary I care about on a 16 GiB card, because that is where the GDN chunked-prefill workspace has OOM'd before (#401). 8/8 at 3648 tok/s is the best I have recorded for this checkpoint.

This is now my serving default rather than a one-off test, so it has continuous use on sm_89. At the model's full 262,144-token context it measures 122.4 tok/s decode and 3398 tok/s prefill at 8k, 8/8, which is at or slightly above what I was getting before this branch.

I confirmed your point 1 directly rather than taking it on trust. `checkpoint_quant_config` on this branch:

```
Avesed/Qwen3.6-35B-A3B-INT4-W4A16      compressed-tensors -> NotImplementedError: weight scheme {...}
cyankiwi/Qwen3.6-35B-A3B-AWQ-4bit      awq                -> NotImplementedError: 'awq' is not supported
cyankiwi/Qwen-AgentWorld-35B-A3B-...   compressed-tensors -> NotImplementedError: weight scheme {...}
nvidia / unsloth / primitive-ai        modelopt / CT      -> ModelOptConfig / CompressedTensorsConfig
```

You are right that the legacy detectors were never protecting those three. I had proposed keeping them as a fallback and that was wrong: they only ever loaded for me because of local int4 patches, not because of anything on `main`. My "never downgrade a verdict" rule was guarding something that does not exist upstream. I have retired those checkpoints rather than keeping a fallback alive, and I am treating #396 as where that work belongs, with the scope you laid out. The repack step in particular I had not thought about, and it is the part that would have bitten me.

Two findings from prototyping the same migration before you posted, in case either is useful.

The `ignore` ancestor-semantics fix is the one that matters most, and it is worth a regression test if there isn't one already. It is not a corner case: unsloth lists `...layers.N.linear_attn` beside the genuinely-bf16 `in_proj_b`/`in_proj_a`/`norm`, and the Ornith mixed export lists all 10,240 `...experts.M` parents. Under ancestor semantics both read as bf16, and on a 16 GiB card that surfaces as `KeyError: '...linear_attn.in_proj.weight'` -- the GDN builds a fused `in_proj` because `scheme_for('...in_proj_qkvz')` returned `None`, while the reader emits the split names. It took me a while to connect those two.

Separately, the `layers/quantization` matchers built from closures cannot be pickled: `AttributeError: Can't pickle local object 'name_set.<locals>.hit'`. It does not bite today because the worker rebuilds the config rather than receiving it, but if a `QuantConfig` is ever placed on a config object that crosses into the scheduler worker it will, and the failure mode is silent-unquantized rather than a pickling error. Making those matchers classes with `__call__` would pre-empt it cheaply.

One more datapoint on this as a base to build on: I carry a per-tensor fp8 KV change of my own, and rebasing it onto this branch needed hand-work on only two hunks, both in `_adjust_config` and both from the `moe_backend` -> `moe_strategy` rename shifting context. Nothing interacted with the scheme reader.

#415 is closed, since this carries it. Thanks for crediting the `_per_row_scale` logic. Happy to keep reporting sm_89 numbers as this evolves, and to run needle-recall or BFCL on it if that would be useful for the PR table.

salekseev:
Follow-up: I've moved my serving default onto this branch, so it now has continuous use on sm_89 rather than just a test run.

Config: `953659d` over the `gfb7f732de` wheels, plus my own per-tensor fp8 KV carry, serving `unsloth/Qwen3.6-35B-A3B-NVFP4-Fast` at the model's full 262,144-token context with 3600 expert slots on a 16 GiB RTX 4080 SUPER.

| | previous (carried patches on `af71ba432`) | now (#438) |
|---|---|---|
| patches carried | 10 | **4** |
| decode @1200 | 121.26 tok/s (n=4) | **122.4** |
| prefill @8k | 3311–3418, 8/8 | 3398, 8/8 |
| KV | fp8, 262,144 tokens | same, **10/10 checkpoint-calibrated scales** |
| free VRAM | 1.91 GiB | 1.47 GiB |

Six patches went away: the three int4/AWQ expert patches (with those checkpoints — see below), `pr275-ct-fp8-crash` (superseded, you rewrote the reader it patched so the `ct_bf16_fuse` promotion can't be reached), and #390 plus my own `ct-fp8-channel`, both superseded by resolving layouts through the `QuantConfig`.

**#415 is closed.** Thanks for crediting the `_per_row_scale` logic.

Two things that may be useful to you:

The fp8-KV carry rebased onto this branch with 3 of 5 `engine.py` hunks landing unchanged and 9 of 10 files clean. The two that needed hand-work were both in `_adjust_config`, from the `moe_backend` → `moe_strategy` rename shifting context — no interaction with the scheme reader at all. The one hunk I dropped was comment-only. So from the outside this is a well-behaved base to carry work on top of.

And I confirmed the int4 side end-to-end rather than just at the config layer: `checkpoint_quant_config` on this branch raises for all three of my int4/AWQ checkpoints (`'awq' is not supported`, and `weight scheme {...}` twice), which matches exactly what you said. I've retired those checkpoints from my stack rather than trying to keep a fallback alive, and I'm treating #396 as the place that work belongs — with the scope you laid out, particularly the repack step, since AutoAWQ and compressed-tensors pack int4 differently.

Happy to keep reporting sm_89 numbers as this evolves, and to run needle/BFCL on it if that's useful for the PR table.

PR #439, opened 2026-09-10
fix(scheduler): read the prefill chunk cap from the pool, not a snapshot

gberasmus87:
fix(scheduler): read the prefill chunk cap from the pool, not a snapshot

`CacheManager.__init__` copied `swa_pool.prefill_chunk_budget` into an instance
attribute, so the cap kept its construction-time value for the life of the
manager and `rebuild` never refreshed it.

`Scheduler.rebuild_cache` recomputes `prefill_budget` from
`cache_manager.prefill_chunk_budget` after a runtime resize, so reading a frozen
value meant it wrote back the number it already had. Two consequences:

* Growing the pool leaves prefill chunking where it was. Measured on
  DSV4-Flash through `POST /v1/cache/rebuild`: the window pool went 100 -> 215
  pages and the chunk budget stayed at 4864, so an 11.7k prompt still took three
  whole-layer expert streams (32.5 s) instead of the one it had just been sized
  for.
* Shrinking is worse, and is the hazard `rebuild_cache`'s own comment warns
  about: the stale cap is then too LARGE for the pool, so the next long prompt is
  chunked past what `_alloc_window` can satisfy.

Making it a property that re-reads the pool on every access fixes both. Nothing
assigns to it -- `scheduler.py` reads it in two places and `DSV4PagedKVCache`
owns the value -- so a read-only property is the whole change.

`test_rebuild_cache_refreshes_prefill_budget` already covers the scheduler half,
but with a `SimpleNamespace` manager whose cap the test sets by hand, so it
cannot see this: the gap is between a real manager and its real pool. The new
test closes exactly that gap. It drives `DSV4PagedKVCache.rebuild` (whose
`_init_paged_state` recomputes `_chunk_budget` from the new window-slot count)
rather than assigning the attribute, and checks both directions -- grown and
shrunk -- asserting the pool's cap actually moved first so the test cannot pass
vacuously.

Verified: the new test fails on the unfixed tree and passes with the fix; the
rest of tests/scheduler/ is unchanged either way.

### DIFF
--- a/python/freetoken/scheduler/cache.py
+++ b/python/freetoken/scheduler/cache.py
@@ -50,11 +50,9 @@ class CacheManager:
         # lifecycle (alloc_swa / out-of-window free / free-on-finish). is_swa gates only the extra
         # SWARadixCache reuse machinery (tree match/insert/evict_swa/swa_uuid lock).
         self.swa_paged = swa_pool is not None and getattr(swa_pool, "swa_paged", False)
-        # Owned-pool capability pickup: a plugged-in swa pool may cap the prefill chunk (DSV4:
-        # ~half the window working set). Instance attrs shadow the class defaults; absent
-        # attributes leave the defaults untouched (Gemma4).
-        if swa_pool is not None:
-            self.prefill_chunk_budget = getattr(swa_pool, "prefill_chunk_budget", None)
+        # Owned-pool capability pickup is a PROPERTY (below), not a snapshot taken here.
+        # This used to read `getattr(swa_pool, "prefill_chunk_budget", None)` into an
+        # instance attribute, which froze the cap at its construction-time value.
         self.prefix_cache = self._make_prefix_cache(device, page_size, type)
         self.device = device
         self.num_pages = num_pages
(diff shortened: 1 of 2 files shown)


huhoo:
Thanks for this — the frozen-snapshot read is exactly right, and the two-direction test (driving `DSV4PagedKVCache.rebuild` instead of assigning the attribute) is the honest way to close the manager↔pool gap.

Wanted to align on related ground we've measured on DSV4-Flash (TP8, 8×RTX 4090, ~330k-token pool), in case it affects what this PR is expected to cover:

The chunk budget is not only stale across `rebuild` — it is also derived from the pool geometry at rest. In `dsv4_paged_pool.py`, the window pool's chunk budget works out to `max(P, (n_win_pages - 1 - reserved) // 2 * P)` with `reserved = dsv4_reserved_window_pages(mrr, radix)`. So growing the pool does grow the cap (your case 1), but the cap it grows **to** is a function of `n_win_pages` — it is never an independent knob. Two symptoms we hit in production:

- **Chunk inflation on pool growth**: raising `--num-tokens` at a fixed `--swa-full-tokens-ratio` roughly quadruples the chunk budget for a ~50% pool increase (that's the shape of upstream #171's 3 GB activation / OOM report at a 13.4k chunk). A live property read makes the freshly-recomputed value *visible* — it does not bound it.
- **Chunk collapse on ratio adjustment**: cutting the ratio to keep the chunk sane shrinks `n_win_pages` until the same formula floors out, and the budget collapses toward the page size.

So we ended up pinning the effective chunk with the existing `--max-prefill-length` (`scheduler.py`: `prefill_budget = min(config.max_extend_tokens, _chunk_cap)`) while evaluating a decoupling — an explicit chunk-budget override that ignores the geometric derivation, the same direction #171 suggests ("cap at a constant"). It's a 6-commit branch with CPU tests for both the grown and shrunk directions you test here, so this PR and that one compose: your property fix is the prerequisite that makes any override land on a live value.

One more thing your `reserved` readers may want to know: `dsv4_reserved_window_pages(mrr, radix)` grows with `--max-running-requests` (`2*(mrr+1) + 3*mrr + 1`), so lowering mrr raises the budget without touching the pool — worth asserting in a test if you extend this, since a rebuild at a different mrr silently moves the cap too.

Happy to share the branch if the geometric-decoupling direction is interesting to maintainers.

Issue #369, opened 2026-09-03
Unified-memory platforms (DGX Spark, sm_121): moe-backend auto resolves to offload, adding ~130s/request of pure overhead


iamanishx:
Splitting this out of #22 based on the field data @jlacroix82 posted there.

   ## Problem

   On GB10 (sm_121, 121 GB unified memory) `--moe-backend auto` resolves to the
   offload family. But unified memory has no host/device split: the offload path
   streams experts from pinned "host" banks into a GPU slot cache, which on this
   platform is a DRAM-to-DRAM copy of memory the GPU could have read directly.

   Measured in #22 (same box, same checkpoint, idle scheduler):

   - `offload` (the auto default): fixed 126-141 s stall added to every request,
     ~360 s boot
   - `fused` (explicit): 0.22-0.77 s probes, 37.8 tok/s decode, ~612 tok/s
     prefill, ~50 s boot

   So the default experience on DGX Spark is a multi-minute stall per request,
   and the fix is currently tribal knowledge (`--moe-backend fused`).

   ## Root cause

   `engine/engine.py` (`_adjust_config`, the `moe_backend == "auto"` branch):
   auto always resolves to offload, with the comment noting it can never pick
   fused because nothing knows whether the experts fit in HBM, and a wrong guess
   is a load-time OOM. That rationale is sound for discrete GPUs but inverts on
   unified memory, where fused residency is the safe default and offload is the
   regression.

   ## Proposal

   1. Detect unified-memory GPUs (cudaDevAttrIntegrated, or equivalent) and
      resolve `auto` to `fused` there, with a log line stating why. `offload`
      stays available as an explicit override.
   2. Audit the sm_121 backend probes while in the area. Two known hazards:
      flashinfer's `mm_fp8` is sm100a-only and crashes at module load on SM121,
      and #335 shows the b12x NVFP4 backend is not CUDA-graph capturable and dies
      at batch >= 2 even on sm_120. `kernel/backend.py` and
      `select_nvfp4_backend` should not select backends that cannot actually run
      on the detected arch.
   3. Document the GB10 path in docs/install.md: the source-build recipe from
      #22 (`TVM_FFI_CUDA_ARCH_LIST="12.1"`), and why fused is the default there.

   ## Scope

   Intentionally narrow: config resolution, probe guards, docs. The deeper
   unified-memory execution mode (zero-copy expert reads, rethinking the slot
   cache entirely) is a separate design discussion I will open once this lands
[...]

Issue #22, opened 2026-08-22
Feature request: ARM64/aarch64 support for NVIDIA DGX Spark


saeedkhattak:
Hi FreeToken team,

Could you please add ARM64/aarch64 support for the NVIDIA DGX Spark?

DGX Spark runs Ubuntu on the NVIDIA GB10 Grace Blackwell platform, with an ARM64 Grace CPU and 128 GB of unified memory. It seems like an ideal machine for testing FreeToken and running large MoE models locally.

Currently, the available Ubuntu package is built for AMD64/x86-64, and the installation documentation also lists Linux x86-64 as a requirement.

Could you please clarify:

1. Is DGX Spark/ARM64 support planned?
2. Can FreeToken currently be compiled from source on DGX Spark?
3. Which dependencies are preventing an ARM64 build?
4. Would you consider providing an ARM64 .deb, container image, or source-build instructions?

I would be happy to test an experimental build on my DGX Spark and report the results.

Thank you.

jessesep:
+1 on this, and an offer of test hardware.

We run four GB10 (Grace Blackwell) hosts, all `Linux aarch64`, driver 580.95 / 580.126 / 580.159, 120 GB unified memory each. The blocker is exactly as described: `freetoken` 0.1.2 publishes `manylinux_2_27_x86_64` wheels only (cp310-cp313), and `docs/install.md` lists Linux x86_64.

For what it is worth, the dependency tree looks mostly ready for aarch64 already:

- `apache-tvm-ffi` 0.1.13.post3 - `manylinux_2_24_aarch64.manylinux_2_28_aarch64` published
- `sglang-kernel` - `manylinux2014_aarch64` published
- `flashlib` 0.3.0 and `flashinfer-python` - `py3-none-any`
- `torch` >= 2.11 has aarch64+CUDA builds

So the missing piece appears to be FreeToken's own compiled extension and the `freetoken-kernel-cache` wheel, rather than the ecosystem around it.

Two questions that would help anyone attempting a source build on GB10:

1. Does the kernel cache need to be regenerated for sm_121 (GB10), or is JIT via `nvcc` the intended fallback there?
2. Are there x86-specific intrinsics in the CPU-side expert compute path (the `cpu` and `hybrid` MoE backends), or is that portable C++/OpenMP?

Happy to build from source on a GB10 box and report back what breaks, if you can point at the right starting commit or branch.

[...]

PR #445, opened 2026-09-11
fix(engine): resolve moe-strategy auto to fused on unified-memory GPUs (GB10)

iamanishx:
## Summary

   - Detect unified-memory GPUs via `cudaDevAttrIntegrated` (`_is_unified_memory_gpu()`),
     with a `FREETOKEN_UNIFIED_MEMORY=0/1` override for platforms where the attribute lies.
   - On unified memory, resolve `--moe-strategy auto` to `fused` instead of the offload
     family when the model's expert format supports residency (`_fused_resident_ok()`),
     and skip the benchbw hybrid upgrade since CPU execution adds no bandwidth when CPU
     and GPU share one memory.
   - Warn when `--moe-strategy offload` is explicitly requested on a unified-memory GPU.

   ## Problem

   The auto resolution always picks the offload family because on discrete GPUs a wrong
   "it fits" guess is a load-time OOM. On unified-memory parts (GB10/DGX Spark, sm_121)
   that rationale inverts: there is no host/device boundary, so the offload path's pinned
   staging and slot gather are DRAM-to-DRAM copies, and the slot cache double-allocates
   memory the GPU could read directly. Measured on GB10 in #22: offload (the default)
   added a fixed 126-141 s stall to every request with ~360 s boot, while fused booted in
   ~50 s and served normally. Discrete-GPU behavior is unchanged: the new branch only
   fires when `is_integrated` is set, which discrete cards never report.

   ## Validation

   Tested on a rented DGX Spark (GB10, sm_121, aarch64, driver 610.57.04, CUDA 13.0):

   - `torch.cuda.get_device_properties(0).is_integrated == 1` on GB10.
   - `pytest tests/engine tests/moe tests/kernels`: 455 passed, 4 skipped.
   - New `tests/engine/test_moe_strategy_uma.py`: 10 passed.
   - Served `Qwen/Qwen3.6-35B-A3B-FP8` with no `--moe-strategy` flag: log shows
     "Unified-memory GPU detected; auto-selecting 'fused'", 31.4 GB experts loaded
     resident, CUDA graphs captured at bs 1-2, API ready in 54 s.
   - OpenAI-compatible generation verified: warm single request 15 completion tokens
     in 0.357 s; two concurrent requests in 1.86 s.

   Refs: #369, #22, roadmap #79 (DGX Spark row).

### DIFF
--- a/python/freetoken/engine/engine.py
+++ b/python/freetoken/engine/engine.py
@@ -999,6 +999,41 @@ def _profile_gpu(index: "int | None" = None) -> Tuple[str | None, str | None]:
     return ident["name"], ident["uuid"]
 
 
+def _is_unified_memory_gpu(index: "int | None" = None) -> bool:
+    """True when the GPU has no separate device memory (cudaDevAttrIntegrated): host
+    banks and the GPU slot cache are the same DRAM, so the offload family's pinned
+    staging + slot gather are DRAM-to-DRAM copies with no PCIe link to hide behind.
+
+    The attribute is reliable on true-UMA parts (Jetson, GB10/DGX Spark) but not on
+    C2C-linked discrete-HBM parts (GH200 reports integrated=0 despite coherent CPU
+    memory), and it has only been verified on GB10 so far --
+    FREETOKEN_UNIFIED_MEMORY=0/1 overrides the probe where the attribute lies."""
+    env = os.environ.get("FREETOKEN_UNIFIED_MEMORY")
+    if env is not None:
+        return env.strip().lower() not in {"0", "false", "no", "off"}
+    if not torch.cuda.is_available():
+        return False
+    try:
+        dev = torch.cuda.current_device() if index is None else index
+        return bool(torch.cuda.get_device_properties(dev).is_integrated)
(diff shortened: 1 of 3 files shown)


iamanishx:
@jason-fxz please have a look on this

jason-fxz:
@iamanishx

Thanks for the PR.

Two requests before merging:

1. Please add a FIXME to `_fused_resident_ok` that names the formats the gate admits today, so it gets dropped once the other quant formats support fused, e.g.:

```python
def _fused_resident_ok(model_config) -> bool:
    """Whether the resident ('fused') MoE path can hold this model's experts.

    FIXME: auto resolves to fused only for bf16 and fp8_block experts; drop this gate once the other quant formats support fused.
    """
```

2. Please drop `tests/engine/test_moe_strategy_uma.py`; the monkeypatch in `tests/moe/test_offload.py` is right and enough, keep that one.

iamanishx:
@jason-fxz i have made the changes

jason-fxz:
Reviewed the update, LGTM.

PR #460, opened 2026-09-13
feat(models): support deepseek v4.1 with native fp8-fp4 kv storage

ArqAlice:
## Summary

Add experimental text and image inference for DeepSeek-V4.1 Flash NVFP4, including `LibertAIDAI/DeepSeek-V4.1-Flash-NVFP4` and the earlier `s-zaizen` checkpoint.

- Implement V4.1 model loading, CSA2 compressed attention, shared index selection, NVFP4 experts, and packed FP8/FP4 Engram lookup.
- Integrate image processing, multimodal request handling, and model-specific tokenization and tool-call parsing.
- Add `--kv-cache-dtype fp8-fp4` to preserve native FP8 window values, FP4 compressed KV, and MXFP4 index keys as codes and scales.
- Update KV allocation, memory budgeting, and cache rebuilding for packed storage. Attention reconstructs only selected tiles rather than expanding the entire cache.
- Document direct `ft serve` usage, hardware requirements, validation results, and current limitations.

Related issue: <issue URL>

## Recorded validation

Hardware: one RTX 5090 with 32 GB VRAM, AMD Ryzen Threadripper PRO 5965WX, and 512 GB host RAM.

Environment: Windows/WSL2 with a private Docker setup, NVIDIA driver 591.86, CUDA 13, PyTorch 2.11.0+cu130, and Triton 3.6.0.

Full-model checks used `LibertAIDAI/DeepSeek-V4.1-Flash-NVFP4` at revision `dfce15b92ed1fa76e80e2a46ba847e5b5451f12c`, a 32,768-token context, 1,024-token prefill chunks, and two request slots.

- 319 focused tests passed, covering codecs, attention, indexing, cache allocation/rebuilding, model execution, and engine configuration.
- An additional CUDA test passed for two requests crossing compression and page boundaries.
- Full-model Japanese, image-description, and multi-turn requests completed successfully. Health checks remained healthy.

The original focused run invoked `python -m pytest` inside a private development container. The equivalent command from the repository root is:

```bash
uv run python -m pytest \
  tests/kernels/test_dsv41_quant.py \
  tests/kernels/test_dsv41_sparse.py \
  tests/kernels/test_dsv41_indexer.py \
  tests/models/test_dsv41_attention.py \
  tests/models/test_deepseek_v41_model.py \
  tests/engine/test_deepseek_v41_engine.py \
  tests/kvcache/test_dsv41_pool.py \
  tests/engine/test_kv_quant_config.py \
  tests/engine/test_attention_backend_matrix.py \
  tests/scheduler/test_dsv4_generic_manager.py -q -p no:cacheprovider
```

## Memory results

Compared with this branch's BF16 V4.1 implementation:

| Allocation | BF16 | FP8-FP4 |
| --- | ---: | ---: |
| KV pool | 363.06 MiB | 176.40 MiB |
| Usable KV pages | 259 | 279 |
| Cached experts | 818 | 830 |

The allocated KV pool decreased by 51.4%. The automatic planner reassigned savings to additional pages and experts, so this is not a fixed-capacity comparison or an equivalent increase in free GPU memory.

## Limitations

- Packed storage preserves the existing quantization, but attention kernels can introduce small floating-point differences. Bit-identical logits and equivalent model quality are not established.
- Eager execution is required. CUDA graphs, MTP/speculative decoding, and additional NVFP4 window quantization are not supported.
- Prefill attention took approximately 2.23x as long in the isolated attention-call measurement. Initial requests can also incur JIT compilation; the API samples do not establish a general throughput improvement.
- Full million-token inference and full-checkpoint FTW startup remain unvalidated.
- The earlier broader suite run was not fully clean: an introduced AOT registration failure was fixed, four GLM failures reproduced at the unchanged base, and an intermittent copy failure passed isolated reruns.

Detailed launch settings, numerical comparisons, and API timings are recorded in `docs/deepseek-v41.md`. The unmodified base at `04d4621` lacks the V4.1 model path, so the reported A/B baseline is the branch's BF16 implementation.


minseoc03:
I went through the PR locally and found a few things that might be worth checking:

1. **`OutputHead.forward` per-token GEMV loop**
   The current loop runs one `bf16_linear_fp32` GEMV per active token, so the full output-head weight gets streamed once per row. For larger decode batches, that could get fairly bandwidth-heavy. I see why this is done though — using the `M > 1` fallback directly would materialize a large FP32 copy of the head, which is probably worse on smaller GPUs. So I don't think this needs to block the PR, but a small-M kernel that shares the weight load across rows might be worth considering as a follow-up.

2. **`num_stages=2 -> 1` in `glm_dsa_sparse.py`**
   This seems to affect the existing non-quantized GLM sparse-attention path as well, even though the shared-memory constraint mentioned in the comment looks specific to the FP8/NVFP4 path. The BF16 path already compiled with `num_stages=2` on main, so would it make sense to use something like `1 if (has_fp8 or has_nvfp4) else 2`, or at least A/B the BF16 path before changing it globally?

3. **Stricter mixed-content handling in the Responses API**
   For image-capable models, requests containing an image together with another part type (including a bare string) can now return 400 where they were previously accepted. This may be intentional, but I didn't see a test covering the new error path. If the stricter behavior is intended, it might be worth adding a regression test and mentioning the API behavior change in the PR description.

Happy to dig into any of these further or put together a patch if useful.

PR #463, opened 2026-09-14
fix(models): compute shared experts before in-place routed experts

taking-lying-flat:
BF16 prefill and resident decode overwrite `hidden_states` with the routed expert output. GLM-4 MoE, GLM MoE DSA, GLM5 Next and MiniMax-M3 currently evaluate their shared experts afterwards, producing `R(x) + S(R(x))` instead of `R(x) + S(x)`.

Compute each shared branch before `routed_forward`, preserving the kernel's existing memory reuse, and document the input mutation contract on both routed entry points.

Local validation (the regression script is retained outside this PR):

- The same tests give **16 failed, 8 passed** on unmodified `main` (`08d728d`) and **24 passed** with this change: 16 CPU cases with mutating/non-mutating expert implementations, and 8 real BF16 GPU cases covering resident prefill and decode against a PyTorch routed-expert reference.
- NVIDIA RTX A1000 Laptop GPU (4 GiB), Intel Core i7-12800HX, NVIDIA driver 595.84, Ubuntu 26.04.1; PyTorch 2.11.0 and Triton 3.6.0.
- Checkpoint: none; tiny randomly initialized blocks. Full-checkpoint parity was not tested.
- `git diff --check` passes.

Exact local test command, run from the checkout's `python/` directory:

```bash
uv run --no-project --python /tmp/freetoken-glm-shared-venv/bin/python python -m pytest /tmp/freetoken-glm-shared-validation/test_shared_expert_input.py -q --tb=short
```


jason-fxz:
Confirmed: `fused_experts_impl` writes the routed sum back into `hidden_states` (bf16 experts, resident both phases / offload prefill), so these four blocks fed `R(x)` to the shared expert; computing the shared branch first is the right minimal fix and the NVFP4/FP8 paths stay bit-identical. Merging.

taking-lying-flat:
> Confirmed: `fused_experts_impl` writes the routed sum back into `hidden_states` (bf16 experts, resident both phases / offload prefill), so these four blocks fed `R(x)` to the shared expert; computing the shared branch first is the right minimal fix and the NVFP4/FP8 paths stay bit-identical. Merging.

How Can I. contact you

jason-fxz:
email?

> > Confirmed: `fused_experts_impl` writes the routed sum back into `hidden_states` (bf16 experts, resident both phases / offload prefill), so these four blocks fed `R(x)` to the shared expert; computing the shared branch first is the right minimal fix and the NVFP4/FP8 paths stay bit-identical. Merging.
> 
> How Can I. contact you

taking-lying-flat:
> email?
> 
> > > Confirmed: `fused_experts_impl` writes the routed sum back into `hidden_states` (bf16 experts, resident both phases / offload prefill), so these four blocks fed `R(x)` to the shared expert; computing the shared branch first is the right minimal fix and the NVFP4/FP8 paths stay bit-identical. Merging.
> > 
> > 
> > How Can I. contact you

yeap

jason-fxz:
you can find it in my GitHub profile

Issue #307, opened 2026-08-31
temperature 0 is not reproducible with thinking on (identical requests diverge on near-ties)


benwilson:
## Summary

With thinking **on**, `temperature: 0` does not give reproducible output: three identical requests produced three different reasoning traces. With thinking **off** the same prompt is byte-identical across runs. Explicit sampling parameters are honoured (so this is not `--sampling-defaults model` overriding the request); it looks like a nondeterministic kernel flipping near-tied argmax picks, which content-heavy reasoning text exposes and peaked code output does not.

**Environment:** FreeToken git `4b94bdc` (HEAD 2026-08-30, no local patches), `pip install --no-build-isolation "freetoken[accel] @ git+…@4b94bdc"`; torch 2.11.0+cu130, CUDA 13.0, driver 595.71.05; RTX 3090 24 GB (sm_86), PCIe 4.0 x16 (`ft bench bw`: H2D 23.8 GB/s), 32-core host, 503 GiB RAM; `dealignai/Qwen3.8-Flash-Next-ABLITERATED-NVFP4` @ `be794b99` (compressed-tensors NVFP4, same layout as the RadixArk stock checkpoint).

`ft serve --max-seq-len-override 262144 --kv-reserve-tokens 262144 --moe-backend offload --moe-cache-auto --nvfp4-backend triton --max-running-requests 1 --max-prefill-length 8192 --sampling-defaults model --reasoning-parser qwen3 --tool-call-parser qwen3_coder`

## Repro

Prompt: `Write a Python function to reverse a linked list. Output only code.`
Request: `{"temperature": 0.0, "max_tokens": 4096}` (thinking on = template default), three times in a row, one at a time.

| run | completion tokens | reasoning sha1 (first 10) | content sha1 |
|---|---|---|---|
| 1 | 206 | 7f4e39e460 | 1c6a96da5e |
| 2 | 218 | bb6999e009 | 1c6a96da5e |
| 3 | 288 | 10921af11d | df8677eeec |

First divergence between runs 1 and 2 is at character 210 of the reasoning, on a near-tie:

```
A: …nal only code no explanation. Could include class? Function reverse linked list. Need maybe iterativ…
B: …nal only code no explanation. Could include class? Function to reverse linked list. Need maybe itera…
```

Controls, same prompt, `max_tokens: 1200`, `chat_template_kwargs: {"enable_thinking": false}`, two runs each:

| params | result |
|---|---|
[...]

PR #471, opened 2026-09-14
fix(engine): preserve greedy sampling in mixed batches

taking-lying-flat:
A greedy request with fixed logits `[1.0, 1.0, 0.0]` returns token 0 alone, but can return token 1 when batched with a random request: the mixed path approximates greedy with temperature `1e-6`, leaving the tied maxima equally likely. Also, `temperature=0, top_p=0.9` unnecessarily enters probability sampling.

Keep a per-request `greedy_mask` and explicitly select `argmax` for greedy rows. Treat `temperature <= 0` or `top_k == 1` as greedy regardless of `top_p`, and use neutral backend parameters for those rows. The all-greedy fast path is preserved. Mixed batches still run probability sampling across all rows; skipping that computation is outside this correctness fix.

Related: #307. This fixes sampling-policy differences for identical logits; it does not establish or resolve the cause of the serial model-level divergence reported there.

Validation:

- Ran local deterministic CPU checks covering tied maxima, standalone versus mixed batches, FP32/FP16/BF16, and all four top-k/top-p modes. Before the fix, the mixed regression returned `[1, 2, 2]` instead of `[0, 2, 1]`, and zero-temperature requests with `top_p < 1` entered probability sampling. The validation file is retained locally and is not included in this PR.
- On upstream base `e0886cc` plus this patch: **73 passed, 4 skipped**, including real Triton CUDA sampling. The four FlashInfer cases were skipped because FlashInfer is not installed. `git diff --check` passed.
- Hardware: NVIDIA RTX A1000 Laptop GPU; Intel Core i7-12800HX; NVIDIA driver 595.84; PyTorch 2.11.0+cu130. Checkpoint: N/A, tests use fixed logits without loading a model.

Exact local validation command, from the repository root, using the untracked test file:

```bash
env PYTHONPATH=python /tmp/freetoken-glm-shared-venv/bin/python -m pytest tests/engine/test_sample.py -q -rs
```


jason-fxz:
LGTM, merging.

PR #502, opened 2026-09-17
feat(deepseek_v4): serve image input on the V4-Flash-Vision release

void-mckenzie:
Image input for deepseek-ai/DeepSeek-V4-Flash-Vision-Exp (`DeepseekV4ForCausalLM`), the last open item under Multimodal in #79, on the multimodal path from #454, #467, #479, #480 and #481.

## What

- **Processor.** `<｜deepseek_image｜>` (id 129264) expands into the block the reference's `prepare_vl_inputs` builds: lead pads, `IMAGE_START`, the N-layout rows, the parity pads, `IMAGE_END`, one content pad id for the whole span instead of the reference's `vocab_size + type` ids. Lead pads align the block's first row token to a 4-token compression stride, so its length depends on where it lands: `prompt_replacement` receives that offset (a defaulted argument, existing families ignore it). `--image-max-tokens` lowers the checkpoint's per-image cap (`vision_max_n_token`, 384).
- **Tower.** A 32-block native-resolution ViT with bidirectional attention and 2-D rope (fp32 tables), then the 3x3/stride-3 aligner into the 4096 text width. Blocks stream from pinned host banks under the default `--mm-encoder-weights host`: 0.59 GiB free after CUDA graph capture against 0.75 GiB text-only, so the tower is ~0.16 GiB resident.
- **Attention.** An image span's tokens stay mutually visible in the sparse prefill path: the window candidates follow the span's own bounds, capped at `vision_max_n_token`, as the reference's `get_image_visible` / `get_window_topk_idxs_visible` pair does. The spans arrive as `batch.mm_spans`, derived in `_gather_multimodal` from the per-token block ends `plan_mm_batch` already reports. Chunks keep images whole through `keep_images_whole` / `mm_chunk_end`; only a block larger than the prefill budget is still cut, with the usual warning.
- **Router.** `bias_vl` shifts expert selection on image rows only, on the hash layers and the score-routed ones alike; the routing weights stay the unbiased scores.
- **Reader.** `vision.*`, `aligner.*` and `image_*` (tower, aligner, the four sentinel vectors) load when an encoder is built and are skipped under `--text-model-only`, which also serves this checkpoint text-only (both APIs then refuse images). `get_mm_processor` reads `vocab_size` off the flat config this checkpoint ships. `docs/models.md` gains the checkpoint and its row of the image-input table.

## Verification

- Layout and preprocessing against the checkpoint's own `inference/` code: `build_image_block` bit-equal over 25 grid/offset combinations, `load_image` bit-equal (patches and grids) over 4 image sizes, and the tokenizer expansion matches `prepare_vl_inputs` on the real tokenizer (same token count, same block window, bit-equal patches; the last two gated on `FREETOKEN_DSV4_VISION_CKPT` / `FREETOKEN_DSV4_VL_MODEL`).
- deepseek-ai/DeepSeek-V4-Flash-Vision-Exp on an RTX 5090 (32 GB), driver 580.95.05, started with `ft serve --model /path/to/Flash_Vision_Exp --gpu 0 --num-tokens 311296 --moe-cache-size 750`; `/v1/stats` reports `input_modalities: ["text", "image"]`, image requests describe the image correctly on `/v1/chat/completions` (data URL) and `/v1/messages` (base64 block), two images in one request are described in order, decode is 20.4-21.0 tok/s at running-req 1.
- Prefix caching through an image span, same 600-token prefix, temperature 0, `#cached-token` from the prefill log: 0 -> 512 (image A, first sight) -> 512 (text again) -> 640 (image A again) -> 512 (different image, same slot). The 640 is the span matched by content; the last request is the control.
- `pytest tests/ -m "not slow"`: 1768 passed, 208 skipped, 3 failed. The three are quant-config tests that fetch their checkpoints from the Hub and fail identically on a clean v0.1.3.

## Not in this PR

- An FTW conversion of this checkpoint: the reader side is in place (`iter_vision_weights`, the new prefixes), but the 163 GB conversion was not run, so that path is untested.


void-mckenzie:
@jason-fxz since youve been working through the vision families, were you already on deepseek v4 flash vision exp or is this useful? its working on both apis here, details are in the description. happy to test more configs (num-tokens, moe cache size, longer prompts) if you want anything specific checked before you review.

PR #505, opened 2026-09-17
fix(scheduler): preserve pending hybrid checkpoint across prefill chunks

eamars:
### Issue: 
FreeToken sometimes processed an identical long prompt from scratch instead of reusing its previous work—even when checkpoint memory was available.

### Model and Hardware:
GPU used | NVIDIA RTX 5090, 32 GB
CPU | AMD Ryzen 9 9950X3D, 16 cores / 32 threads
System RAM | Approximately 128 GB; 127.64 GiB visible
Operating system | Windows 11 Pro, build 26200
Linux environment | Ubuntu 26.04 LTS under WSL2
NVIDIA driver | 610.74
CUDA / PyTorch | CUDA toolkit 13.3 / PyTorch 2.11.0+cu130
**Model | orcarouter/Qwen3.8-Flash-Next-Uncensored-NVFP4**
Baseline | FreeToken v0.1.3, commit cac247a -> Issue exists on v0.1.2 too. 

### Below is generated by gpt-6 for detailed fix:

```
The model weights loaded successfully. The defect was in retaining reusable inference state afterward.
This model needs a saved internal state and its exact position in the prompt. Think of a saved calculation labelled “completed through token 16,320.”
FreeToken creates a new request object for each input chunk. It copied the saved-state buffers into the next request, but forgot to copy the token-position label: mamba_last_track_seqlen.
Usually, the final chunk creates a newer snapshot, hiding this omission. A final chunk of 64 tokens or fewer cannot create that replacement. The existing snapshot can therefore become unavailable for subsequent requests.
The fix carries the missing position into the continuation request. If a newer snapshot is created, the existing code replaces it normally.
That is the whole production change: five added lines, including two comment lines. It does not enlarge the cache, change model weights, or change GPU kernels. The existing rules for saving and freeing buffers remain intact.
```

### Next Actions:
I will continue to test in the next few days to see if this will actually fix the issue.


eamars:
Tested for two days with deepseek harness and no checkpoints was missed.

